PR #20, opened 2018-04-09
fix(account switcher): can be closed by clicking below

drazik:
Fixing the bug in which clicking below the account switcher (green areas in following screenshots) modal didn't close it.

desktop
mobile

### DIFF
--- a/src/ducks/account/AccountSwitch.styl
+++ b/src/ducks/account/AccountSwitch.styl
@@ -53,10 +53,10 @@
     position fixed
     top 78px
     left 0
-    bottom 10px
     width 440px
     overflow auto
     z-index: 60
+    max-height: "calc(100vh - %s)" % @top
 
 .account-switch-menu
     margin 20px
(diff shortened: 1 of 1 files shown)


src/ducks/account/AccountSwitch.styl
```diff
@@ -53,10 +53,10 @@
     position fixed
     top 78px
     left 0
-    bottom 10px
     width 440px
     overflow auto
     z-index: 60
+    max-height: calc(100vh - 78px)
```
ptbrowne: would be good to store the 78px in a variable for clarity

src/ducks/account/AccountSwitch.styl
```diff
@@ -137,11 +137,10 @@
         z-index 61
         position fixed
         top 58px
-        bottom 10px
         left 50%
         margin-left -200px
         width 400px
-        max-height inherit
+        max-height calc(100vh - 68px)
```
ptbrowne: same here

src/ducks/account/AccountSwitch.styl
drazik: I used [property lookup](http://stylus-lang.com/docs/variables.html#property-lookup), is it OK for you ?

PR #27, opened 2018-04-12
fix: handle operations labels that are too long

drazik:
Add ellipsis if the operation label overflows : 

image

### DIFF
--- a/src/ducks/transactions/Transactions.styl
+++ b/src/ducks/transactions/Transactions.styl
@@ -32,6 +32,9 @@
     .bnk-op-caticon
         cursor pointer
 
+    .bnk-op-label
+        max-width 100%
+
     +medium-screen()
         .bnk-op-desc
             cursor pointer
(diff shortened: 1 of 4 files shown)


ptbrowne:
This feels weird to me :/ Can you check with flo ?

drazik:
He told me that since we already do that on the mobile side, we must do the same on desktop.

src/ducks/transactions/Transactions.styl
```diff
@@ -32,6 +32,9 @@
     .bnk-op-caticon
         cursor pointer
 
+    .bnk-op-label
+        max-width 100%
```
ptbrowne: I think this should be done automatically on `media__body`

ptbrowne: Can you add a TODO here ?

ptbrowne: I don't understand why this is necessary ? Has the ListItemText a problem with Media Bd ?

PR #84, opened 2018-04-23
fix: actions are more robust

drazik:
There were some problems on actions on mobile (wrong button type in the modal mostly).

I also fixed some little things after Joel's design review (transaction row padding, category icon size).

### DIFF
--- a/src/ducks/transactions/TransactionActions.styl
+++ b/src/ducks/transactions/TransactionActions.styl
@@ -20,3 +20,14 @@
 
 .TransactionActionMenuItem
     text-align left
+
+.TransactionActionButton
+    +medium-screen()
+        // This element increases the click area of the button
+        &:before
+            content ''
+            position absolute
+            top 0
+            bottom 0
+            width 1rem
+            right 100%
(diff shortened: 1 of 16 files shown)


src/ducks/transactions/TransactionActions.styl
```diff
@@ -20,3 +20,13 @@
...
+
+.TransactionActionButton
+    +medium-screen()
+        &:before
+            content ''
+            position absolute
+            top 0
+            bottom 0
+            width 1rem
+            right 100%
```
ptbrowne: can you add comment saying this is to enlarge the click area (if I've correctly understood)

src/ducks/transactions/Transactions.styl
```diff
@@ -26,16 +26,16 @@
     .bnk-transaction-mobile
         & > td
             width 100%
+            padding-top .75rem
+            padding-bottom .75rem
```
ptbrowne: why is this necessary ?

drazik: Because the base vertical padding on tables is `1rem`, but here we want the padding to be a little bit smaller.

PR #94, opened 2018-04-27
Infinite scrolling for transaction page

ptbrowne:
The transaction pages gets a 2 way infinite scrolling.
The date selector is showing the current month.
It also enables jumping to another month.


src/ducks/filters/index.js
```diff
@@ -64,7 +64,7 @@ export const getTransactionsFilteredByAccount = createSelector(
     if (accountIds) {
       transactions = filterByAccountIds(transactions, accountIds)
     }
-    return transactions
+    return sortBy(transactions, t => -new Date(t.date))
```
drazik: If find it's not very obvious that it sorts from most recent to most ancient. Maybe putting the result in a variable with a clear name before returning it would help reading.

src/ducks/transactions/InfiniteScroll.jsx
```diff
@@ -0,0 +1,156 @@
...
+  constructor(props) {
+    super(props)
+    this.checkForLimits = throttle(this.checkForLimits, 100)
+  }
+
+  componentDidMount() {
+    this.main = this.getScrollingElement()
+    if (!this.main) {
+      throw new Error(
+        'getScrollingArea returned null, make sure it returns a node.'
```
drazik: That's `getScrollingElement`, not `getScrollingArea`

src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -79,26 +85,52 @@ const Separator = () => (
   <hr style={{ fontSize: 0, border: 0, height: '0.5rem' }} />
 )
 
+const getMonth = date => date.slice(0, 7)
```
drazik: Why `slice(0, 7)` ?

src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -109,6 +141,62 @@ class TransactionsPage extends Component {
...
+          : [filterAfter, filterBefore]
+      for (let filter of filters) {
+        const months = validMonths.filter(filter)
+        if (months.length > 0) {
+          month = months[0]
+          break
+        }
+      }
+      limitMin = findMonthIndex(month)
+    }
```
drazik: Can you explain what's going on here ? Actually I understand that you are trying to get something if the requested month has no transactions, but I don't understand the filtering part

src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -109,6 +141,62 @@ class TransactionsPage extends Component {
...
+      limitMin: limitMin,
+      limitMax: limitMin + 5,
+      currentMonth: month,
+      infiniteScrollTop: false
+    })
+  }
+
+  checkToActivateTopInfiniteScroll = getScrollInfo => {
+    const scrollInfo = getScrollInfo()
+    if (scrollInfo.scroll > 150) {
```
drazik: What is 150 ? Is it possible to put it in a constant with a clear name ?

src/ducks/transactions/Transactions.jsx
```diff
@@ -192,28 +192,74 @@ const groupByDateAndSort = transactions => {
...
     getScrollInfo => {
       this.props.onScroll(getScrollInfo)
+      this.updateTopMostVisibleTransaction()
     },
     100,
     { leading: false, trailing: true }
   )
+
+  handleRefRow = (transactionId, ref) => {
+    const node = ReactDOM.findDOMNode(ref) // eslint-disable-line
```
drazik: `ref` is not supposed to be the `node` already ?

ptbrowne: a ref can be an instance of a React component.

> When the ref attribute is used on an HTML element, the ref created in the constructor with React.createRef() receives the underlying DOM element as its current property.
When the ref attribute is used on a custom class component, the ref object receives the mounted instance of the component as its current.

https://reactjs.org/docs/refs-and-the-dom.html

src/ducks/transactions/TransactionsPage.jsx
ptbrowne: dates are strings in the transactions, ex: 2014-08-25T08:00, so a slice until the 7th character will only take the month part 2014-08.

PR #98, opened 2018-04-30
fix: account switch was appearing in account/group settings

ptbrowne:


src/ducks/settings/GroupsSettings.jsx
```diff
@@ -28,7 +28,7 @@ const GroupList = withRouter(
         <tbody>
           {groups.map(group => (
             <tr
-              key={group.label}
+              key={group._id}
```
kosssi: À quoi ça sert de mettre un attribut `key` ?

PR #101, opened 2018-05-03
Move account switch inside page, improved select dates

ptbrowne:
- Account switch is no longer in the sidebar and has been moved
inside the transactions and categorization page

- Date selector has been improved to facilitate comparing years and months


src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -275,16 +287,20 @@ class TransactionsPage extends Component {
             period={this.state.currentMonth}
             onChange={this.handleChangeMonth}
           />
-          <TransactionTableHead />
+          {filteredTransactions.length > 0 ? (
+            !isMobile ? (
```
kosssi: `!isMobile && <TransactionTableHead />` pourrait fonctionner ?

kosssi: pareil https://github.com/cozy/cozy-banks/pull/101/files#diff-ceb283ed4feb6c879d56b055bed20371R273 et https://github.com/cozy/cozy-banks/pull/101/files#diff-ceb283ed4feb6c879d56b055bed20371R277

ptbrowne: fixed in https://github.com/cozy/cozy-banks/pull/101/commits/d651abc3b74a1c2455330d74991275a8f368f1b3

src/components/Select/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+        <option key={option.value} value={option.value}>
+          {option.name}
+        </option>
+      ))}
+    </select>
+    <SmallArrow />
+  </span>
+)
+
+export default translate()(props => <Select {...props} />)
```
drazik: Why not just `export default translate()(Select)` here ?
Also, it seems you are not using the `t` prop, so maybe the `translate` HOC is not required ?

src/components/Select/styles.styl
```diff
@@ -0,0 +1,19 @@
+@require 'settings/palette'
+
+.Select
+    appearance none
```
drazik: Did you test this on IE ? I remember it doesn't support `appearance`, so it renders the system arrow (see https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance). There are workarounds for IE : https://stackoverflow.com/questions/1895476/how-to-style-a-select-dropdown-with-css-only-without-javascript/13968900#13968900

src/ducks/account/AccountSwitch.jsx
```diff
@@ -276,8 +277,7 @@ class AccountSwitch extends Component {
       breakpoints: { isMobile, isTablet, isDesktop }
     } = this.props
     const { open } = this.state
-    let { accounts, groups, groupsDocs } = this.props
-    const isFetching = isLoading(accounts) || isLoading(groupsDocs)
+    let { accounts, groups /* groupsDocs */ } = this.props
```
drazik: There is an unneeded comment here I think

src/components/Select/index.jsx
ptbrowne: true, that's weird

src/components/Select/styles.styl
ptbrowne: OK thanks !

PR #114, opened 2018-05-09
feat(dates filter): new style on mobile

drazik:
There's no shadow on the bottom side for the moment because Joël wants it to be deployed everywhere at the same time.

image


ptbrowne:
> wants it to be deployed everywhere at the same time.

What do you mean ?

drazik:
It means on all our apps.

PR #126, opened 2018-05-16
repair styleguide     🐝 fix: repair styleguide     🔧 chore: use components from cozy-ui     🔨 refactor: makes renderable by react     🔧 chore: remove unused components from styleguide     🔧 chore: extract AppLike from test     🔧 chore: bump cozy-ui

ptbrowne:


styleguide.config.js
```diff
@@ -36,17 +16,10 @@ module.exports = {
     {
       name: 'General',
       components: () => [
-        'src/components/Table/index.jsx',
+        // 'src/components/Table/index.jsx',
```
drazik: Should this line be removed if it is not required anymore ?

src/ducks/filters/scrollAware.js
```diff
@@ -37,7 +37,9 @@ export default Wrapped =>
       }
     }, 16)
 
-    render(props, { scrolling }) {
+    render() {
+      const props = this.props
+      const { scrolling } = this.state
```
drazik: Should we do the same thing everywhere ? I think I saw some components using props and state from `render` arguments « the Preact way » while working on categorization pages

ptbrowne: Yes I think it would be good to do it step by step. For example enzyme only works with react.

styleguide.config.js
ptbrowne: Not really, it's only because Table requires stuff directly from cozy-ui and there is a problem with the cozy-client-js required in cozy-ui, in the styleguide.

PR #131, opened 2018-05-18
feat(CategoriesChart): always sort parts in reverse order

drazik:
The colors were not sorted in the right order in the categories chart. It's now the case.

https://testbanksfixcategorieschart-banks.cozy.works/#/categories


ptbrowne:
Do you mind deploy2cozy ? I would like to see the resulting behavior :)

drazik:
Yep I edited the original comment with the URL.

PR #137, opened 2018-05-21
Some fixes of CategoriesChart

drazik:
* Fix a problem with the chart colors when showing subcategories
* Pass the selected category as an object instead of only its name
* Pass the subcategories if a category is selected, categories otherwise


ptbrowne:
Great 🎉

PR #141, opened 2018-05-22
Show a modal on mobile app when there are some warnings

drazik:
On app start and resume, we check if the stack has some warnings to show. If it's the case, we show a modal with the warnings.


src/components/App.jsx
```diff
@@ -1,27 +1,70 @@
...
+        {warnings &&
+          warnings.map(warning => (
+            <WarningsModal
+              key={warning.code}
+              code={warning.code}
+              title={warning.title}
+              detail={warning.detail}
+              links={warning.links}
+            />
+          ))}
```
ptbrowne: Not blocking : I would prefer to have a Warning component, always displayed, in charge of the events and of rendering the WarningsModal. This way, the warnings logic stays outside the App.

PR #146, opened 2018-05-24
Some fixes on categories page

drazik:
* Remove padding on income toggle on mobile
* Fix "transaction(s)" plural
* If a category has no data, then show "－ €" as its total
* When the selected period has no data, don't show the income toggle and the graph

### DIFF
--- a/src/ducks/categories/Categories.jsx
+++ b/src/ducks/categories/Categories.jsx
@@ -134,6 +134,7 @@ class Categories extends Component {
     const rowClass = stRow
     const onClick = subcategory || isCollapsed ? this.handleClick : () => {}
     const key = (subcategory || category).name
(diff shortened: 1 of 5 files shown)


src/ducks/categories/Categories.jsx
```diff
@@ -259,7 +259,7 @@ class Categories extends Component {
                     <span>
                       {`· ${transactionsNumber} ${t(
                         `Categories.headers.transactions.${
-                          transactionsNumber > 1 ? 'single' : 'plural'
+                          transactionsNumber > 1 ? 'plural' : 'single'
```
ptbrowne: genius !

drazik: Héhé

PR #164, opened 2018-05-28
feat(TransactionModal): use Page component on mobile

drazik:
This introduce a `Page` component that is 99% like a `Modal`, but shows the footer. It could be extracted in Cozy UI if it's needed. But maybe it needs a cleaning step before.

With that, we don't have problems with the footer being above modals anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "cozy-client": "0.4.3",
     "cozy-client-js": "0.9.0",
     "cozy-konnector-libs": "^3.0.13",
-    "cozy-ui": "git://github.com/cozy/cozy-ui.git#v9.0.1",
+    "cozy-ui": "9.1.0",
     "date-fns": "^1.28.2",
     "hammerjs": "^2.0.8",
     "localforage": "^1.5.0",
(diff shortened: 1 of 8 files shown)


drazik:
Actually I think I totally messed up with this one. I just thought about something better : the `Page` component should not duplicate some `Modal` styles but reuse them.

My plan is : 

* Duplicate the `render` method of the `Modal` component : I don't see how to achieve the same structure but with an additional className since `Modal` doesn't take `className` props for its `Overlay` component and its root element
* Take the same props as the original `Modal`
* Force `mobileFullScreen` prop to `true`
* Add some styles specific to the `Page` component (notably on the `Overlay` and the root element `height`)

What do you think about it ? There's still duplication, but I don't see a way to do it without any.

ptbrowne:
> since Modal doesn't take className props for its Overlay component and its root element

Maybe `Modal` should take `rootClassName` and `overlayClassName`. I would not be against it.

drazik:
That was what I asked in https://github.com/cozy/cozy-ui/issues/482. But after re-reading it, I feel I wasn't very clear about my intentions.

PR #187, opened 2018-06-06
 performance improvements for ios

ptbrowne:
✨ feat: add manual buttons for infinite scroll on ios    (deactivate automatic infinite scroll on ios)
 🐝 fix: deactivate touchmove as it makes the app unusable     
✨ feat: when loading transactions backward, do not stop in the middle of     
🚀 perf: add fastclick


src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -172,12 +184,21 @@ class TransactionsPage extends Component {
...
+      {
+        limitMin: limitMin,
+        limitMax: limitMin + 10,
+        currentMonth: month,
+        infiniteScrollTop: false
+      },
+      () => {
+        // need to scroll past the LoadMore button
+        if (onIOS()) {
+          const scrollTo = limitMin > 0 ? 65 : 0
```
drazik: Does that mean that if for any reason we change the style of the LoadMore button, we also need to change the value here ? Is it possible to retrieve the button height dynamically ?

ptbrowne: Yes, will try to do that :)

drazik: But it's not blocking for me. It's just because I'm 100% sure that we will forget to update it if we update the LoadMore button :p

src/ducks/transactions/Transactions.jsx
```diff
@@ -256,12 +284,12 @@ class TransactionsD extends React.Component {
...
   }
 
-  handleScroll = throttle(
+  handleScroll = (onIOS() ? debounce : throttle)(
     getScrollInfo => {
       this.props.onScroll(getScrollInfo)
       this.updateTopMostVisibleTransaction()
     },
-    100,
+    300,
```
kosssi: pourquoi avoir changé cette valeur ?

ptbrowne: parce que je pensais pas que updater le selecteur de date et le check pour savoir si on doit scroller vers l infini doit etre fait toutes les 100ms.

PR #196, opened 2018-06-11
feat: show institution label in transaction rows

ptbrowne:
Tiny change to show the bank name along the account name
in transaction rows.

browser

image

mobile

image

I took this opportunity to fix and add tests.
Tests should be run automatically from Travis from now on :)


ptbrowne:
Some tests seem to be failing but there are not related to the part I worked on, we have to keep this in mind :)

ptbrowne:
Thanks Cyrille !

PR #207, opened 2018-06-13
🐝 fix: disable next button when months after have no transactions

ptbrowne:
The next button is disabled if all options after it are disabled.


src/components/SelectDates/utils.js
```diff
@@ -0,0 +1,16 @@
+const rangedSome = (arr, predicate, start, end) => {
+  start = Math.max(start, 0)
+  end = Math.min(end, arr.length)
+  const step = start > end ? -1 : 1
+  if (!arr.length > 0) {
```
drazik: This condition is weird. Why not `arr.length === 0` ?

ptbrowne: True

PR #219, opened 2018-06-18
chore: Add script to list brands from github ⚙️

kosssi:


scripts/brands.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const getRepos = limit => githubJsonUrl(`https://api.github.com/orgs/konnectors/repos?per_page=${limit}`)
+const repoFilter = repos => repos
+  .filter(repo => !includes(NOT_KONNNECTORS, repo.name))
+  .map(repo => repo.name)
+const getManifest = repo => githubJsonUrl(`https://raw.githubusercontent.com/konnectors/${repo}/build/manifest.konnector`)
+const getIndex = repo => githubUrl(`https://raw.githubusercontent.com/konnectors/${repo}/build/index.js`)
+
+const getIdentifiers = async repo => {
```
drazik: I think adding some comments to this function would help understand what it does. that's not obvious at first sight

scripts/brands.js
```diff
@@ -0,0 +1,108 @@
...
+const githubUrl = url => {
+  return new Promise((resolve, reject) => {
+    request.get(getRequestOptions(url), (error, response, body) => {
+      if (error) {
+        reject(error)
+      }
+      resolve(body)
+    })
+  })
+}
```
ptbrowne: request-es6-promise would give you promises with requests automatically

scripts/brands.js
kosssi: I add more comments ;)

kosssi:
Il faut comprendre pourquoi le build ne fonctionne plus ...

ptbrowne:
Que fait-on de cette PR kosssi ?

PR #220, opened 2018-06-18
Add some links to email notifications

drazik:
Add the following links in the email notifications : 

* Banks home
* Transactions
* Balances
* Settings

I also added a `tGlobal` helper to handlebars to translate some common strings across all types of notifications.


src/ducks/notifications/helpers.spec.js
```diff
@@ -0,0 +1,13 @@
+import { getBanksUrl } from './helpers'
+
+describe.only('getBanksUrl', () => {
+  it('should should append "-banks" to the first component', () => {
+    expect(getBanksUrl('test.cozy.works')).toBe('test-banks.cozy.works/#/')
```
ptbrowne: The URL would normally contain `https://` right ? If so, I think it should be also in the test

PR #227, opened 2018-06-19
refactor: Use cozy-device-helper 🔥

kosssi:


src/ducks/mobile/push.js
```diff
@@ -1,6 +1,6 @@
 /* global PushNotification, cozy */
 import { hashHistory } from 'react-router'
-import { onIOS } from 'utils/platform'
+import { isIosApp } from 'cozy-device-helper'
```
ptbrowne: Not blocking : isIOSApp would be more coherent

PR #233, opened 2018-06-21
Informative modal health

drazik:
This add a variation of the informative modal for health connectors : 

A different label for the CTA : 
image

A different content : 
image

For that, I made `InformativeModal` more generic by passing its content to its props.

### DIFF
--- a/src/ducks/transactions/actions/KonnectorAction.jsx
+++ b/src/ducks/transactions/actions/KonnectorAction.jsx
@@ -28,7 +28,15 @@ import iconCollectAccount from 'assets/icons/icon-collect-account.svg'
 
 const name = 'konnector'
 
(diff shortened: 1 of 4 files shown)


src/ducks/transactions/actions/KonnectorAction.jsx
```diff
@@ -132,9 +144,16 @@ class Component extends React.Component {
           )}
           {this.state.showInformativeModal && (
             <InformativeModal
-              brand={brand}
+              brandName={brand.name}
```
ptbrowne: This is no longer used I guess ?

drazik: Exactly

PR #250, opened 2018-06-25
fix(TransactionSelectDates): show all the months of the transactions

drazik:
Currently, a month is missing on the date selector. Eg. if I have transactions from 03/2018 to 06/2018, I see april, may and june in the month select, but not march.

This fixes it.


drazik:
Au temps pour moi, on close alors :)

PR #253, opened 2018-06-26
refactor: use Layout UI component

GoOz:
Use Layout component from Cozy-UI.

It should be better on iPhone X too because of the Layout component.


src/styles/main.styl
```diff
@@ -13,6 +14,22 @@ html, body
...
+.c-content
+    width 100%
+    padding content-padding
+
+    +small-screen()
+        &
+            padding content-padding-small
+
+    +medium-screen()
+        display block
```
ptbrowne: Why do we need to have this on our side ?

src/styles/main.styl
```diff
@@ -13,6 +14,22 @@ html, body
         overflow-y scroll // necessary for momentum scroll on ios
         -webkit-overflow-scrolling touch // necessary for momentum scroll on ios
 
+    +medium-screen()
+        min-height 100vh
+        overflow auto
+        overflow-x hidden
```
ptbrowne: Why do we need to have this on our side ?

GoOz: It's Banks' layout specificities that aren't in Cozy-UI. 
It was in Layout.styl before but since the component is no longer useful I moved this here.
Did you have another place in mind?

src/styles/main.styl
GoOz: Same answer as before.

src/styles/main.styl
ptbrowne: No but I did not know we had specifities. Shouldn't this specificities be integrated into the normal layout ?

GoOz: Honestly I'm not sure what it is for, what it's supposed to fix. I removed it at first and didn't see any change but I thought It had a reason so I kept it. Especially because it's related to the layout and you don't mess with the layout. 😝 
But you know Banks better than I do, maybe you have a way of knowing if it's meant for something or not, or if it's still useful or not. Other apps don't need it though.

As for the class `.c-content` it is useful, for Bank only, so it should remain here.

ptbrowne: Ok thanks, we keep in mind that we may try to remove it :)

ptbrowne:
LGTM

PR #260, opened 2018-06-27
🐝 fix: supports legacy triggers

ptbrowne:
Claire has a legacy trigger whose content is encoded in base64. Banks did not recognize it as a konnector trigger and would display the "Install the connector" CTA, even if the konnector is already installed (and has a working trigger).

### DIFF
--- a/src/utils/triggers.js
+++ b/src/utils/triggers.js
@@ -1,0 +1,11 @@
+const getKonnectorFromTrigger = trigger => {
+  if (trigger.message && trigger.message.konnector) {
+    return trigger.message.konnector
+  } else if (trigger.message.Data) {
+    // Legacy triggers
+    const message = JSON.parse(atob(trigger.message.Data))
+    return message.konnector
+  }
+}
+
+export { getKonnectorFromTrigger }
--- a/src/utils/triggers.js
+++ b/src/utils/triggers.js
@@ -1,8 +1,11 @@
+const { logInfo } = require('lib/sentry')
+
 const getKonnectorFromTrigger = trigger => {
   if (trigger.message && trigger.message.konnector) {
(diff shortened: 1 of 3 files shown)


src/utils/triggers.js
```diff
@@ -0,0 +1,11 @@
...
+  if (trigger.message && trigger.message.konnector) {
+    return trigger.message.konnector
+  } else if (trigger.message.Data) {
+    // Legacy triggers
+    const message = JSON.parse(atob(trigger.message.Data))
+    return message.konnector
+  }
+}
+
+export { getKonnectorFromTrigger }
```
ptbrowne: Do you think I should put this in cozy-doctypes/triggers ?

src/utils/triggers.js
```diff
@@ -0,0 +1,11 @@
+const getKonnectorFromTrigger = trigger => {
+  if (trigger.message && trigger.message.konnector) {
+    return trigger.message.konnector
+  } else if (trigger.message.Data) {
+    // Legacy triggers
+    const message = JSON.parse(atob(trigger.message.Data))
```
kosssi: est ce que tu peux ajouter un log sentry pour que l'on puisse vérifier si on passe encore dans cette fonction dans 6 mois ?

src/utils/triggers.js
drazik: I don't think this is related to doctypes, so I would keep it on banks, at least for now.

src/utils/triggers.js
ptbrowne: Bonne idée

ptbrowne: https://sentry.cozycloud.cc/sentry/cozy-banks-0s/issues/59669/

PR #291, opened 2018-07-11
Use native selects on mobile

drazik:
The cozy-ui select has a problem on Android: when we click on the (transparent) overlay to close the select without interacting with elements on the page, the first time acts like if there was no overlay, then everything is normal on next tries.

Since we have been searching why we have this behaviour for a long time, without any success, and since we have another problem with the cozy-ui select, on iOS this time (the click handler is not triggered everywhere), I think it's better for us to just use native selects on mobile.

The advantages are:

* Better UX : options are large enough to click on it easily, whereas cozy-ui select's options are very small
* No more bugs : Android and iOS bugs are past

The disadvantage is: the design of the options is platform-dependent. But to me it's actually an advantage, because the UX is way better with it.

### DIFF
--- a/src/components/Select/styles.styl
+++ b/src/components/Select/styles.styl
@@ -4,14 +4,17 @@
 select-date-height-small=2.75rem
 
 .Select
+    position relative
+
+.Select__control
     appearance none
     border 0
     outline 0
     background transparent
+    color inherit
 
     // So that you can click on arrow icon to open the select
     padding-right 2rem
-    margin-right -2rem
     cursor pointer
 
 .Select::-ms-expand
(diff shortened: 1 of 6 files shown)


src/components/Select/styles.styl
```diff
@@ -23,14 +26,13 @@ select-date-height-small=2.75rem
     cursor pointer
     color coolGrey
 
+.Select .Select__Icon
+    position absolute
+    top 50%
+    transform translateY(-50%)
+    right 0.875rem
```
ptbrowne: It seems weird to me to have the arrow in absolute position. Maybe we could use flexbox to center the icon vertically ?

drazik: We need it to be on top of the select, so clicking on it actually clicks on the select. If we use flexbox, they will be next to each other, no ? Do you have something in mind ?

ptbrowne: Ah yes OK I see.

PR #298, opened 2018-07-17
feat: Improvements on Onboarding 🚀

kosssi:
I advise you to review each commit separately.

On browser
capture d ecran 2018-07-17 a 19 48 34

On mobile
capture d ecran 2018-07-17 a 18 57 02


src/ducks/onboarding/EnsureHasAccounts.jsx
```diff
@@ -27,41 +33,75 @@ const LayoutContent = props => (
...
+      hasParameter(location.query, 'onboarding')
+    ) {
+      return (
+        <LayoutContent>
+          <Onboarding />
+        </LayoutContent>
+      )
+    }
+
+    this.stopInterval()
```
drazik: I find it weird to execute this function in the `render` method.

I think doing the following would be better : 

* `componentDidMount`: if there are no accounts, `startInterval`
* `componentDidUpdate`: if there are some accounts, `stopInterval`
* and store the interval id outside of the state, since it's not used in the render method (put it in `this.intervalID`, for example)

PR #303, opened 2018-07-19
fix: First fetch accounts😇

kosssi:


src/ducks/onboarding/EnsureHasAccounts.jsx
```diff
@@ -34,7 +34,11 @@ const LayoutContent = props => (
  * Replaces its children by Onboarding if we have no accounts
  */
 class EnsureHasAccounts extends Component {
-  intervalId = false
+  constructor(props) {
+    super(props)
+    this.intervalId = false
+    props.fetchAccounts()
```
drazik: You should put it in `componentDidMount`

PR #315, opened 2018-07-31
feat: integrate new cozy-client

ptbrowne:


src/ducks/settings/GroupSettings.jsx
```diff
@@ -49,18 +52,20 @@ class GroupSettings extends Component {
...
+  toggleAccount = async (accountId, group, enabled) => {
+    const accounts = group.accounts
     if (enabled) {
       accounts.addById(accountId)
     } else {
       accounts.removeById(accountId)
     }
+    if (!group.id) {
+      this.updateOrCreate(dehydrateDoc(group))
+      this.setState({})
```
drazik: What's the purpose of this `setState` ?

src/utils/client.js
```diff
@@ -21,6 +21,12 @@ export const getClient = async () => {
   const root = document.querySelector('[role=application]')
   const data = root.dataset
 
+  const protocol = window.location.href.split('://')[0]
```
drazik: You can access the protocol directly with `window.location.protocol`. It just includes a trailing `:`, so you can omit it on line 26 while initializing `cozy.client`

src/ducks/settings/GroupSettings.jsx
ptbrowne: It should not be necessary after the work I did on the queries, I am going to try thanks !

ptbrowne:
> Is it planned/possible to replace cozy-client-js by cozy-client ? So we have only one client in only one version.

Yes it is planned :) https://github.com/cozy/cozy-client/issues/67

PR #317, opened 2018-08-01
feat: Remove stateful on balance lower notification 🛎 

kosssi:


src/ducks/notifications/services.js
```diff
@@ -60,15 +60,20 @@ const getEnabledNotificationClasses = config => {
     const enabled = klassConfig && klassConfig.enabled
     if (!enabled) {
       log('info', Klass.name + ' is not enabled')
+    } else {
+      log('info', Klass.name + ' is enabled')
```
drazik: You can invert this condition to make it more readable : `if (enabled) { ... } else { ... }`

PR #326, opened 2018-08-06
Balance history chart

drazik:
It uses static data for now. The balance history is passed in a rolling window algorithm with a size of 15. The result can be seen on the Balance page or in the styleguide

image

### DIFF
--- a/src/components/Chart/LineChart.jsx
+++ b/src/components/Chart/LineChart.jsx
@@ -1,0 +1,113 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import * as d3 from 'd3'
+
+class LineChart extends Component {
+  componentDidMount() {
+    this.createChart()
+  }
+
+  componentDidUpdate() {
+    this.createChart()
+  }
+
+  empty() {
+    this.root.innerHTML = ''
+  }
+
+  createChart() {
+    this.empty()
+
+    const {
+      data,
+      width,
+      height,
+      margin,
+      lineWidth,
+      lineColor,
+      nbTicks = data.length,
+      tickFormat,
+      tickPadding,
+      xScale,
+      yScale
+    } = this.props
+
+    const innerWidth = width - margin.left - margin.right
+    const innerHeight = height - margin.top - margin.bottom
+
+    const x = xScale().range([0, innerWidth])
+    const y = yScale().range([innerHeight, 0])
+
+    x.domain([d3.min(data, d => d.x), d3.max(data, d => d.x)])
+    y.domain([d3.min(data, d => d.y), d3.max(data, d => d.y)])
+
+    const svg = d3
+      .select(this.root)
+      .append('g')
+      .attr('transform', `translate(${margin.left}, ${margin.top})`)
+
+    const line = d3
+      .line()
+      .x(d => x(d.x))
+      .y(d => y(d.y))
+
+    svg
+      .append('path')
+      .datum(data)
+      .attr('stroke', lineColor)
+      .attr('stroke-width', lineWidth)
+      .attr('fill', 'none')
+      .attr('d', line)
+
+    const xAxisGenerator = d3.axisBottom(x).ticks(nbTicks)
+
+    if (tickPadding !== undefined) {
+      xAxisGenerator.tickPadding(tickPadding)
+    }
+
+    if (tickFormat) {
+      xAxisGenerator.tickFormat(tickFormat)
+    }
+
+    svg
+      .append('g')
+      .attr('transform', `translate(0, ${innerHeight})`)
+      .call(xAxisGenerator)
+  }
+
+  render() {
+    const { width, height } = this.props
+
+    return (
+      <svg ref={node => (this.root = node)} width={width} height={height} />
+    )
+  }
+}
+
+LineChart.propTypes = {
+  width: PropTypes.number.isRequired,
+  height: PropTypes.number.isRequired,
+  margin: PropTypes.shape({
+    top: PropTypes.number,
+    bottom: PropTypes.number,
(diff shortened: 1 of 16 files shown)


src/ducks/balance/History.jsx
```diff
@@ -1,32 +1,87 @@
...
 
 class History extends Component {
+  getCurrentBalance() {
+    return sumBy(this.props.accounts, a => a.balance)
+  }
+
+  getBalanceHistory() {
+    const { accounts, transactions } = this.props
+    const balanceHistories = getBalanceHistories(accounts, transactions)
+    const balanceHistory = Object.entries(balanceHistories.all)
```
ptbrowne:
```
import { sortBy } from 'lodash
const balanceHistory = sortBy(Object.entries(balanceHistories.all), x => x[0])
```

ptbrowne: I don't understand what is this first element, is it the date ? Maybe this sort logic could be extracted into a function

```
const sortBalanceHistoryByDate = sortBy(balanceHistories, x => x[0])
```

src/ducks/balance/History.jsx
```diff
@@ -1,32 +1,87 @@
...
+        if (a[0] > b[0]) {
+          return -1
+        }
+
+        return 0
+      })
+      .map(([date, balance]) => ({
+        x: parseDate(date),
+        y: balance
+      }))
```
ptbrowne: All this work is done at each render, I think it should be done only when the data changes.

src/ducks/balance/History.jsx
```diff
@@ -11,6 +11,8 @@ import { getBalanceHistories } from './helpers'
 import styles from './History.styl'
 
 class History extends Component {
+  INTERVAL_BETWEEN_TICKS = 57
```
ptbrowne: `DAY_WIDTH` is clearer IMHO

src/ducks/balance/Balance.styl
```diff
@@ -95,3 +96,7 @@
     .Balance__bank
         maxed-flex-basis 60%
 
+    .Balance__history
+        margin-left -1rem
+        margin-right -1rem
+        margin-top -1rem
```
ptbrowne: What is those -1rem ? Is it to cancel the padding of the container ? I understand why this is done here but I think it is not clear that it is coupled to the Layout of Cozy-UI.

I think regrouping the two declarations  would make it clear that they are working on the same "feature" (cancelling the paddings of the Layout container).

```
.Balance__history
    margin-left -2rem
    margin-right -2rem

    +small-screen()
        &
	        margin-left -1rem
	        margin-right -1rem
	        margin-top -1rem
```

Then I wonder if it could be something shipped in Cozy-UI.

```
.Balance__history
  @extends $Layout__fullwidth
```

drazik: Actually I found [this solution](https://css-tricks.com/full-width-containers-limited-width-parents/), but it doesn't work on the large layout with the fixed sidebar on the left. Maybe cozy-ui can ship something I will look at it.

src/components/Chart/LineChart.jsx
```diff
@@ -0,0 +1,136 @@
...
+import PropTypes from 'prop-types'
+import * as d3 from 'd3'
+
+class LineChart extends Component {
+  componentDidMount() {
+    this.createChart()
+  }
+
+  componentDidUpdate() {
+    this.createChart()
```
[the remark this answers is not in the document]
drazik: :heavy_check_mark:

src/components/Chart/LineChart.jsx
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  empty() {
+    this.root.innerHTML = ''
+  }
+
+  createChart() {
+    this.empty()
```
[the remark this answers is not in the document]
drazik: :heavy_check_mark:

src/components/Chart/LineChart.jsx
```diff
@@ -0,0 +1,136 @@
...
+      .append('g')
+      .attr('transform', `translate(${margin.left}, ${margin.top})`)
+
+    const line = d3
+      .line()
+      .x(d => x(d.x))
+      .y(d => y(d.y))
+
+    svg
+      .append('path')
```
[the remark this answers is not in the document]
drazik: :heavy_check_mark:

src/components/Chart/LineChart.jsx
```diff
@@ -0,0 +1,136 @@
...
+    this.empty()
+
+    const {
+      data,
+      width,
+      height,
+      margin,
+      lineWidth,
+      lineColor,
+      nbTicks = data.length,
```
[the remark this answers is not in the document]
drazik: :heavy_check_mark:

src/ducks/balance/History.styl
```diff
@@ -0,0 +1,20 @@
...
+
+.History__currentBalanceCurrency
+    color inherit
+
+.History__subtitle
+    margin-top 0.25rem
+    text-transform uppercase
+    font-size 0.85rem
+    font-weight bold
+    color #a2c4f9
```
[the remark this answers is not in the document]
drazik: This is specific to the chart. I don't think it's intended to land in cozy-ui until we have charts elsewhere.

src/ducks/balance/History.jsx
drazik: :heavy_check_mark:

src/ducks/balance/Balance.styl
ptbrowne: Ah, I knew there was something for this pattern. Check src/styles/mixins.styl ;)

PR #328, opened 2018-08-07
Transactions balances chart

drazik:
This adds the balances history chart on the transactions page (under a flag). It shows the last 60 days only.


src/components/Chart/LineChart.jsx
```diff
@@ -11,10 +11,19 @@ class LineChart extends Component {
...
   }
 
+  getRootWidth() {
+    if (typeof this.props.width === 'number') {
+      return this.props.width
+    }
+
+    const { width } = getComputedStyle(this.root)
+
+    return +width.slice(0, -2)
```
ptbrowne: ```parseInt(width)``` would also work and would be clearer IMH

```
> parseInt('12px')
12
```

src/components/Chart/LineChart.jsx
```diff
@@ -150,7 +161,13 @@ class LineChart extends Component {
 }
 
 LineChart.propTypes = {
-  width: PropTypes.number.isRequired,
+  data: PropTypes.arrayOf(
+    PropTypes.shape({
+      x: PropTypes.any,
+      y: PropTypes.any
```
ptbrowne: 👍 

PropTypes.number for x and y

src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -266,9 +313,27 @@ class TransactionsPage extends Component {
...
+    const chartNbTicks = chartData.length
+    const chartIntervalBetweenPoints = 10
+
     return (
-      <div className={styles.TransactionPage}>
+      <div className={styles.TransactionPage} ref={node => (this.root = node)}>
         {subcategoryName ? <BackButton /> : null}
+        {flag('balance-history') && (
+          <History
+            className={styles2.Balance__history}
```
kosssi: J'ai l'impression que le style devrait être directement dans le composant `History` tu en penses quoi ?

src/components/Chart/LineChart.jsx
drazik: TIL `parseInt` stops parsing when it finds a character that is not a number :+1:

src/components/Chart/LineChart.jsx
drazik: Actually it can also be a `Date` (like in our current case). I don't know if we can really enforce more than just a shape `{ x, y }`

src/ducks/transactions/TransactionsPage.jsx
drazik: Yep :+1:

PR #330, opened 2018-08-07
Use cozy-flags

drazik:
Replace the internal flags util by cozy-flags everywhere


src/main.jsx
```diff
@@ -10,7 +10,7 @@ import 'number-to-locale-string'
 
 import { setupHistory } from 'utils/history'
 import { getClient } from 'utils/client'
-import 'utils/flag'
+import 'cozy-flags/dist/flag'
```
ptbrowne: `import flag from 'cozy-flags'` ?

drazik: Actually we don't use `flag` in this file. We just want to execute some of the code contained in cozy-flags : https://github.com/cozy/flags/blob/master/src/flag.js#L45-L75

`import flag from 'cozy-flags'` would execute this code, but we will never use `flag` in this file :thinking:

kosssi: Maybe an `init` function, I'm not fan of :

```
import { init } from 'flags'

init()
```

drazik: Oh that's just some code that should be managed by the application. Maybe not all apps want to set the `switcher` flag to true just because they use cozy-flags ?

PR #333, opened 2018-08-08
Balance history UI

drazik:
There are two distincts UIs : 

One for the balances page, that shows the X axis and the current balance:

image

And one for the transactions page, that only shows the line: 

image

Some UI things that are related to interactions are not present in this PR, but will be in a next one.

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -12,6 +12,7 @@ import { connect } from 'react-redux'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { or } from 'airbnb-prop-types'
+import * as d3 from 'd3'
 
 import { withRouter } from 'react-router'
 import { translate, Button, Icon, withBreakpoints } from 'cozy-ui/react'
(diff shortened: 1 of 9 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -380,10 +381,19 @@ class Balance extends React.Component {
...
+              height: 103,
+              margin: {
+                top: 10,
+                bottom: 35,
+                left: 16,
+                right: 16
+              },
+              showAxis: true,
+              axisMargin: 10,
+              tickFormat: d3.timeFormat('%b')
```
ptbrowne: Minor: I would extract the d3.timeFormat outside the component

src/ducks/balance/Balance.jsx
```diff
@@ -380,10 +381,19 @@ class Balance extends React.Component {
...
               data: chartData,
-              width: chartNbTicks * chartIntervalBetweenPoints
+              width: chartNbTicks * chartIntervalBetweenPoints,
+              height: 103,
+              margin: {
+                top: 10,
+                bottom: 35,
+                left: 16,
+                right: 16
+              },
```
ptbrowne: Minor: I would extract those margins outside

src/ducks/balance/Balance.jsx
drazik: I don't understand where you would put it. Can you tell me more ?

ptbrowne: sure :) at the top of the component

```
const TICK_FORMAT = d3.timeFormat('%b')

tickFormat: TICK_FORMAT
```

PR #334, opened 2018-08-08
fix: wepback providing cozy-bar

ptbrowne:
There was a risk that the ProvidePlugin from webpack.dev would
override the one from webpack.mobile, meaning that the bar would
be in non mobile mode even when we were building in mobile mode.

It is not the case in production mode since the webpack.production
does not define the ProvidePlugin.


kosssi:
Il y a un changement à faire dans les snapshots, comment ça se fait ?

ptbrowne:
Le test dépend de la date, ca n'a pas à voir avec cette PR.

PR #339, opened 2018-08-10
chore: Use cheap-module-eval-source-map

drazik:
I propose to use cheap-module-eval-source-map when building for dev to get lines in source files in error stacktrace while developing.


ptbrowne:
When using eval, did you have to disable the CSPs ? I am OK to do that since we already have to do that for hot reload but we need to be aware of that.

PR #340, opened 2018-08-10
Chart interation

drazik:
This adds some interaction on the line chart : 

peek 10-08-2018 20-06

As you can see, it lacks some things:

* The tooltip doesn't properly align when it reaches the edges of its container
* The little triangle is commented out in CSS file. Actually it also need some calculation to align properly when the tooltip reaches the edges of its container
* It doesn't animate on mount, so it's visible before the line and other elements

Feel free to edit this PR directly to add missing features :pray:

### DIFF
--- a/src/components/Chart/LineChart.jsx
+++ b/src/components/Chart/LineChart.jsx
@@ -4,6 +4,8 @@ import * as d3 from 'd3'
 import { maxBy, sortBy } from 'lodash'
 
 class LineChart extends Component {
+  dragging = false
+
   componentDidMount() {
     this.createChart()
   }
@@ -54,6 +56,8 @@ class LineChart extends Component {
     this.x = xScale().range([0, innerWidth])
     this.y = yScale().range([innerHeight, 0])
 
+    const drag = d3.drag()
+
     this.svg = d3
       .select(this.root)
       .append('g')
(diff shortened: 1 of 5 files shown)


src/components/Chart/LineChart.jsx
```diff
@@ -238,7 +386,29 @@ LineChart.defaultProps = {
...
+class Tooltip extends Component {
+  render() {
+    const { children, x } = this.props
+
+    return (
+      <div
+        ref={node => (this.node = node)}
+        className={styles.LineChartTooltip}
+        style={{
+          left: x + 'px'
```
ptbrowne: A `transform: translateX` would not trigger a relayout here.

https://csstriggers.com/transform

> left
Changing left alters the geometry of the element. That means that it may affect the position or size of other elements on the page, both of which require the browser to perform layout operations.
Once those layout operations have completed any damaged pixels will need to be painted and the page must then be composited together.

> transform
Changing transform does not trigger any geometry changes or painting, which is very good. This means that the operation can likely be carried out by the compositor thread with the help of the GPU.

src/components/Chart/LineChart.jsx
```diff
@@ -176,21 +237,103 @@ class LineChart extends Component {
...
+    this.pointLine
+      .attr('x1', x)
+      .attr('y1', 0)
+      .attr('x2', x)
+      .attr('y2', this.props.height - this.props.margin.bottom)
+  }
+
+  setTooltipContent(content) {
+    this.tooltip.html(content)
+  }
```
ptbrowne: This is unused.

src/components/Chart/LineChart.jsx
```diff
@@ -12,6 +20,10 @@ class LineChart extends Component {
     this.updateData()
   }
 
+  componentWillUnmount() {
+    this.tooltip.remove()
```
ptbrowne: This is unused

ptbrowne:
Thanks Drazik ! This is very clear :)

PR #354, opened 2018-08-16
feat: Add description on notifications 💬

kosssi:
I add header part on `TogglePane.jsx` to add a description on notifications.

You can see the result:

capture d ecran 2018-08-16 a 16 27 01

### DIFF
--- a/src/ducks/settings/TogglePane.jsx
+++ b/src/ducks/settings/TogglePane.jsx
@@ -6,6 +6,10 @@ import { flowRight as compose } from 'lodash'
 import ToggleRow from './ToggleRow'
 import { getSettingsFromCollection } from './helpers'
 
+const ToogleDescription = ({ children }) => (
+  <p className="u-coolGrey">{children}</p>
+)
+
 class TogglePane extends Component {
   onToggle = (setting, checked) => {
     const { settingsCollection, settingsKey } = this.props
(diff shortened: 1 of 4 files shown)


src/ducks/settings/TogglePane.styl
```diff
@@ -0,0 +1,5 @@
+.TogglePane__main
+    margin-bottom 3rem
+
+.TogglePane__description
+    color coolGrey
```
ptbrowne: We should have a description label in cozy ui for that.

src/ducks/settings/TogglePane.styl
```diff
@@ -0,0 +1,5 @@
+.TogglePane__main
+    margin-bottom 3rem
```
ptbrowne: Why do we need another class for this ? Can we use the spacing styles from cozy-ui ?

src/ducks/settings/TogglePane.jsx
```diff
@@ -26,12 +27,22 @@ class TogglePane extends Component {
...
   render() {
-    const { rows, settingsCollection, settingsKey, t, title } = this.props
+    const {
+      rows,
+      settingsCollection,
+      settingsKey,
+      t,
+      title,
+      description
+    } = this.props
```
ptbrowne: This is starting to get ugly, I think we should be able to refactor to have less properties here.

src/ducks/settings/TogglePane.styl
ptbrowne: I don't understand why we need such a space. Maybe it  should be the job of an utility class to do that ? The spacing classes from Cozy-UI do not do the job ?

src/ducks/settings/TogglePane.styl
kosssi: have you a link to the documentation ?

PR #364, opened 2018-08-21
feat: Use cozy-interapp lib 🚜

kosssi:
To remove cozy-client-js we should extract intent in a new lib.
This lib is cozy-interapp, we change instantiation to pass cozy-client

### DIFF
--- a/src/components/FileIntentDisplay.jsx
+++ b/src/components/FileIntentDisplay.jsx
@@ -1,20 +1,15 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import { Intents } from 'cozy-interapp'
 import { withClient } from 'cozy-client'
 
 import FullscreenIntentModal from 'components/FullscreenIntentModal'
 
 class FileIntentDisplay extends Component {
-  constructor(props) {
-    super(props)
-    this.intents = new Intents({ client: props.client })
-  }
-
   createIntent = () => {
     const id = this.props.fileId
     const doctype = 'io.cozy.files'
-    return this.intents.create('OPEN', doctype, { id })
+    const cozyClient = this.props.client
+    return cozyClient.intents.create('OPEN', doctype, { id })
   }
 
   showModal = () => {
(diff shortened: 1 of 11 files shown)


ptbrowne:
Pourquoi on passe `fetchJSON` et pas le client directement ?

Pourquoi l'application a besoin de savoir que les intents utilisent `fetchJSON` ? IMHO c'est une abtraction qui *leak*.

src/components/FileIntentDisplay.jsx
```diff
@@ -1,13 +1,20 @@
...
 import PropTypes from 'prop-types'
+import { Intents } from 'cozy-interapp'
+import { withClient } from 'cozy-client'
+
 import FullscreenIntentModal from 'components/FullscreenIntentModal'
 
 class FileIntentDisplay extends Component {
+  constructor(props) {
+    super(props)
+    this.intents = new Intents({ client: props.client })
```
kosssi: New instantiation of Intents

test/AppLike.jsx
```diff
@@ -1,13 +1,19 @@
...
 import store from 'test/store'
+import getClient from 'test/client'
 
-const AppLike = ({ children, store }) => (
-  <I18n lang={'en'} dictRequire={() => langEn}>
-    <Provider store={store}>{children}</Provider>
-  </I18n>
+const AppLike = ({ children, store, client }) => (
+  <Provider store={store}>
+    <CozyProvider client={client || getClient()}>
```
kosssi: Add CozyProvider to have client on context

test/client.js
```diff
@@ -0,0 +1,16 @@
+import CozyClient from 'cozy-client'
```
kosssi: maybe this file needs to be moved in cozy-client

ptbrowne: Yep. Though I think that in the tests from Cozy-client, we already have stuff like this where we mock much more. So I think we're going to do this in cozy-client (move code inside cozy-client and export it) and then we're going to be able to import it here. I hope I am understandable.

PR #365, opened 2018-08-23
Show virtual groups

drazik:
Virtual groups were not shown anymore after we refactored the code to use the new cozy-client. This fixes it.

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -31,6 +31,7 @@ import { ACCOUNT_DOCTYPE, GROUP_DOCTYPE, SETTINGS_DOCTYPE } from 'doctypes'
 import History from './History'
 import historyData from './history_data.json'
 import { getBalanceHistories } from './helpers'
+import { buildVirtualGroups } from 'ducks/groups/helpers'
 import sma from 'sma'
 import { parse as parseDate, format as formatDate } from 'date-fns'
 
@@ -334,7 +335,7 @@ class Balance extends React.Component {
     }
 
     const accounts = accountsCollection.data
-    const groups = groupsCollection.data
+    const groups = [...groupsCollection.data, ...buildVirtualGroups(accounts)]
     const settings = settingsCollection.data
 
     const accountsSorted = sortBy(accounts, ['institutionLabel', 'label'])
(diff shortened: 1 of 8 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -30,7 +30,7 @@ import { getAccountInstitutionLabel } from 'ducks/account/helpers'
 import { ACCOUNT_DOCTYPE, GROUP_DOCTYPE, SETTINGS_DOCTYPE } from 'doctypes'
 import History from './History'
 import historyData from './history_data.json'
-import { getBalanceHistories } from './helpers'
+import { getBalanceHistories, generateVirtualGroups } from './helpers'
```
ptbrowne: Suggestion : `s/generate/build/` ?

generate gives me the impression that they will stay on the server I don't know why. build is shorter also :)

src/ducks/balance/helpers.js
```diff
@@ -77,3 +78,21 @@ const sumAllBalancesByDate = accountsBalances => {
 
   return balances
 }
+
```
ptbrowne: Debate : I see the virtual groups more something that would belong in `ducks/accounts` than in balance.

drazik: I think we can go further and create a new `groups` duck, no?

ptbrowne: Yes maybe :) As you like :)

src/ducks/groups/helpers.spec.js
```diff
@@ -0,0 +1,36 @@
...
+
+    const virtualGroups = buildVirtualGroups(accounts)
+    const expected = [
+      {
+        _id: 'checkings',
+        _type: 'io.cozy.bank.groups',
+        label: 'checkings',
+        accounts: {
+          data: accounts
+        },
```
ptbrowne: Since this is an internal structure of `cozy-client`, I think it would be better if it was hidden in a function, maybe as a helper in utils/cozy-client that could later be exported in `cozy-client`. The goal being that at some point if cozy-client decides to change the structure of its relationships, we should have our test that fail. WDYT ?

ptbrowne: This is related to https://github.com/cozy/cozy-banks/blob/master/src/utils/client.js#L62

drazik: You think about a function like `associateDocuments(originalDocument, associationName, associatedDocuments)` ?

PR #378, opened 2018-08-31
Clean on logout

drazik:
Clean the app on logout:

* Revoke client
* Call logout on client so links are unregistered
* Delete token from local redux store
* Recreate links on re-login


drazik:
Currently WIP as we need to merge https://github.com/cozy/cozy-client/pull/124 and use it here before merging.

src/ducks/authentication/MobileRouter.jsx
```diff
@@ -24,11 +26,31 @@ import {
...
+    // eslint-disable-next-line
+    console[level] && console[level](...logs)
+  }
+}
+
+export const onLogout = async (store, cozyClient, replaceFn) => {
+  try {
+    await stopPushNotifications()
+
+    logIfDev('info', 'Stopped push notifications')
```
ptbrowne: Nit: I think this cannot be stripped from production code. In React code, they directly write `if (__DEV___)` and have code to strip out `warning` function calls.

src/ducks/authentication/MobileRouter.jsx
```diff
@@ -64,6 +86,7 @@ const withAuth = Wrapped => (props, { store }) => {
       }
     }
 
+    cozyClient.login()
     await registerPushNotifications(cozyClient, clientInfos)
```
ptbrowne: Thought: Push notifications could be extracted from cozy-banks as a plugin for cozy-client

src/ducks/authentication/MobileRouter.jsx
```diff
@@ -64,6 +86,7 @@ const withAuth = Wrapped => (props, { store }) => {
       }
     }
 
+    cozyClient.login()
```
ptbrowne: We do not pass the `token` to the login function ?

src/ducks/authentication/MobileRouter.jsx
ptbrowne: Yes. This can be done later since it is not the subject of the PR.

src/ducks/authentication/MobileRouter.jsx
drazik: Actually we already retrieve and set the right token from the app's persisted state in https://github.com/cozy/cozy-banks/blob/master/src/ducks/client/mobile.js#L17-L20

So for me it seems OK.

src/ducks/authentication/MobileRouter.jsx
drazik: I will do it here since I added these logs in this PR.

src/ducks/authentication/MobileRouter.jsx
ptbrowne: If it's already done there, shouldn't we do the login() method there instead of in the router ?

drazik: `getClient` is executed only when we start the app. We want to execute `login` when the user authenticates. It may occurs on app start, but also during the life of the app, for example if the user logout then login on another instance. That's why the `login` must be in the `onAuthentication` function inside the `MobileRouter` (it is executed whenever the user authenticates, whereas the client is created only on app start to be provided to the `CozyProvider`).

ptbrowne:
LGTM. I still have trouble with the way MobileRouter and getClient etc... are tangled but let's move on.

PR #398, opened 2018-09-10
feat: PouchLink pauses/resumes on app focus/blur

ptbrowne:
To save battery when the application is not in foreground, the
replication is stopped when the application does not have the 
focus.
Maybe this could done directly in PouchLink ?


src/ducks/client/links.js
```diff
@@ -1,4 +1,4 @@
-/* global __TARGET__ */
+/* global __TARGET__, __DEV__ */
```
drazik: If I'm not mistaken, the global is `__DEVELOPMENT__`, as seen in https://github.com/cozy/cozy-banks/blob/master/config/webpack.config.dev.js#L26

PR #411, opened 2018-09-17
Relationships account reimbs

ptbrowne:
⚠️Tests need https://github.com/cozy/cozy-client/pull/158 to be merged.

Too big a PR, I am sorry for that :/

The main bulk of this PR is to use relationships for accounts and reimbursements instead of using our own hydratation method. This fixes the bug where one view had the transaction hydrated and another one had not. Moreover, in the future, it will allow us to hydrate objects more precisely since we will be able to hydrate only where we need it (typically on the innermost component).

Also : 

- Removed custom methods for cozy-client (utils/client is almost empty now 🎉 ) :
  + Use queryConnect from cozy-client
  + Use the new makeNewDocument from cozy-client to create new document with relationships already instantiated (useful for creating a new group)
  + Dehydrate is now a method from cozy-client (since it knows about the schema)


src/actions/accounts.js
```diff
@@ -11,10 +11,12 @@ const removeAccountFromGroup = (group, account) => {
 }
 
 const getStackCollection = doctype => {
+  const links = getLinks()
   return links.stack.client.collection(doctype)
 }
 
 const deleteOrphanOperations = async (client, account) => {
+  const links = getLinks()
```
drazik: I don't understand why we always want to use the stack link here. We can't use the pouch synchronization for this ?

ptbrowne: Maybe we could but here the main thing was to fix the wrongly used links.

PR #417, opened 2018-09-19
Get app slug from stack

CPatchane:
Necessary to use the new cozy-bar navigation
and the home app detection


ptbrowne:
Thanks ! What is it used for ?

CPatchane:
The cozy-bar is using it to know the current app served. Without that nothing will be broken but the app will miss some UI features like a bold name to say that is the current app or the home way display if the application is set as the home of the Cozy.

PR #420, opened 2018-09-24
chore(deps): Update cozy-bar to v6.1.3

drazik:
We come from v5.1.0. So we must be careful if something breaks. According to the [changelog](https://github.com/cozy/cozy-bar/releases/tag/v6.0.0), we only need to use `name_prefix` in our manifest, which we alreaydy do (see https://github.com/cozy/cozy-banks/blob/master/manifest.webapp#L41).

Ping @CPatchane and @gregorylegarec can you have a look and tell us if we miss something ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "classnames": "^2.2.0",
     "cordova": "^7.1.0",
     "cozy-authentication": "1.3.0",
-    "cozy-bar": "^5.1.0",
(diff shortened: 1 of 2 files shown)


CPatchane:
The two things (`name_prefix` and `{{.AppSlug}}` in `.ejs`) are already done in Banks so I think you are good for this version 👍

drazik:
Thanks @CPatchane

PR #434, opened 2018-10-17
Use React

ptbrowne:
Use react@16.2 if `process.env.USE_REACT` is true during build.


.eslintrc
```diff
@@ -3,5 +3,10 @@
   "extends": ["cozy-app/react"],
   "globals": {
     "jest": true
+  },
+  "settings": {
+    "react": {
+      "version": "15.6.2", // React version, default to the latest React stable release
```
drazik: Why not `16.5.2` ? Actually it seems you inverted the `5` and the `6`.

ptbrowne: Thanks, it is 16.2, I had it before upgrading to 16, and cozy-ui actually needs the 16 for the Portal to works. 16.2 and not 16.3 not to have the lifecycle hooks deprecations (componentDidUpdate -> UNSAFE_componentDidUpdate in 16.3)

src/ducks/settings/TogglePane.jsx
```diff
@@ -82,9 +81,4 @@ TogglePane.propTypes = {
   saveDocument: PropTypes.func.isRequired
 }
 
-export default compose(
-  translate(),
-  // We keep `connect` just so the component still receive the `dispatch` function as prop
-  // This has to be removed when we handle the mutations with the new cozy-client
-  connect()
-)(TogglePane)
+export default compose(translate())(TogglePane)
```
kosssi: Il y a plus besoin de compose du coup.

```suggestion
export default translate()(TogglePane)
```

package.json
```diff
@@ -122,11 +122,11 @@
     "postcss-loader": "3.0.0",
     "pretty": "2.0.0",
     "prop-types": "15.6.2",
-    "react": "15.6.2",
-    "react-addons-test-utils": "15.6.2",
-    "react-dom": "15.6.2",
+    "react": "16.2",
+    "react-addons-test-utils": "^16.0.0-alpha.3",
```
kosssi: renovate repassera par la puisqu'il fixe les versions donc sans `^`

.eslintrc
drazik: :+1:

PR #443, opened 2018-10-26
chore: Fix build:mobile script

drazik:
The script failed because the target was not detected as `mobile` but `browser`.

When did we make some changes in the target detection behavior ?


ptbrowne:
Fixed in https://github.com/cozy/cozy-banks/commit/7f105bd26c8424e398c84c9c834538b6865ed83d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 . Sorry...

PR #449, opened 2018-10-30
Perf memory

ptbrowne:
Trying to decrease memory usage.

### DIFF
--- a/src/ducks/transactions/Transactions.jsx
+++ b/src/ducks/transactions/Transactions.jsx
@@ -43,6 +43,7 @@ import styles from './Transactions.styl'
 import { InfiniteScroll, ScrollRestore } from './scroll'
 import TransactionModal from './TransactionModal'
 import { isIOSApp } from 'cozy-device-helper'
+import PropTypes from 'prop-types'
 
 const sDate = styles['bnk-op-date']
 const sDesc = styles['bnk-op-desc']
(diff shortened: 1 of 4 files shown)


src/ducks/transactions/Transactions.jsx
```diff
@@ -283,6 +283,11 @@ class TransactionsD extends React.Component {
   constructor(props) {
     super(props)
     this.scrollSpy = new ScrollSpy(this.getScrollingElement)
+    this.handleScroll = (isIOSApp() ? debounce : throttle)(
+      this.handleScroll.bind(this),
+      300,
+      { leading: false, trailing: true }
+    )
```
drazik: For my personal culture, what's different between doing this in the constructor VS in a class property ?

src/ducks/transactions/actions/index.js
```diff
@@ -10,26 +10,33 @@ import AttachAction from './AttachAction'
...
-      matchingActions.push(action)
+  for (const [actionName, action] of Object.entries(actions)) {
+    try {
+      const matching = await action.match(transaction, actionProps)
+      if (matching) {
+        matchingActions.push(action)
+      }
+    } catch (e) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.log('action failed', actionName) // eslint-disable-line no-console
```
drazik:
```suggestion
        console.log('action matching failed', actionName) // eslint-disable-line no-console
```

src/ducks/transactions/Transactions.jsx
ptbrowne: It's mainly for testing and mocking. It is impossible to mock class via property (as far as I tried) since the method is created is the constructor and not in the prototype.

PR #458, opened 2018-10-31
feat(service): Rename to onOperationOrBillCreate

drazik:
This way it's in sync with its trigger


ptbrowne:
Il y a aussi onOperationCreate dans `docs/categorization.md`

PR #460, opened 2018-10-31
fix(GroupSettings): Wrong toggle state

drazik:
We were not passing an id to `existsById`, thus the account was never found in the collection and the toggle never checked.


ptbrowne:
Can you add a test :) ?

drazik:
Here it is. Thanks again @ptbrowne :+1:

PR #461, opened 2018-10-31
fix: Save settings when Incomes toggle change

kosssi:
This PR fix 2 bugs:
- save correctly settings
- when Cozy don't have `io.cozy.bank.settings` it create it with default value


src/ducks/categories/CategoriesPage.jsx
```diff
@@ -47,7 +56,7 @@ class CategoriesPage extends Component {
     } = this.props
     const isFetching =
       isCollectionLoading(transactions) || isCollectionLoading(settings)
-    const { showIncomeCategory } = (settings.data && settings.data[0]) || false
+    const { showIncomeCategory } = this.getSettings()
```
drazik: Nit: since we need to retrieve the settings in other places, I think we should remove `getSettings` and just to `getSettingsFromCollection(this.props.settingsCollection)` where we need it.

PR #473, opened 2018-11-02
feat: Add pouch on web to develop easily

kosssi:
## Pouch On Web

If you want activate Pouch on web app, you need:
- a token
- launch your build with `FORCE_POUCH`

```
cozy-stack instances token-cli cozy.tools:8080 $(cat manifest.webapp | jq -r '[.permissions[] | .type] | join(" ")')
env FORCE_POUCH=true yarn watch
```

Launch your web application and in console javascript fill in the token and refresh the page:

```
flag('cozyToken', <your token>)
```

Now you can play with replication:

```
cozyClient.links[0].startReplication()
cozyClient.links[0].stopReplication()
```

### DIFF
--- a/config/webpack.target.browser.js
+++ b/config/webpack.target.browser.js
@@ -15,7 +15,8 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      __TARGET__: JSON.stringify('browser')
+      __TARGET__: JSON.stringify('browser'),
+      __POUCH__: JSON.stringify(process.env.FORCE_POUCH ? true : false)
     }),
     new HtmlWebpackPlugin({
       alwaysWriteToDisk: true,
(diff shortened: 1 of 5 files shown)


kosssi:
Ok I change it with a variable __POUCH__ ?

ptbrowne:
FORCE_POUCH ?

config/webpack.target.browser.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
   },
   plugins: [
     new webpack.DefinePlugin({
-      __TARGET__: JSON.stringify('browser')
+      __TARGET__: JSON.stringify('browser'),
+      __POUCH__: JSON.stringify(process.env.FORCE_POUCH ? true : false)
```
drazik: Are you sure that you want to get a string here ? It means that you can't just write `if (__POUCH__)`, since it will always be a non empty string, so truthy, no?

drazik: Oh, OK thanks !

PR #508, opened 2018-11-07
fix: Don't use color var in stylus func

GoOz:
Some cleaning in preparation for a future [release of Cozy-UI](https://github.com/cozy/cozy-ui/pull/641) with color variables as CSS custom properties.

We can't have custom properties in Stylus functions, it just breaks.
Most of `alpha()` color actually need ligter colors, not transparency. I think that was the case and that should do it.

Correct me if I'm wrong.

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/src/ducks/balance/Balance.styl
+++ b/src/ducks/balance/Balance.styl
@@ -24,15 +24,15 @@
     .Balance__row
         &:active
         &:focus
-            background alpha(dodgerBlue, 0.1)
+            background zircon
 
 .Balance__row--selected
     border-left 0.25rem dodgerBlue solid
-    background alpha(dodgerBlue, 0.1)
+    background zircon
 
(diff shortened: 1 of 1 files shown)


GoOz:
[Utiliser custom properties sur UI pour la personnalisation](https://trello.com/c/oHPjhtmr/336-utiliser-custom-properties-sur-ui-pour-la-personnalisation)

ptbrowne:
I think it's alright, but I am not sure without seeing the result. Can you post a before/after screenshot ?

kosssi:
@GoOz can you update description with screenshot ?

GoOz:
@kosssi I can't make the app Banks work locally and ultimately I don't know what `Balance_row`refers to. :(

PR #511, opened 2018-11-08
fix: Infinite scroll

drazik:
We messed with the DOM structure, breaking the infinite scroll. I didn't find precisely when. I remember we did some refactoring on the balance history chart someday, but can't find the PR. And I didn't want to lose too many time looking for it. If you think it's important to find it, I can do more research.

### DIFF
--- a/src/ducks/transactions/TransactionHeader.jsx
+++ b/src/ducks/transactions/TransactionHeader.jsx
@@ -8,8 +8,18 @@ import { Breadcrumb } from 'components/Breadcrumb'
 import { ConnectedSelectDates } from 'components/SelectDates'
 import { AccountSwitch } from 'ducks/account'
 import TransactionSelectDates from 'ducks/transactions/TransactionSelectDates'
+import { TransactionTableHead } from './Transactions'
+import flag from 'cozy-flags'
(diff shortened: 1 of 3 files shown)


kosssi:
I don't understand the fix, can you explain it ?

kosssi:
Changing the structure of the DOM makes it possible to re-run the scroll at the right place 👍

src/ducks/transactions/TransactionHeader.jsx
```diff
@@ -8,9 +8,19 @@ import { Breadcrumb } from 'components/Breadcrumb'
 import { ConnectedSelectDates } from 'components/SelectDates'
 import { AccountSwitch } from 'ducks/account'
 import TransactionSelectDates from 'ducks/transactions/TransactionSelectDates'
+import { TransactionTableHead } from './Transactions'
+import flag from 'cozy-flags'
+import HistoryChart from 'ducks/balance/HistoryChart'
```
ptbrowne: Au final on pourrait avoir le TransactionTableHead ici non ? Au lieu de l'importer ?

drazik: Yep, je fais cette modification

PR #515, opened 2018-11-09
fix: Autoreload data on onboarding

drazik:
The auto-refresh of the onboarding when there is no data but some triggers may be running was broken.

### DIFF
--- a/src/ducks/onboarding/EnsureHasAccounts.jsx
+++ b/src/ducks/onboarding/EnsureHasAccounts.jsx
@@ -25,10 +25,37 @@ const LayoutContent = props => (
   </Layout>
 )
 
+export function EnsureHasAccountsView(props) {
+  const { children, accounts, location } = props
+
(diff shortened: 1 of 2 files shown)


src/ducks/onboarding/EnsureHasAccounts.jsx
```diff
@@ -43,7 +53,7 @@ class EnsureHasAccounts extends Component {
     if (!client) {
       return
     }
-    client.query(accountsConn.query)
+    client.query(accountsConn.query(client), { as: accountsConn.as })
```
ptbrowne: C'est nécessaire pourquoi ça ?

ptbrowne:
Can you add a test ? 😊

src/ducks/onboarding/EnsureHasAccounts.jsx
drazik: Vu ensemble.

PR #521, opened 2018-11-13
chore: Remove preact

ptbrowne:
Not needed anymore ;)


package.json
```diff
@@ -177,8 +177,6 @@
     "number-to-locale-string": "1.2.0",
     "piwik-react-router": "0.8.2",
     "pouchdb-adapter-cordova-sqlite": "2.0.5",
-    "preact": "8.3.1",
-    "preact-compat": "3.18.4",
     "preact-portal": "1.1.3",
```
drazik: Wouldn't it be nice to also drop `preact-portal`? It's only used in `FullScreenIntentModal` (which is only used by `FileIntentDisplay`, which is used by nothing, so it seems we can just remove both) and in `TransactionModal` tests.

ptbrowne:
Thanks Drazik !

PR #527, opened 2018-11-14
chore: Use developement mode in dev

drazik:
Watch mode was very slow because we didn't set webpack `mode` to `development`. Since the default is `production`, it was applying unnecessary optimizations.

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -10,6 +10,7 @@ if (target !== 'mobile') {
 }
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


Crash--:
https://github.com/CPatchane/create-cozy-app/blob/master/packages/cozy-scripts/config/webpack.environment.dev.js#L39 :-° 

Do you have any plan to use cozy-script ?

drazik:
We want to use cozy-scripts but didn't begin this work yet.

PR #545, opened 2018-11-19
Update local categorization model

drazik:
The alpha model was previously fixed at `0.1`, but we saw that it doesn't handle all cases well. So François reworked it and came with some new expectations for this parameter:

* Never lesser than a minimum value (`0.1`)
* Never higher than a maximum value (`10`)
* Equals to `max smoothing parameter / number of unique categories` (where `max smoothing = 20`)

The initialization of the local model also needed a little rework:

* If the number of recategorized transactions is lower than a limit (`2`), give a higher weight to the recategorized transactions
* If the number of unique `manualCategoryId`s equals to `1`, add a fake transaction to the learn sample

Finally, we had a bug in which not only new transactions were passed in the categorization models.

### DIFF
--- a/src/ducks/categorization/services.js
+++ b/src/ducks/categorization/services.js
@@ -51,7 +51,11 @@ const getTransWithManualCat = async (transactions, index, limit, skip) => {
   return transactions
 }
 
-export const createLocalClassifier = (transactionsToLearn, options) => {
+export const createLocalClassifier = (
+  transactionsToLearn,
+  classifierOptions,
+  options
+) => {
   if (transactionsToLearn.length === 0) {
     localModelLog(
       'info',
@@ -60,14 +64,16 @@ export const createLocalClassifier = (transactionsToLearn, options) => {
     return null
   }
 
-  const classifier = bayes(options)
+  const classifier = bayes(classifierOptions)
 
   localModelLog('info', 'Learning from manually categorized transactions')
-  for (const transaction of transactionsToLearn) {
-    classifier.learn(
-      getLabelWithTags(transaction),
-      transaction.manualCategoryId
-    )
+  for (let i = 0; i < options.learnSampleWeight; ++i) {
+    for (const transaction of transactionsToLearn) {
+      classifier.learn(
+        getLabelWithTags(transaction),
+        transaction.manualCategoryId
+      )
+    }
   }
 
   return classifier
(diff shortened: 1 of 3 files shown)


src/ducks/categorization/services.js
```diff
@@ -34,24 +34,24 @@ export const getAlphaParameter = (
...
+  }
+  const results = await cozyClient.data.query(index, query)
+  transactions = [...transactions, ...results.docs]
+
+  if (results.next) {
+    return getTransWithManualCat(transactions, index, limit, skip + limit)
   }
 
+  return transactions
+}
```
ptbrowne: Nit: It's a bit cumbersome to have to call the function with an empty array at first. It can be useful to have a function that does this automatically.

```
const myCumbersomeRecursiveFunction = (res, i) => {
  if (res.length < 100) {
    res.push(i)
    return myCumbersomeRecursiveFunction(res, i+1)
  } else {
    return res
  }
}

const recursiveFunction = i => {
  return myCumbersomeRecursiveFunction({}, i)
} 
```

A default argument is also good but you need to reverse the arguments.

src/ducks/categorization/services.js
```diff
@@ -169,8 +175,21 @@ const localModel = async (classifierOptions, transactions) => {
...
-  const options = { tokenizer, alpha }
-  const classifier = createLocalClassifier(transactionsWithManualCat, options)
+
+  const MIN_SAMPLE_WEIGHT = 1
+  const MAX_SAMPLE_WEIGHT = 3
+  const MIN_WEIGHT_LIMIT = 2
+  const learnSampleWeight =
+    transactionsWithManualCat.length > MIN_WEIGHT_LIMIT
+      ? MIN_SAMPLE_WEIGHT
+      : MAX_SAMPLE_WEIGHT
```
ptbrowne: What is the rationale behind this ?

src/ducks/categorization/services.js
```diff
@@ -188,6 +192,14 @@ const localModel = async (classifierOptions, transactions) => {
   )
   localModelLog('debug', 'Alpha parameter value is ' + alpha)
 
+  if (nbUniqueCategories === 1) {
+    localModelLog(
+      'info',
+      'Not enough different categories, adding a fake transaction to balance the weight of the categories'
+    )
+    transactionsWithManualCat.push(FAKE_TRANSACTION)
+  }
```
ptbrowne: What is the rationale behind this ? Is there some doc to refer to ?

PR #555, opened 2018-11-21
fix(transactions): Align headers on medium screen

drazik:
Headers were not properly aligned on medium screen:

| Before | After |
|--------|-------|
|image|image|


ptbrowne:
Mmh ca me parait bizarre ? Pourquoi y-a-t il besoin d'avoir ces valeurs magiques comme padding ?

drazik:
We don't use a single table for the headings and the content, so we can't count on this to automatically align headings and content. We need to align them manually with paddings. We already to this on desktop for the "description" column. That's why there's this space on the left on medium screens.

PR #562, opened 2018-11-23
feat(TransactionHeader): Don't show columns headings on tablet

drazik:
We just want to be consistent with the mobile style, which don't show this.

### DIFF
--- a/src/ducks/transactions/Transactions.styl
+++ b/src/ducks/transactions/Transactions.styl
@@ -49,21 +49,6 @@
     .bnk-op-caticon
         cursor pointer
 
-    +medium-screen()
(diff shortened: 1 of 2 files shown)


src/ducks/transactions/Transactions.styl
```diff
@@ -49,21 +49,6 @@
...
-        .bnk-op-amount
-            cursor pointer
-            maxed-flex-basis 40%
-            padding-right 3.5rem
-        .bnk-op-actions
-            width 150px
-        .bnk-op-date
-        .bnk-op-action
-            display none
-
```
ptbrowne: 🎉

PR #565, opened 2018-11-23
Plug history chart to real data

drazik:
This removes the static data and retrieve real data to show balance history graphs.

Performance seems to be acceptable, but it will be subject to more tests on more different cases.

I also added tests for helpers that builds the balance history from current balance and transactions.

To show data as soon as it is available, we do the following:

* Show the history block with a spinner while transactions and accounts are fetching
* Show the account balance + a spinner when accounts have been fetched but transactions still fetching


src/ducks/balance/helpers.spec.js
```diff
@@ -35,27 +39,78 @@ describe('getBalanceHistory', () => {
...
+      { date: '2018-06-25T00:00:00Z', amount: 100 },
+      { date: '2018-06-24T00:00:00Z', amount: 10 },
+      { date: '2018-06-23T00:00:00Z', amount: -300 },
+      { date: '2018-06-22T00:00:00Z', amount: -15 }
+    ]
+
+    const history = getBalanceHistory(
+      account,
+      transactions,
+      new Date(2018, 5, 26)
```
kosssi: Peut être que mettre `parseDate('2018-06-26')` pour rendre vraiment le test clair (vive la gestion des mois dans `new Date`...)

src/ducks/balance/helpers.js
```diff
@@ -46,8 +46,13 @@ export const getBalanceHistories = (
  * @param {Object[]} transactions - The transactions to filter
  * @returns {Object[]}
  */
-export const getTransactionsForAccount = (accountId, transactions) =>
-  transactions.filter(t => t.account === accountId)
+export const getTransactionsForAccount = (accountId, transactions) => {
+  return transactions.filter(t => {
+    const toCompare = t.account.data ? t.account.data._id : t.account
```
kosssi: Peut être ajouter un commentaire spécifique pour éxpliquer que l'on gère ici les accounts avec les relations ou sans les relations

src/ducks/balance/helpers.js
```diff
@@ -29,8 +29,14 @@ export const getBalanceHistories = (accounts, transactions) => {
...
-const getTransactionsForAccount = (account, transactions) =>
-  transactions.filter(t => t.account === account._id)
+/**
+ * Get transactions that match an account id in a set of transactions
+ * @param {number} accountId - The ID of the account
+ * @param {Object[]} transactions - The transactions to filter
+ * @returns {Object[]}
+ */
+export const getTransactionsForAccount = (accountId, transactions) =>
+  transactions.filter(t => t.account === accountId)
```
ptbrowne:
```
transactions.filter(t => t.account === accountId)
getTransactionsForAccount(account._id, transactions)
```

Nit : the second line has more characters than the first and its meaning is less clear since it uses the name `get` when instead it is filtering. I find the first way easier to understand.

src/ducks/balance/helpers.js
```diff
@@ -9,7 +9,11 @@ import {
   format as formatDate
 } from 'date-fns'
 
-export const getBalanceHistories = (accounts, transactions) => {
+export const getBalanceHistories = (
+  accounts,
+  transactions,
+  from = startOfToday()
```
ptbrowne: The startOfToday will only be executed once, when the function is parsed so if you ahve a long running application, over several days (which I think will not be uncommon), we won't have the real start of today.

```
if (!from) { from = startOfToday() }
```

is better in this case.

src/ducks/balance/helpers.spec.js
```diff
@@ -35,27 +39,78 @@ describe('getBalanceHistory', () => {
...
       '2018-06-25',
       '2018-06-24',
       '2018-06-23',
       '2018-06-22'
     ]
+
+    const expectedValues = [8000, 7900, 7890, 8190, 8205]
+
+    expect(Object.keys(history)).toEqual(expectedKeys)
+    expect(Object.values(history)).toEqual(expectedValues)
```
ptbrowne: Why separate keys and values ? Keeping the object would make it clearer IMHO.

```
expect(history).toEqual({
      '2018-06-26': 8000
      '2018-06-25': 7900,
      '2018-06-24': 7890,
      '2018-06-23': 8190,
      '2018-06-22': 8205
})
```

src/ducks/balance/helpers.js
ptbrowne: `filterTransactionsByAccount` would be clearer IMHO.

src/ducks/balance/Balance.jsx
```diff
@@ -286,29 +290,76 @@ class Balance extends React.Component {
...
+            right: 16
+          },
+          showAxis: true,
+          axisMargin: 10,
+          tickFormat: TICK_FORMAT
+        }}
+      />
+    )
+  }
+
```
ptbrowne: I'm thinking a dedicated component (linking the Balance and History) would be better than a method here. It would benefit from not being re-rendered if it is pure.

src/ducks/balance/Balance.jsx
```diff
@@ -286,29 +290,76 @@ class Balance extends React.Component {
...
     return balanceHistory
   }
 
-  getChartData() {
-    const history = this.getBalanceHistory()
+  getChartData(balanceHistory) {
     const WINDOW_SIZE = 15
 
-    const balancesSma = sma(history.map(h => h.y), WINDOW_SIZE, n => n)
+    const balancesSma = sma(balanceHistory.map(h => h.y), WINDOW_SIZE, n => n)
```
ptbrowne: I thought we were not doing windowing anymore ?

drazik: Yep, I removed it in a later commit.

src/ducks/balance/helpers.js
drazik: Actually I removed this default parameter later.

src/ducks/balance/helpers.js
```diff
@@ -52,37 +51,51 @@ export const getTransactionsForAccount = (accountId, transactions) => {
...
 
   const transactionsByDate = groupBy(transactions, t =>
     formatDate(t.date, DATE_FORMAT)
   )
-  const mostDistantDay =
-    Object.keys(transactionsByDate).length > 0
-      ? getEarliestDate(...Object.keys(transactionsByDate))
-      : from
+
+  // const dateRange = getDateRange(Object.keys(transactionsByDate))
```
ptbrowne: Forgotten comment ?

src/ducks/balance/helpers.js
```diff
@@ -52,37 +51,51 @@ export const getTransactionsForAccount = (accountId, transactions) => {
   })
 }
 
+export const getDateRange = dates => {
+  return {
+    from: getEarliestDate(...dates),
```
ptbrowne: Nit ?

```
getEarliestDate.apply(null, dates)
```

drazik: Actually this function is not used. I remove it.

src/ducks/balance/Balance.jsx
drazik: Can you tell me more?

kosssi: We discussed with @drazik that we should then do a refacto to separate all that is graph in a separate duck ;)

src/ducks/balance/History.jsx
```diff
@@ -35,7 +42,9 @@ class History extends Component {
 History.propTypes = {
   accounts: PropTypes.array.isRequired,
   chartProps: PropTypes.object,
-  className: PropTypes.string
+  className: PropTypes.string,
+  isAccountsLoading: PropTypes.bool,
+  isChartLoading: PropTypes.bool
```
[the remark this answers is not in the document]
drazik: You're right, I refactored the component so it only takes an accounts collection and a transactions collection and do its magic.

PR #569, opened 2018-11-28
Fix date selector update when scrolling

ptbrowne:
The date selector was not updating because transactions rows refs were not passed to the TopMost instance : while scrolling, we could not figure out the highest visible row.

### DIFF
--- a/src/components/List/index.jsx
+++ b/src/components/List/index.jsx
@@ -8,8 +8,10 @@ export const Header = ({ children }) => (
   <div className={styles['c-list-header']}>{children}</div>
 )
 
-export const Row = ({ children }) => (
-  <div className={styles['c-list-row']}>{children}</div>
+export const Row = props => (
+  <div ref={props.onRef} className={styles['c-list-row']}>
+    {props.children}
+  </div>
 )
 
 export const Content = ListItemText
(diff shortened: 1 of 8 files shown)


src/ducks/transactions/TransactionRow.jsx
```diff
@@ -32,8 +32,14 @@ const sAmount = styles['bnk-op-amount']
 const sAction = styles['bnk-op-action']
 
 class _RowDesktop extends React.PureComponent {
-  onSelectTransaction = () =>
+  constructor(props) {
+    super(props)
+    this.onSelectTransaction = this.onSelectTransaction.bind(this)
```
drazik: Why do you need to change this? Is there a problem with class properties and fat arrows?

src/components/List/index.jsx
```diff
@@ -8,8 +8,10 @@ export const Header = ({ children }) => (
   <div className={styles['c-list-header']}>{children}</div>
 )
 
-export const Row = ({ children }) => (
-  <div className={styles['c-list-row']}>{children}</div>
+export const Row = props => (
+  <div ref={props.onRef} className={styles['c-list-row']}>
+    {props.children}
+  </div>
```
drazik: Nit: it seems it's a good use case for [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html)

Crash--: WHAT! Thank you! Didn't know about this one.

src/ducks/transactions/TransactionRow.jsx
ptbrowne: Yes, it is not possible to mock fat arrow methods in tests.

ptbrowne: https://github.com/airbnb/enzyme/issues/365#issuecomment-217368222

PR #573, opened 2018-11-29
feat: Use Portal to display AccountSwithMenu

kosssi:
To avoid z-index problems

### DIFF
--- a/src/ducks/account/AccountSwitch.styl
+++ b/src/ducks/account/AccountSwitch.styl
@@ -7,16 +7,10 @@
 
 .account-switch
     position relative
-    padding-bottom 20px
-
-    a
-    a:visited
-        color            dodgerBlue
-        text-decoration  none
-
-    a:hover
-    a:active
-        text-decoration  underline
+    top 0
+    padding 0
+    border 0
+    margin 0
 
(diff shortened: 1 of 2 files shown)


src/ducks/account/AccountSwitch.styl
```diff
@@ -7,16 +7,10 @@
...
-        color            dodgerBlue
-        text-decoration  none
-
-    a:hover
-    a:active
-        text-decoration  underline
+    top 0
+    padding 0
+    border 0
+    margin 0
```
ptbrowne: What are the default values ? Are they really necessary ?

src/ducks/account/AccountSwitch.styl
```diff
@@ -50,9 +44,6 @@
 
 baseZ=80
```
ptbrowne: It seems baseZ is used only at one other place, where it isn't needed now that we are rendered in a Portal. I think we can remove it.

src/ducks/account/AccountSwitch.styl
kosssi: This is a cut and paste but I had forgotten the deleted part

PR #574, opened 2018-11-29
fix: Warning that prevents me from seeing correctly the real mistakes

kosssi:
To remove warnings or errors

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -11,11 +11,10 @@ import {
 import { connect } from 'react-redux'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import { or } from 'airbnb-prop-types'
 import * as d3 from 'd3'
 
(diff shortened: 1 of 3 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -137,17 +136,6 @@ const BalanceRow = connect(state => ({
...
-  accountOrGroup: or([
-    {
-      account: PropTypes.object.isRequired
-    },
-    {
-      group: PropTypes.object.isRequired
-    }
-  ]).isRequired
-}
-
```
ptbrowne: So the BalanceRow has no propTypes ?

kosssi: I create a custom validator https://github.com/cozy/cozy-banks/pull/574/files#diff-e7d9783b37787e694e296c56a99fba5cR140 I remove `airbnb-prop-types` dependency ;)

PR #575, opened 2018-11-29
feat: Use ResponsiveContent on layout

kosssi:
Remove default padding on cozy-ui Content, it's used to not add negative margin after.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -6,7 +6,6 @@ import { Layout, Main, Content, Sidebar, IconSprite } from 'cozy-ui/react'
 import Nav from 'ducks/commons/Nav'
 import { Warnings } from 'ducks/warnings'
 import { FlagSwitcher } from 'cozy-flags'
-import styles from 'styles/main.styl'
 
 const ReactHint = ReactHintFactory(React)
 
@@ -18,7 +17,7 @@ const App = ({ children }) => (
     </Sidebar>
 
     <Main>
-      <Content className={styles['c-content']}>{children}</Content>
+      <Content>{children}</Content>
     </Main>
 
     {/* Outside every other component to bypass overflow:hidden */}
--- a/src/components/Spacing/Padded.jsx
+++ b/src/components/Spacing/Padded.jsx
@@ -1,0 +1,16 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import styles from './Padded.styl'
+
+const Padded = ({ children, className }) => (
+  <div className={cx(styles.Padded, className)}>{children}</div>
+)
+
+Padded.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string
+}
+
+export default Padded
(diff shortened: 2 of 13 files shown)


kosssi:
can you add a test? ^^

src/components/ResponsiveContent/ResponsiveContent.jsx
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import styles from './ResponsiveContent.styl'
+
+const ResponsiveContent = ({ children, className }) => {
+  const responsiveContentClassName = cx(
+    styles.ResponsiveContentSize_default,
+    className
+  )
```
ptbrowne: Nit: I would remove this variable, it does serve any purpose.

src/components/ResponsiveContent/ResponsiveContent.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import styles from './ResponsiveContent.styl'
+
+const ResponsiveContent = ({ children, className }) => {
+  const responsiveContentClassName = cx(
+    styles.ResponsiveContentSize_default,
```
ptbrowne: Nit: I don't like this class name, it is very verbose.

src/components/ResponsiveContent/ResponsiveContent.jsx
```diff
@@ -0,0 +1,27 @@
...
+    className
+  )
+
+  return <div className={responsiveContentClassName}>{children}</div>
+}
+
+ResponsiveContent.propTypes = {
+  /**
+   * The content of the element.
+   */
```
ptbrowne: Nit: No need for comments when they are self describing.

src/components/ResponsiveContent/ResponsiveContent.jsx
```diff
@@ -0,0 +1,27 @@
...
+  children: PropTypes.node.isRequired,
+  /**
+   * The CSS class name of the element to extend it.
+   */
```
ptbrowne: Nit: No need for comments when they are self describing.

For short comments I prefer this way, it adds less lines and is more compact.

```
Component.propTypes = {
  /** @type {String} - Name of the person */
  name: PropTypes.string.required
}
```

This is understood by styleguidist.

src/components/ResponsiveContent/ResponsiveContent.styl
```diff
@@ -0,0 +1,7 @@
+@require 'settings/breakpoints'
+
+.ResponsiveContentSize_default
+    padding 1.5rem 2rem
+
+    +small-screen()
+        padding 1rem
```
ptbrowne: I don't like the ResponsiveContent name, I would prefer Padded or something like this.

In some context, we could have something like:

```
class MyPage extends Component {
  render () {
    return <div>
      <ResponsiveContent>
           Blalbla with margins
      </ResponsiveContent>
      Blabla without margins, for example a table, that could be responsive.
      <ResponsiveContent>
           Re-blabla with margins
      </ResponsiveContent>
    </div> 
  }
```

I think ResponsiveContent is not really the right name since in this context, the content in the middle would certainly be responsive also, but just not padded.

PR #583, opened 2018-12-04
Fix health expense status action

drazik:
There was two flaws with this action:

* The `brands` property used in the `match` function was too ambiguous (which brands ? All brands ?). In this action we need to know brands without trigger. In other actions we need to know brands with trigger. So I refactored this to add a `hasTrigger` property to brands and pass all brands instead of only the ones without trigger. This way, the prop name is no ambiguous anymore
* The `userCanInstallHealthKonnector` function did what it was intended to do, but we actually didn't want that condition in this action. What we want to do is to always show the status if there are reimbursements, or if there is at least one health konnector with a trigger

### DIFF
--- a/src/ducks/transactions/TransactionActions.jsx
+++ b/src/ducks/transactions/TransactionActions.jsx
@@ -91,7 +91,11 @@ class TransactionActions extends Component {
     const { transaction } = this.props
     if (transaction) {
       const { urls, brands, bill } = this.props
-      const actionProps = { urls, brands, bill }
+      const actionProps = {
+        urls,
+        brands,
+        bill
+      }
       const actions = await findMatchingActions(transaction, actionProps)
       if (!this.unmounted) {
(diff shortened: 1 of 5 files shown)


ptbrowne:
LGTM, thanks Drazik 🎉

src/ducks/transactions/TransactionActions.jsx
```diff
@@ -117,7 +121,8 @@ class TransactionActions extends Component {
     if (
       nextProps.transaction !== this.props.transaction ||
       nextProps.urls !== this.props.urls ||
-      nextProps.brands !== this.props.brands ||
+      nextProps.brandsWithoutTrigger !== this.props.brandsWithoutTrigger ||
+      nextProps.brandsWithTrigger !== this.props.brandsWithTrigger ||
```
ptbrowne: Je crois qu'on peut enlever ce changement.

PR #606, opened 2018-12-06
Transaction: Update design on table head

kosssi:
Changes :
- Add a flag (transaction-history) to display separately the graph current balance and transaction
- Add blue frame on transaction
- Add new design on table head

You can see design with and without `transaction-history` flag:

without | with
--------|-----
capture d ecran 2018-12-06 a 09 00 24 | capture d ecran 2018-12-06 a 09 01 19
capture d ecran 2018-12-06 a 09 02 39 | capture d ecran 2018-12-06 a 09 01 47

### DIFF
--- a/src/components/Header/Header.styl
+++ b/src/components/Header/Header.styl
@@ -1,5 +1,15 @@
 @require 'settings/breakpoints'
 
+.HeaderFixed
+    position sticky
+    position -webkit-sticky
+    z-index 1
+    top 0
+    box-shadow 0 2px 4px 0 rgba(0, 0, 0, 0.24), 0 0 4px 0 rgba(0, 0, 0, 0.16)
+
+    +medium-screen()
+        top 3rem
+
 .HeaderColor_default
     background-color var(--white)
 
--- a/src/ducks/transactions/TransactionsPage.jsx
+++ b/src/ducks/transactions/TransactionsPage.jsx
@@ -31,7 +31,6 @@ import Loading from 'components/Loading'
 
 import { TransactionsWithSelection } from './Transactions'
 import TransactionHeader from './TransactionHeader'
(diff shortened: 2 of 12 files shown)


src/components/Header/Header.styl
```diff
@@ -1,5 +1,15 @@
 @require 'settings/breakpoints'
 
+.HeaderFixed
+    position sticky
```
drazik: Beware, it doesn't work in IE11. Do we have officially dropped the support of IE? I remember there's a Paper about it but I can't find it.

src/ducks/transactions/TransactionsPage.jsx
```diff
@@ -243,24 +242,20 @@ class TransactionsPage extends Component {
     }
 
     return (
-      <div
-        className={styles.TransactionPage__bottom + ' js-transactionPageBottom'}
```
drazik: This element was used in https://github.com/cozy/cozy-banks/blob/master/src/ducks/transactions/Transactions.jsx#L136-L143, but I don't see the corresponding change in the PR. Is it normal? Does the infinite scroll still work?

kosssi: I change it with `js-scrolling-element`

src/components/Header/Header.styl
kosssi: I think not

drazik: Talked about it on mattermost. We don't support IE anymore so it's OK :)

src/ducks/transactions/TransactionsPage.jsx
drazik: Oh yes sorry I missed the change in which you update the `querySelector`.

PR #610, opened 2018-12-06
Fixes on local categorization model

drazik:
* Update model parameters so it works better
* Move the appending of a fake transaction (it was not done at the right moment)
* Refactor to make the `localModel` function more readable
* Remove useless tests and add real interesting tests for `localModel`

### DIFF
--- a/src/ducks/categorization/services.spec.js
+++ b/src/ducks/categorization/services.spec.js
@@ -10,8 +10,8 @@ import { Transaction } from '../../models'
 jest
   .spyOn(Transaction, 'queryAll')
   .mockResolvedValue([
-    { label: 'AAAA BBBB', manualCategoryId: '200110' },
-    { label: 'AAAA BBBB', manualCategoryId: '200110' }
+    { amount: 3001.71, label: 'AAAA BBBB', manualCategoryId: '200110' },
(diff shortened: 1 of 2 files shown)


src/ducks/categorization/services.spec.js
```diff
@@ -1,13 +1,26 @@
...
 import { tokenizer } from '.'
-
-const mockLearn = jest.fn()
-jest.mock('classificator', () => jest.fn(() => ({ learn: mockLearn })))
+import { Transaction } from '../../models'
+
+jest
+  .spyOn(Transaction, 'queryAll')
+  .mockResolvedValue([
+    { label: 'AAAA BBBB', manualCategoryId: '200110' },
```
francoisWeber: It would be preferable to add an `amount` field in the learning transactions. Under the hood, these operations will pass through `getLabelWithTags()` that waits for an `amount`field. Currently the tests are certainly OK because the fallback of `getLabelWithTags()` if no amount is givent corresponds to a salary *and* because these operations actually correspond to a salary. But the tests will probably fail if we add some operations not supposed to be associated with `{ amount: 3001.71, label: 'AAAA BBBB' }`

PR #612, opened 2018-12-07
fix(settings): Remove _type from default settings

drazik:
It led to an error when saving settings in the service. There was no error in the app because it uses the new cozy-client that handles it automatically.


ptbrowne:
LGTM, thanks Drazik!

PR #647, opened 2018-12-17
feat: Add a setting to show the balance history

drazik:
The balance history on balance page is not behind a flag anymore. Instead, the user can enable it via a setting.

image

I refactored the components used in this page so we have more flexibility and can achieve what we wanted easily.

### DIFF
--- a/src/ducks/settings/Configuration.jsx
+++ b/src/ducks/settings/Configuration.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import TogglePane, {
   TogglePaneTitle,
   TogglePaneSubtitle,
@@ -13,7 +13,7 @@ import { flowRight as compose, set } from 'lodash'
 import Loading from 'components/Loading'
 import { getDefaultedSettingsFromCollection } from './helpers'
 
-class Configuration extends Component {
+class Configuration extends React.PureComponent {
   saveDocument = async doc => {
     const { saveDocument } = this.props
     await saveDocument(doc)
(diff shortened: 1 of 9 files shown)


src/ducks/settings/TogglePane.jsx
```diff
@@ -1,83 +1,31 @@
-import React, { Component } from 'react'
-import PropTypes from 'prop-types'
+import React from 'react'
```
kosssi: You can import PureComponent to not repeat React ?
```suggestion
 import React, { PureComponent } from 'react'
```

src/ducks/settings/ToggleRow.jsx
```diff
@@ -31,7 +31,7 @@ class ToggleRow extends Component {
             {hasValue && (
               <input
                 type="number"
-                onChange={e => onChangeValue(name, parseNumber(e.target.value))}
+                onChange={e => onChangeValue(parseNumber(e.target.value))}
```
kosssi: As we used a function here the ToggleRow always rerender even if Props don't change. Same line 49

src/ducks/settings/Configuration.jsx
```diff
@@ -1,17 +1,41 @@
...
 import { translate } from 'cozy-ui/react'
 import { isCollectionLoading } from 'ducks/client/utils'
 import { queryConnect, withMutations } from 'cozy-client'
 import { settingsConn } from 'doctypes'
-import { flowRight as compose } from 'lodash'
+import { flowRight as compose, set } from 'lodash'
 import Loading from 'components/Loading'
+import { getDefaultedSettingsFromCollection } from './helpers'
 
 class Configuration extends Component {
```
kosssi: It can be a PureComponent ;)

PR #665, opened 2018-12-19
fix: Notifications

drazik:
We had a problem due to code minification on production mode. The classes names were mangled by the minification process, but since we relied on it to do the link between settings and classes, it was broken. The result was that even if the user had notifications enabled, we treated it as disabled.

After some research, I found that it broke between webpack 3 (no problem) and webpack 4. I looked at Terser options, but I felt it was better to just not rely on classes `name` property, so we don't rely on a tool configuration to make our code work. I mean if in the future we change the minifier, but forget to set the right configuration, this code will again. I think it's not safe at all.

### DIFF
--- a/src/ducks/notifications/services.spec.js
+++ b/src/ducks/notifications/services.spec.js
@@ -1,0 +1,37 @@
+// Mocking handlersbars since we don't want to test HTML rendering here and it fails in the tests context
+jest.mock('handlebars')
+
+import { getEnabledNotificationClasses } from './services'
+import BalanceLower from './BalanceLower'
+import TransactionGreater from './TransactionGreater'
+import HealthBillLinked from './HealthBillLinked'
+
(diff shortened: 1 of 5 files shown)


src/ducks/notifications/services.spec.js
```diff
@@ -0,0 +1,36 @@
+jest.mock('handlebars')
```
kosssi: It was forgotten

drazik: It's here on purpose because handlebars was throwing an error in test context. And since we are not testing the HTML part here, I just mocked it for now. We can see what's wrong with handlebars in tests when we try to test the HTML rendering.

kosssi: maybe you can add a comment?

drazik: Yep, here it is !

PR #697, opened 2018-12-27
Alert for konnectors waiting for manual update

drazik:
When a change in a connector requires the user to accept new TOS, we want to show an alert with a button that redirects the user to the list of connectors waiting for a manual update.

The `Info` component used is a copy/pasted and improved version of the cozy-ui one. I'm going to have a look with designers to see if we merge the update in cozy-ui or not.

Instance for validation : https://testbanksbiinfo-banks.cozy.works/#/balances
You should activate the flag: `flag('konnector-update-info', true)`. I installed a "Axa Banque" connector and updated it so the alert will be shown. But the whole feature may not whole correctly since the connector is not really waiting for a manual update.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -10,6 +10,7 @@ import { settingsConn } from 'doctypes'
 import { queryConnect } from 'cozy-client'
 import { getDefaultedSettingsFromCollection } from 'ducks/settings/helpers'
 import { Alerter } from 'cozy-ui/react'
+import KonnectorUpdateInfo from './KonnectorUpdateInfo'
 
 const ReactHint = ReactHintFactory(React)
 
@@ -25,7 +26,10 @@ const App = props => {
       </Sidebar>
 
       <Main>
-        <Content>{props.children}</Content>
+        <Content>
+          {flag('konnector-update-info') && <KonnectorUpdateInfo />}
+          {props.children}
+        </Content>
       </Main>
 
       {/* Outside every other component to bypass overflow:hidden */}
@@ -37,4 +41,5 @@ const App = props => {
     </Layout>
   )
 }
+
 export default queryConnect({ settingsCollection: settingsConn })(App)
--- a/src/components/Info/index.js
+++ b/src/components/Info/index.js
@@ -1,0 +1,34 @@
+import React from 'react'
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import styles from './styles.styl'
+import { Icon, SubTitle } from 'cozy-ui/react'
+
+const Infos = ({ icon, children, className, title }) => {
+  return (
+    <div className={cx(styles['infos'], 'u-p-1', className)}>
+      <div className={styles['infos--container']}>
+        {icon && <Icon icon={icon} className={styles['infos--icon']} />}
+        <div
+          className={cx(styles['infos--content'], {
+            ['u-pl-half']: icon !== null
+          })}
+        >
+          {title && (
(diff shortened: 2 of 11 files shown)


src/components/KonnectorUpdateInfo/styles.styl
```diff
@@ -0,0 +1,4 @@
+.KonnectorUpdateInfo__message a
+    font-weight bold
+    text-decoration none
+    color dodgerBlue
```
kosssi: can you use `var(--primary)`
```suggestion
     color var(--primary)
```

drazik: Oh yes. And I have other hard coded colors that I forgot to manage. Thanks for the headsup

src/components/Info/styles.styl
```diff
@@ -0,0 +1,24 @@
+@require 'settings/palette.styl'
+@require 'tools/mixins'
+
+.infos
+    background paleGrey
+    display flex
+    align-items center
+    background-color #fff2f2
```
kosssi: Do not use color code anymore, only [variable](https://docs.cozy.io/cozy-ui/styleguide/section-settings.html) or rgba black or white.

Same on https://github.com/cozy/cozy-banks/pull/697/files#diff-2b34118a76524e9121686aa071cf6a28R24

src/components/Info/styles.styl
```diff
@@ -0,0 +1,24 @@
...
+
+.infos--icon
+    width rem(16)
```
kosssi: why you don't use 1rem directly?

src/components/Info/styles.styl
```diff
@@ -0,0 +1,24 @@
+@require 'settings/palette.styl'
```
kosssi: you can remove it. It's not necessary and I duplicate code https://github.com/cozy/cozy-ui/issues/762

src/components/Info/styles.styl
kosssi: If you remove it you can remove `@require 'tools/mixins'`

src/components/Info/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import styles from './styles.styl'
+import { Icon, SubTitle } from 'cozy-ui/react'
+
+const Infos = ({ icon, children, className, title }) => {
```
kosssi: Can you use an PureComponent?

src/components/App.jsx
```diff
@@ -25,7 +26,10 @@ const App = props => {
       </Sidebar>
 
       <Main>
-        <Content>{props.children}</Content>
+        <Content>
+          {flag('bi-alert') && <KonnectorUpdateInfo />}
```
kosssi: How does it work on the phone when you are not connected?

kosssi:
Can you add a screenshot?

drazik:
> Can you add a screenshot?

I want to deploy it using `deploy2cozy` but it has a small bug. I'm fixing it, then I deploy it.

src/components/Info/styles.styl
drazik: It's a copy/paste from https://github.com/cozy/cozy-ui/blob/master/react/Infos/styles.styl. I don't really know why we use this helper.

src/components/App.jsx
drazik: You want to say when we are offline?

PR #703, opened 2018-12-28
chore: Flags at build time

drazik:
This adds the possibility to enable some flags at build time.

For example: `COZY_FLAGS=balance-history,transaction-history yarn build` will make a build with `balance-history` and `transaction-history` flags enabled by default.

This can be useful to generate mobile builds with enabled flags, so testing the feature is easier. Or to generate web builds with enabled flags and deploy them with `deploy2cozy`. Maybe this kind of feature can be integrated to CCA in the future (for example when we switch to cozy-scripts :p).

The only downside I don't know how to bypass is that since we use `localStorage` for web flags, when we enable the flag `f` for a build, then we do another build without enabling it, it will remain enabled on the instance.


ptbrowne:
Nice !

CPatchane:
You still can integrate this to CCA without switching for now :P

PR #713, opened 2019-01-02
Group balance panels

drazik:
This is the first implementation of the balance page with group panels.

It contains the MUI's `ExpansionPanel` overrides that fits our needs. It is intended to go in cozy-ui when it's fully implemented, but it's easier to keep it in Banks for now.

The feature is behind the `balance-panels` flag. You can see it live on https://testbanksexpansionpanel-banks.cozy.works/


src/ducks/balance/helpers.js
```diff
@@ -135,3 +135,16 @@ export const balanceHistoryToChartData = history => {
...
   return data
 }
+
+/**
+ * Returns a group balance (all its accounts balance sumed)
+ * @param {Object} group
+ * @returns {number}
+ */
+export const getGroupBalance = group => {
+  if (!group || !group.accounts || !group.accounts.data) {
```
kosssi: you can write:
```
const data = get(group, 'accounts.data')
if (!data) {
```

src/ducks/balance/helpers.js
```diff
@@ -135,3 +135,16 @@ export const balanceHistoryToChartData = history => {
...
+    return 0
+  }
+
+  return sumBy(group.accounts.data, account => get(account, 'balance') || 0)
```
kosssi: you can simplify with default value of get function:
```
  return sumBy(group.accounts.data, account => get(account, 'balance', 0))
```

src/ducks/balance/helpers.js
```diff
@@ -148,5 +148,5 @@ export const getGroupBalance = group => {
     return 0
   }
 
-  return sumBy(group.accounts.data, account => get(account, 'balance') || 0)
+  return sumBy(group.accounts.data, account => get(account, 'balance', 0))
```
kosssi: you can use accounts:
```
    return sumBy(accounts, account => get(account, 'balance', 0))
```

PR #743, opened 2019-01-08
Balance panels desktop layout

drazik:
On desktop we show more informations. I used Grid layout since it's widely compatible and the `fr` unit is very useful here. I tested it via Browserstack to make sure the compatibility is good: it is.

You can see the changes live on https://testbanksgrouppanelsdesktop-banks.cozy.works/#/balances

### DIFF
--- a/src/ducks/balance/components/AccountsList.jsx
+++ b/src/ducks/balance/components/AccountsList.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { sortBy, flowRight as compose } from 'lodash'
 import { withRouter } from 'react-router'
+import cx from 'classnames'
 import { Figure } from 'components/Figure'
 import withFilteringDoc from 'components/withFilteringDoc'
 import { getAccountLabel } from 'ducks/account/helpers'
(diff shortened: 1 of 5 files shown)


src/ducks/balance/components/AccountsList.jsx
```diff
@@ -18,23 +20,52 @@ class AccountsList extends React.PureComponent {
...
+                className={cx(
+                  styles.AccountsListItem__column,
+                  styles['AccountsListItem__column--secondary']
+                )}
+              >
+                N°
+                {a.number}
+              </span>
+            )}
+            {!isMobile && (
```
kosssi: this is the same condition of https://github.com/cozy/cozy-banks/pull/743/files#diff-93efe2b339535bf4f631602808c2b4d6R39 maybe you can use:

```
{!isMobile && (
  <React.Fragment>
    <span
      className={cx(
        styles.AccountsListItem__column,
        styles['AccountsListItem__column--secondary']
      )}
    >
      N°
      {a.number}
    </span>
    <span
      className={cx(
        styles.AccountsListItem__column,
        styles['AccountsListItem__column--secondary']
      )}
    >
      {a.institutionLabel}
    </span>
  </React.Fragment>
)}
```

PR #748, opened 2019-01-09
feat(BalancePanels): Sort groups with undefined last

drazik:
Groups were sorted on their translated label. So "other accounts" was not always the last group.

### DIFF
--- a/src/ducks/groups/helpers.js
+++ b/src/ducks/groups/helpers.js
@@ -1,4 +1,4 @@
-import { groupBy } from 'lodash'
+import { groupBy, sortBy } from 'lodash'
 import { ACCOUNT_DOCTYPE, GROUP_DOCTYPE } from 'doctypes'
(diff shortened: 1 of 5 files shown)


src/ducks/groups/helpers.js
```diff
@@ -22,3 +22,40 @@ export const buildVirtualGroups = accounts => {
...
+ */
+export const sortGroups = (groups, translate) => {
+  const [othersGroup] = groups.filter(g => g.virtual && g.label === 'undefined')
+
+  const sortedGroups = sortBy(
+    groups
+      .filter(g => g !== othersGroup)
+      .map(g => translateGroup(g, translate)),
+    g => g.label
+  )
```
ptbrowne: The sort function can be improved by checking if the group is special (ie `label === undefined`) and returning `Infinity` if it does. This way there is no need for the find and the push at the end.

```
g => g.label === undefined ? Infinity : g.label 
```

PR #756, opened 2019-01-10
Show time since last update on account rows

drazik:
This uses the `cozyMetadata.updatedAt` property to show the time elapsed since the last update on each bank accounts.

Live here: https://testbanksaccountupdatedate-banks.cozy.works/#/balances


src/ducks/balance/components/AccountRow.jsx
```diff
@@ -0,0 +1,94 @@
...
+            <Icon
+              icon="sync"
+              width="10"
+              color="currentColor"
+              className={styles.AccountRow__updatedAtIcon}
+            />
+            {updatedAt}
+          </div>
+        </div>
+        {!isMobile && (
```
kosssi: you can use only one condition:
```
{!isMobile && (
  <React.Fragment>
    ...
  </React.Fragment>
)}

PR #778, opened 2019-01-14
fix(Notification): Config value can be null, so we disable it

kosssi:
I test it with cozyToDeploy after lunch: https://testbanksfixnotification-banks.cozy.works/#/settings

### DIFF
--- a/src/ducks/notifications/BalanceLower.js
+++ b/src/ducks/notifications/BalanceLower.js
@@ -116,5 +116,6 @@ class BalanceLower extends Notification {
 }
 
 BalanceLower.settingKey = 'balanceLower'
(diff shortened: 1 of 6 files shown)


src/targets/services/onOperationOrBillCreate.js
```diff
@@ -169,6 +169,7 @@ const onOperationOrBillCreate = async options => {
   log('info', `COZY_URL: ${process.env.COZY_URL}`)
   log('info', 'Fetching settings...')
   const setting = await Settings.fetchWithDefault()
+  log('info', JSON.stringify(setting))
```
drazik: Do we really want this log?

PR #779, opened 2019-01-14
Refactor virtual groups logic

drazik:
This PR refactors the grouping logic of the accounts to build virtual groups for each type. The goal is to be able put multiple account types to a single virtual group. For example types `Unknown` and `None` are mapped to `Other` so they all belong to the same group. Same for `Perco`, `Perp` and `Article83` that are mapped to `RetirementPlan`.

We also wanted to put all types we don't have a translation for to the `Other` type. So we never show an not translated type.

To make it more easier, I made a helper function `getAccountType` that does the mapping logic.

You can see it live on https://testbanksgroupaccounttypes-banks.cozy.works/#/balances where :

* "Isabelle checkings" has type `Perco` mapped to `Plan Épargne retraite`
* "Claude chekings" has type `Article83` mapped to `Plan Épargne retraite`
* "Genevieve checkings" has type `Perp` mapped to `Plan Épargne retraite`
* "Louise checkings" has type `Unkown` mapped to `Autres comptes`
* "Isabelle savings" has type `None` mapped to `Autres comptes`


src/ducks/account/helpers.js
```diff
@@ -46,3 +46,71 @@ export const distanceInWords = distance => {
...
 }
+
+/**
+ * Get the translated type of an account
+ * @param {Object} account
+ * @param {Function} translate
+ * @returns {String}
+ */
+export const getAccountTranslatedType = (account, translate) => {
+  const accountTypesMap = {
```
kosssi: This const can be extract to use it on `getAccountTranslatedType` and `getAccountType` ?

src/ducks/account/helpers.js
```diff
@@ -46,3 +46,71 @@ export const distanceInWords = distance => {
...
   return 'Balance.updated_at.n_days_ago'
 }
+
+/**
+ * Get the translated type of an account
+ * @param {Object} account
+ * @param {Function} translate
+ * @returns {String}
+ */
+export const getAccountTranslatedType = (account, translate) => {
```
kosssi: this function is not used ?

src/ducks/account/helpers.js
drazik: Actually I forgot to remove `getAccountTranslatedType`

src/ducks/account/helpers.js
drazik: Yep, thanks

PR #796, opened 2019-01-17
Onboarding refactoring

kosssi:
Some improvements...


src/ducks/categories/CategoriesHeader.jsx
```diff
@@ -25,7 +25,13 @@ class CategoriesHeader extends PureComponent {
       <Fragment>
         <AccountSwitch small={selectedCategory !== undefined} {...colorProps} />
         {selectedCategory && (
-          <BackButton onClick={previousItem && previousItem.onClick} />
+          <BackButton
+            onClick={
+              previousItem && previousItem.onClick
+                ? previousItem.onClick
+                : undefined
```
drazik: If `onClick` doesn't exist on the `previousItem` object, then it's already `undefined`. There is no need to explicitly check it here.

PR #810, opened 2019-01-22
Groups and accounts checked/disabled state management

drazik:
This adds the possibility to check/uncheck groups and accounts.

The rules are the following:

* When we uncheck a group, we want the accounts rows to be disabled, but their checked state remains the same (ie. unchecking a group doesn't uncheck the switches of all its accounts)
* We need to know the state of each account in each group. It means we can't flatten the object holding the state. This is because we still put an account balances in the balance history computation until it is unchecked in all the groups it's part of (this is not part of this PR because I wanted to cut the work to avoid monstrous PR)

I used an object with the following shape to hold the state:
```
{
  [GROUP_ID]: {
    checked: boolean,
    accounts: {
      [ACCOUNT_ID]: {
        checked: boolean,
        disabled: boolean
      }
    }
  }
}
```

The reason is because this object can be serialized and put in the local storage or `io.cozy.bank.settings` to remember the user's choices.

Finally, each `Switch` has an `id` that is the path to its value in the state object. So we can easily update it even if it's nested.

peek 22-01-2019 15-35

Result is live on https://testbankshandleswitch-banks.cozy.works

### DIFF
--- a/src/ducks/balance/components/AccountRow.jsx
+++ b/src/ducks/balance/components/AccountRow.jsx
@@ -63,7 +63,8 @@ class AccountRow extends React.PureComponent {
       <li
         className={cx(styles.AccountRow, {
           [styles['AccountRow--hasWarning']]: hasWarning,
-          [styles['AccountRow--hasAlert']]: hasAlert
+          [styles['AccountRow--hasAlert']]: hasAlert,
+          [styles['AccountRow--disabled']]: !checked || disabled
         })}
         onClick={onClick}
       >
--- a/src/ducks/balance/components/AccountRow.jsx
+++ b/src/ducks/balance/components/AccountRow.jsx
@@ -25,6 +25,7 @@ class AccountRow extends React.PureComponent {
     t: PropTypes.func.isRequired,
     warningLimit: PropTypes.number.isRequired,
     checked: PropTypes.bool.isRequired,
+    disabled: PropTypes.bool.isRequired,
     switchIdPrefix: PropTypes.string.isRequired,
     onSwitchChange: PropTypes.func.isRequired
   }
@@ -41,6 +42,7 @@ class AccountRow extends React.PureComponent {
       t,
       warningLimit,
       checked,
+      disabled,
       switchIdPrefix,
       onSwitchChange
     } = this.props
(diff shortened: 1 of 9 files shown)


src/ducks/balance/helpers.spec.js
```diff
@@ -271,3 +272,43 @@ describe('getGroupBalance', () => {
     expect(getGroupBalance(group)).toBe(8900)
   })
 })
+
+fdescribe('getSwitchesState', () => {
```
ptbrowne:
```suggestion
describe('getSwitchesState', () => {
```

src/ducks/balance/components/AccountsList.jsx
```diff
@@ -18,7 +21,13 @@ class AccountsList extends React.PureComponent {
   }
 
   render() {
-    const { accounts, warningLimit } = this.props
+    const {
+      accounts,
+      warningLimit,
+      switches,
+      switchesIdPrefix,
```
ptbrowne: The switchesIdPrefix directly maps to the groups, I think it would be simpler not to introduce a new concept here (switch prefix id) and pass simply the group.

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -123,10 +132,13 @@ class AccountRow extends React.PureComponent {
             currencyClassName={styles.AccountRow__figure}
           />
           <Switch
-            defaultChecked
+            disabled={disabled}
+            checked={checked}
             color="primary"
             onClick={this.handleSwitchClick}
             className={styles.AccountRow__switch}
+            id={`${switchIdPrefix}.accounts[${account._id}]`}
```
ptbrowne: ... this way we have `${group.id}.accounts[${account._id}]` here, I find it personnally simpler.

ptbrowne: I find it a bit unusual to use the HTML id attribute when using event handlers. Is there a particular reason not to use a more conventional :

```
handleChange() {
   const { group, account, onSwitchChange } = this.props
   onSwitchChange(`${group.id}.accounts[${account._id}]`)
}
```

That said, I see no particular reason for not doing this.

PR #811, opened 2019-01-22
fix: Gradiant should start on min y

kosssi:
I change just value of `y0` to start gradient on minimum

capture d ecran 2019-01-22 a 15 37 34


ptbrowne:
Simple, basique, LGTM

PR #813, opened 2019-01-22
Update chart when checking/unchecking groups/acounts

drazik:
This adds the link between groups/accounts switch and the balance history chart on the balance page.

We look the switches state object to find which accounts are enabled, then we pass them to the chart and we compute the balance only on them.

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -149,7 +149,7 @@ class Balance extends PureComponent {
               <div className={styles.Balance__subtitle}>
                 {t('BalanceHistory.subtitle')}
               </div>
-              <History accounts={accountsCollection} />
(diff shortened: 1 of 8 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -59,6 +59,33 @@ class Balance extends PureComponent {
...
+
+    return Object.values(switches).reduce((occurrences, group) => {
+      const occurrence = group.accounts[account._id]
+
+      if (occurrence) {
+        return [...occurrences, occurrence]
+      }
+
+      return occurrences
+    }, [])
```
ptbrowne: ca serait plus clair avec `Object.values(switches).map(group =>  group.accounts[account._id]).filter(Boolean)` IMHO.

PR #856, opened 2019-01-30
Fix account type translations in account details

drazik:
I forgot to update the accounts details pages when I updated accounts types translations.

https://testbanksfixaccountdetails-banks.cozy.works


ptbrowne:
Tests ? :)

PR #862, opened 2019-01-31
Fix transactions sorting

drazik:
We were not showing the transactions correctly because we were slicing on unordered data. Now we sort the transactions by date, then we slice it, then we do the grouping.


ptbrowne:
Wow, tricky bug ! Nice catch 👍 Can you add a test please ?

ptbrowne:
Could we sortBy inside updateTransactions ? It would at least cut the number of sortBy ?

drazik:
Here it is @ptbrowne: we now sort in `updateTransactions` (that is called on mount and on update if the transactions changed), and the related test.

PR #871, opened 2019-02-04
Handle history graph resize

drazik:
This fixes:
- blank line width
- window resizing don't move tooltip
- update all element on window resizing

[You can test it](https://testbanksgraphresize-banks.cozy.works)

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,7 @@
     "localforage": "1.7.3",
     "lodash": "4.17.11",
     "matomo-tracker": "2.2.0",
+    "memoize-one": "5.0.0",
     "mjml": "3.3.5",
     "node-polyglot": "2.3.0",
     "null-loader": "0.1.1",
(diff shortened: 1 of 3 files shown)


drazik:
Maybe we could delegate the selected item logic to the parent component? Here the generic `LineChart` component is saving a `date` in its state, but this component must be generic in case we want to draw a line chart for something else in the app.

If the parent component handles it, it can also do some optimization to prevent unnecessary renders (like you discussed on mattermost).

If we think this is optional for now, then LGTM (but I can't approve the PR because I opened it)

kosssi:
@drazik J'ai fait 2 optimisations autour du composant :
- pour rechercher un item à partir de bisect
- une mise à jour des éléments seulement quand c'est nécéssaire

Je retiens ton commentaire mais je ne vais pas passer plus de temps pour avancer sur d'autres sujets.

package.json
```diff
@@ -180,6 +180,7 @@
     "localforage": "1.7.3",
     "lodash": "4.17.11",
     "matomo-tracker": "2.2.0",
+    "memoize-one": "5.0.0",
```
ptbrowne: lodash has memoize.

drazik:
Lgtm

PR #872, opened 2019-02-04
fix: First fix for local categ of historical users 🦎

francoisWeber:
Safeguard applied to the local categorization.

_Problem we try to solve:_
The local model is used on a label even if no tokens extracted from this label appears in the vocabulary. On some extreme cases (*), it can lead to a wrong categorization with high probability only based on the amount. 
_Proposed solution:_
If less than `LOCAL_MODEL_PCT_TOKENS_IN_VOC_THRESHOLD` tokens extracted from the label appear in the local model vocabulary, then no prediction is done using the local model and its associated `(localCategoryId, localCategoryProba)` is set to a fallback value `('0', 0.10)`

(*) such extreme cases will be tackled in a future PR


src/ducks/categorization/services.js
```diff
@@ -213,24 +216,54 @@ export const localModel = async (classifierOptions, transactions) => {
       'No classifier, impossible to categorize transactions'
     )
     return
+  } else {
+    return classifier
```
drazik: You should always return `classifier` and check if it is null in the `localModel` function. Here, we can return `undefined` and it seems we don't check it in `localModel` so there's potentially a bug.

src/ducks/categorization/services.js
```diff
@@ -213,24 +216,54 @@ export const localModel = async (classifierOptions, transactions) => {
...
+}
+
+export const pctOfTokensInVoc = (tokens, vocabularyArray) => {
+  const n_tokens = tokens.length
+  const intersection = tokens.filter(t => -1 !== vocabularyArray.indexOf(t))
+  return intersection.length / n_tokens
+}
+
+export const localModel = async (classifierOptions, transactions) => {
+  const classifier = await createLocalModel(classifierOptions)
```
drazik: Here, we should check if `classifier` really exists and exit early if it doesn't.

src/ducks/categorization/services.js
```diff
@@ -213,24 +216,54 @@ export const localModel = async (classifierOptions, transactions) => {
...
+      // If KO : abort with category '0' and proba 1/nbUniqueCat
+      localModelLog('info', `Giving up for ${label}`)
+      const category = LOCAL_MODEL_CATEG_FALLBACK
+      const proba = LOCAL_MODEL_PROBA_FALLBACK
+      transaction.localCategoryId = category
+      transaction.localCategoryProba = proba
+  
+      localModelLog('info', `Results for ${label} :`)
+      localModelLog('info', `localCategory: ${category}`)
+      localModelLog('info', `localProba: ${proba}`)
```
drazik: These 3 lines are duplicated in both `if/else` branches. You can put it outside the conditions.

src/ducks/categorization/services.js
```diff
@@ -206,31 +209,57 @@ export const localModel = async (classifierOptions, transactions) => {
...
+  const n_tokens = tokens.length
+  const intersection = tokens.filter(t => -1 !== vocabularyArray.indexOf(t))
+  return intersection.length / n_tokens
+}
+
+export const localModel = async (classifierOptions, transactions) => {
+  const classifier = await createLocalModel(classifierOptions)
 
-  if (!classifier) {
+  if (classifier !== undefined) {
```
drazik:
```suggestion
  if (classifier) {
```

Actually `createLocalClassifier` returns `null` if there is no manually categorized transactions. So checking only `undefined` is not enough and will try to categorize transactions with no model at all, crashing the process.

Sorry if I was not clear in my previous comment.

francoisWeber:
I just added the test you mentioned, Drazik

src/ducks/categorization/services.js
```diff
@@ -22,6 +22,9 @@ const FAKE_TRANSACTION = {
   label: 'thisisafaketransaction',
   manualCategoryId: '0'
 }
+const LOCAL_MODEL_CATEG_FALLBACK = '0'
```
ptbrowne: Why is it a string 0 ?

src/ducks/categorization/services.js
```diff
@@ -206,31 +209,57 @@ export const localModel = async (classifierOptions, transactions) => {
     { ...classifierOptions, ...options.initialization },
     options.configuration
   )
+  return classifier
+}
+
+export const pctOfTokensInVoc = (tokens, vocabularyArray) => {
+  const n_tokens = tokens.length
```
ptbrowne: nTokens is more javascript-y

src/ducks/categorization/services.js
francoisWeber: The category `'0'` means `uncategorized` in our data model. See [here](https://github.com/cozy/cozy-banks/blob/master/src/ducks/categories/tree.json)

src/ducks/categorization/services.js
francoisWeber: My (pythonic) mistake

src/ducks/categorization/services.js
ptbrowne: Can you add a comment ? Or better (IMHO), a variable `const UNCATEGORIZED_CAT_ID = '0'`. I feel like in 2 weeks, we won't remember this otherwise :).

PR #880, opened 2019-02-05
chore(publish): Use cozy-app-publish instead of custom script

drazik:
Drop our custom publish script to use cozy-app-publish only.

Since we were already using cozy-app-publish manual mode, we already have most of the required configuration. I just added the mattermost channel variable in travis config.


package.json
```diff
@@ -139,7 +139,6 @@
     "stylus": "0.54.5",
     "stylus-loader": "3.0.2",
     "svg-sprite-loader": "3.9.2",
-    "tar": "4.4.8",
```
ptbrowne: Cool si t'as trouvé ca via madge :)

kosssi:
😍

PR #881, opened 2019-02-05
fix: Second safeguard to stabilize local categ 

francoisWeber:
Added another safeguard to avoid a too high strenght of amounts tags in categorization. After having learnt the local model's parameter, we cast to log the sub-parameters associated to amounts.

Also added a new test file based on snapshot. 

EDIT 06/02/2019 5:40pm : used the latest version of global model within the tests


src/ducks/categorization/services.js
```diff
@@ -186,6 +204,37 @@ const getLocalClassifierOptions = transactionsWithManualCat => {
...
+}
+
+export const reweightModel = classifier => {
+  // loop over categories in the wordFrequencyCat attribute
+  const wordFrequencyCount = classifier.wordFrequencyCount
+  // for each category
+  for (const category in wordFrequencyCount) {
+    if (wordFrequencyCount.hasOwnProperty(category)) {
+      // extract its word-frequency count `wfc`
+      const categoryWordsFrequencyClounts = wordFrequencyCount[category];
```
drazik: There's a typo I think: `categoryWordsFrequencyClounts` => `categoryWordsFrequencyCounts`. No bug because you use the same name after, but maybe you can rename the variable.

src/ducks/categorization/services-fixtures.spec.js
```diff
@@ -0,0 +1,123 @@
+import { localModel } from "./services";
+import { tokenizer } from ".";
+import { Transaction } from "../../models";
+const path = require("path");
+
+const fixturePath = path.join(__dirname, "fixtures");
+const cat2name = require("/Users/fweber/Code/auto-bank-categorization/data/tree.json");
```
drazik: This should be more generic. Tests in CI can't find this file to it fails. According to the name of the file, it already exists in our code: https://github.com/cozy/cozy-banks/blob/master/src/ducks/categories/tree.json. Is it the same file?

src/ducks/categorization/services-fixtures.spec.js
```diff
@@ -0,0 +1,123 @@
...
+    // embed results informations
+    op.method = method;
+    op.usedProba = proba;
+    op.catNameTrue = cat2name[trueCatId];
+    op.catNameDisplayed = cat2name[displayedCatId];
+    // output final status
+    let status;
+    if (trueCatId === "0") {
+      // special status if the op was not categorized at all
+      status = 2;
```
ptbrowne: A string could be clearer. 

status = "NOT_CATEGORIZED"

src/ducks/categorization/services.js
```diff
@@ -22,6 +22,24 @@ const FAKE_TRANSACTION = {
...
+                          'tag_neg tag_noise_neg',
+                          'tag_pos',
+                          'tag_noise_pos',
+                          'tag_pos tag_noise_pos',
+                          'tag_income',
+                          'tag_pos tag_income',
+                          'tag_b_income',
+                          'tag_pos tag_b_income',
+                          'tag_activity_income',
+                          'tag_pos tag_activity_income']
```
ptbrowne: I think this is not correcly linted, can you please run `yarn lint --fix` ?

src/ducks/categorization/services.js
```diff
@@ -186,6 +204,37 @@ const getLocalClassifierOptions = transactionsWithManualCat => {
   }
 }
 
+export const atomicReweight = (classifier, category, word, frequencyCount) => {
+  const newFrequencyCount = 1 + Math.log(frequencyCount)
+  const deltaFrequencyCount = frequencyCount - newFrequencyCount
+  // update the right entries of the classifier's parameters
+  classifier.vocabulary[word] -= deltaFrequencyCount
+  classifier.wordCount[category] -= deltaFrequencyCount
```
ptbrowne: Is it normal that we are substracting a frequency from a count ?

src/ducks/categorization/services.js
```diff
@@ -186,6 +204,37 @@ const getLocalClassifierOptions = transactionsWithManualCat => {
...
+  classifier.wordFrequencyCount[category][word] = newFrequencyCount
+}
+
+export const reweightModel = classifier => {
+  // loop over categories in the wordFrequencyCat attribute
+  const wordFrequencyCount = classifier.wordFrequencyCount
+  // for each category
+  for (const category in wordFrequencyCount) {
+    if (wordFrequencyCount.hasOwnProperty(category)) {
```
ptbrowne: `for (const category in Object.keys(wordFrequencyCount))` will help get rid of the `if` and the extra nesting level :)

BTW, to help with unnecessary nesting, an early continue is clearer.

```
for () {
  if (!condition) { continue }

  // rest of the loop is not indented and deals with the happy case
}

File: src/ducks/categorization/services.js [hunk 7]
```diff
@@ -186,6 +204,37 @@ const getLocalClassifierOptions = transactionsWithManualCat => {
...
+      // extract its word-frequency count `wfc`
+      const categoryWordsFrequencyClounts = wordFrequencyCount[category];
+      // and search for tokens to reweight in it
+      TOKENS_TO_REWEIGHT.map(wordToReweight => {
+        if (categoryWordsFrequencyClounts.hasOwnProperty(wordToReweight)) {
+          // for every tokens to reweight : re-compute frequency count `fc`
+          const frequencyCount = categoryWordsFrequencyClounts[wordToReweight]
+          if (frequencyCount !== 1) {
+            atomicReweight(classifier, category, wordToReweight, frequencyCount)
```
ptbrowne: atomicReweight means `reweightWord` ?

src/ducks/categorization/services.js
francoisWeber: The habbit of typing "(Cozy) Cloud" ;) Thanks

src/ducks/categorization/services-fixtures.spec.js
francoisWeber: Yes it is the same file and my mistake for the hard-link

src/ducks/categorization/services.js
francoisWeber: Frequency in the statistical meaning : it's a count. Will double-check it anyway

francoisWeber:
Will re-submit the PR later : I tried to manage 2 different subjects here ...

PR #885, opened 2019-02-08
fix: Second safeguard for local categ by reweighting amounts tokens

francoisWeber:
This is the first part of a too large closed PR https://github.com/cozy/cozy-banks/pull/881

Added another safeguard to avoid a too high strenght of amounts tags in categorization. After having learnt the local model's parameter, we cast to log the sub-parameters associated to amounts.

### DIFF
--- a/src/ducks/categorization/services.js
+++ b/src/ducks/categorization/services.js
@@ -22,6 +22,26 @@ const FAKE_TRANSACTION = {
   label: 'thisisafaketransaction',
   manualCategoryId: '0'
 }
+const TOKENS_TO_REWEIGHT = [
+  'tag_neg',
+  'tag_v_b_expense',
+  'tag_neg tag_v_b_expense',
+  'tag_b_expense',
+  'tag_neg tag_b_expense',
+  'tag_expense',
+  'tag_neg tag_expense',
+  'tag_noise_neg',
+  'tag_neg tag_noise_neg',
+  'tag_pos',
+  'tag_noise_pos',
+  'tag_pos tag_noise_pos',
+  'tag_income',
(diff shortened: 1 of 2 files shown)


src/ducks/categorization/services.js
```diff
@@ -186,6 +206,35 @@ const getLocalClassifierOptions = transactionsWithManualCat => {
   }
 }
 
+export const reweightWord = (classifier, category, word, frequencyCount) => {
```
ptbrowne: Can you explain in the function description why this is necessary and how the reweighting works ? Why is log a good choice here ?

src/ducks/categorization/services.js
```diff
@@ -22,6 +22,26 @@ const FAKE_TRANSACTION = {
   label: 'thisisafaketransaction',
   manualCategoryId: '0'
 }
+const TOKENS_TO_REWEIGHT = [
```
ptbrowne: How those tags work ? How are they extracted ?

francoisWeber: It comes from here : https://github.com/cozy/cozy-banks/blob/90d9de15b3f8722b3fe2274b09c582df90ec54b4/src/ducks/categorization/services.js#L95
This list is the result of the combinations of (every possible signs tags) x (every possible amount tag)

Remark : it will soon be necessary to simplify them because their redundancy of information and the "amount tag" is sufficient.

francoisWeber: I just added a comment in code

src/ducks/categorization/services.js
francoisWeber: I added a doc to the function

PR #891, opened 2019-02-11
Rationalize account panels

drazik:
The goal of this PR is to group more account types than before. This way we don't have too much virtual groups that contain only one account type, and the types are more understandable than before.

I added some documentation so we don't lose the original types list.

https://testbanksmergeaccounttypes-banks.cozy.works/#/balances

### DIFF
--- a/docs/account-types.md
+++ b/docs/account-types.md
@@ -1,0 +1,116 @@
+# Account types
+
+Here is the complete list of the account types managed by the app:
+
+* Asset
+* Bank
+* Capitalisation
+* Cash
+* Checkings
+* ConsumerCredit
+* CreditCard
+* Credit Card
+* Deposit
+* Joint
+* Liability
+* LifeInsurance
+* Loan
+* Madelin
+* Market
+* Mortgage
+* PEA
+* PEE
+* RevolvingCredit
+* RSP
+* Savings
+* Other
+* Unkown
+* None
+* Credit card
+* Perco
+* Perp
+* Article83
+
+## Mapping
+
+Some types are dynamically mapped to other types:
+
+<table>
+  <thead>
+    <tr>
+      <th>Original type</th>
+      <th>Mapped type</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>`Article83`</td>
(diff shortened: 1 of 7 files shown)


docs/account-types.md
```diff
@@ -0,0 +1,116 @@
...
+      <td>`Deposit`</td>
+    </tr>
+    <tr>
+      <td>`ConsumerCredit`</td>
+      <td rowspan="2">`Loan`</td>
+    </tr>
+    <tr>
+      <td>`RevolvingCredit`</td>
+    </tr>
+    <tr>
```
kosssi: Maybe `Credit card` can be before `ConsumerCredit` to regroup Loan mapped type

drazik: Actually the second Loan was a mistake: it's Other. Thanks.

src/ducks/account/helpers.spec.js
```diff
@@ -37,21 +37,60 @@ describe('getAccountType', () => {
   it('should map Unkown/None to Other', () => {
     const accounts = [{ type: 'Unkown' }, { type: 'None' }, { type: 'Other' }]
 
-    expect(getAccountType(accounts[0])).toBe('Other')
-    expect(getAccountType(accounts[1])).toBe('Other')
-    expect(getAccountType(accounts[2])).toBe('Other')
+    for (const account of accounts) {
```
kosssi: Maybe you can use:

```accountsTypes = {
  'Other': ['Unkown', 'None', 'Other'],
  '...': [...]
}
```

docs/account-types.md
```diff
@@ -0,0 +1,116 @@
...
+<table>
+  <thead>
+    <tr>
+      <th>Original type</th>
+      <th>Mapped type</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>`Article83`</td>
```
ptbrowne: Nit: Backticks (`) do not work in HTML

src/ducks/account/helpers.spec.js
drazik: Thanks

PR #892, opened 2019-02-12
feat: Show konnector update info only on balance page

drazik:
Updated the konnctor update info block style so it can be nicely shown under the balance history chart. I removed the `Info` component because it was a duplicate from cozy-ui's one with some customisations. Now we have so much customisation that it's a totally different component. So I removed it and applied all needed styles to `KonnectorUpdateInfo` directly.

Mobile:

screen shot 2019-02-12 at 12 10 17

Desktop:

image


src/components/App.styl
```diff
@@ -0,0 +1,2 @@
+.App__content
+    flex-shrink 0 !important
```
ptbrowne: What is the problem otherwise here ?

src/components/KonnectorUpdateInfo/styles.styl
```diff
@@ -1,4 +1,31 @@
...
+
+    +small-screen()
+        padding-left 0.5rem
+        padding-right 0.5rem
+
+.KonnectorUpdateInfo__inner
+    padding 1.5rem
+    background-color var(--chablis)
+    border-radius 6px
+    text-align left
```
ptbrowne: Isn't it the default ?

src/components/KonnectorUpdateInfo/index.js
```diff
@@ -59,22 +60,28 @@ class KonnectorUpdateInfo extends React.PureComponent {
...
-          icon={<Icon icon="openwith" />}
-          label={t('KonnectorUpdateInfo.cta')}
-          theme="secondary"
-          href={url}
-          extension={breakpoints.isMobile ? 'full' : 'narrow'}
-        />
-      </Info>
+      <Padded className={styles.KonnectorUpdateInfo}>
+        <div className={styles.KonnectorUpdateInfo__inner}>
+          <SubTitle className={styles.KonnectorUpdateInfo__title}>
```
ptbrowne: className='u-monza'

src/components/KonnectorUpdateInfo/styles.styl
drazik: The parent element has `text-align: center` so we need to reset it.

src/components/App.styl
drazik: The header was not growing with its content. But it was a terrible solution, I found a better one by using `flex-basis` and `flex-shrink` on the header.

PR #897, opened 2019-02-12
fix: Bis: First fix for local categ of historical users

francoisWeber:
Due to merge conflicts and git mistakes by me, this is a replacement for previous PR 
https://github.com/cozy/cozy-banks/pull/872#

Description: 
Safeguard applied to the local categorization.

_Problem we try to solve:_
The local model is used on a label even if no tokens extracted from this label appears in the vocabulary. On some extreme cases (*), it can lead to a wrong categorization with high probability only based on the amount.
_Proposed solution:_
If less than `LOCAL_MODEL_PCT_TOKENS_IN_VOC_THRESHOLD` tokens extracted from the label appear in the local model vocabulary, then no prediction is done using the local model and its associated `(localCategoryId, localCategoryProba)` is set to a fallback value `('0', 0.10)`

(*) such extreme cases will be tackled in a future PR


francoisWeber:
conflicts again

PR #899, opened 2019-02-13
Improvements

ptbrowne:
- Groups beginning by 'é' would be at the end after "Z" groups. The deburr function of lodash is useful to remove diacritics from letters. Lowercasing is also necessary not to have `a, b, c, A, B, C` but `a, A, b, B, c, C` 

- Reduced the margin of the Figure in BalancePanel

- Add activated accounts as caption

- Increase touch area of group panel expand icon and switch

image

https://testbankssortdeburr-banks.cozy.works/#/balances


package.json
```diff
@@ -166,7 +166,7 @@
     "cozy-konnector-libs": "4.13.1",
     "cozy-pouch-link": "5.7.6",
     "cozy-stack-client": "5.6.1",
-    "cozy-ui": "18.13.1",
+    "cozy-ui": "^19.0.0",
```
kosssi: est ce que tu peux fixé le numéro de version

ptbrowne: j ai pas integré cozy-ui finalement.

PR #902, opened 2019-02-14
fix(mobile): Scroll after logout

drazik:
The problem was that after logout on the mobile app, we couldn't scroll anymore. This was caused by the fact that when we open the cozy-bar, it sets an `overflow: hidden` (and other stuff to prevent scroll) on the `body`. But on logout, we simply destroy the cozy-bar DOM node. So we still had the `overflow: hidden` on the `body`. Now I just erase all styles applied on `body` on logout, but we may think of a better API like `destroy` or `unmount` directly in the cozy-bar to handle this correctly.


drazik:
See https://github.com/cozy/cozy-bar/issues/460

PR #907, opened 2019-02-15
perf: Smaller build 1

ptbrowne:
Bigger issue : https://github.com/cozy/cozy-banks/issues/926

Baseline: 2421ko

image

- Remove duplicate lib via webpack aliasing : 2317ko

image

- Do not bundle all d3 inside build : 2179ko

image


nono:
Some other ideas:

- remove the Fetch polyfill (it is no longer needed for the browsers we support)
- use lazy loading for the graphs libs
- redux-raven-middleware and raven.js seems to load the same files: are both necessary?
- try svgo on icon-collect-account.svg

nono:
> Cozy Authentication build ? Why is it so big ?

It embeds 65ko of CSS from cozy-ui, and some stylus things

kosssi:
I don’t see the interest in minifying image, it should be webpack that handles this, no ?

nono:
Well, there are some pros and cons to do that. For example, minifying images can take a lot of time and make the build really slower.

ptbrowne:
> I don’t see the interest in minifying image, it should be webpack that handles this, no ?

Since

PR #909, opened 2019-02-15
feat(mobile): Ensure the token is refreshed before it expires

drazik:
This is a temporary solution to prevent the bar from having an expired
token. See https://github.com/cozy/cozy-bar/issues/449 for more
informations.


src/main.jsx
```diff
@@ -68,6 +69,34 @@ const setupApp = async persistedState => {
...
+
+      if (!issuedAt || !isValid(issuedAt) || isBefore(issuedAt, sixDaysAgo)) {
+        try {
+          client.stackClient.refreshToken()
+        } catch (err) {
+          // eslint-disable-next-line no-console
+          console.error('Error while refreshing token:' + err)
+        }
+      }
+    })
```
ptbrowne: Can you add a test ?

PR #912, opened 2019-02-18
feat: Add lamutuellegeneral to brand dictionnary

LucsT:
Adding a brand for an important connector.
No \\b at the end because know identifiers is 'LA MUTUELLE GEN' but we probably want to match something like 'LA MUTUELLE GENERALE' too if it happens.


LucsT:
Connector will be in production today.

PR #917, opened 2019-02-18
fix(Error): Refresh link doesn't fail with CSP

drazik:
The inline onclick handler was failing with real stack CSPs.

### DIFF
--- a/src/components/ErrorBoundary/Error.jsx
+++ b/src/components/ErrorBoundary/Error.jsx
@@ -3,36 +3,74 @@ import PropTypes from 'prop-types'
 import styles from './Error.styl'
 import { Empty, translate } from 'cozy-ui/react'
 import brokenIcon from 'assets/icons/icon-broken.svg'
+import { setBarTheme } from 'ducks/mobile/utils'
 
-const Error = props => {
-  const { t } = props
-  const update = t('Error.update')
-    .replace('#{LINK}', `<a onClick="javascript:window.location.reload(true)">`)
-    .replace('#{/LINK}', '</a>')
-  const lang = 'fr'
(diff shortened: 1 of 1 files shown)


src/components/ErrorBoundary/Error.jsx
```diff
@@ -1,38 +1,56 @@
...
-          />
-        }
-      />
-    </div>
-  )
-}
+class Error extends React.Component {
+  static propTypes = {
+    t: PropTypes.func.isRequired
+  }
```
ptbrowne: TIL static propTypes 👍

src/components/ErrorBoundary/Error.jsx
```diff
@@ -1,38 +1,56 @@
...
+            />
+          }
+        />
+      </div>
+    )
+  }
 }
 
-export default translate()(Error)
+export default withRouter(translate()(Error))
```
ptbrowne: Tu as encore besoin du withRouter ?

src/components/ErrorBoundary/Error.jsx
```diff
@@ -1,38 +1,56 @@
...
+  static propTypes = {
+    t: PropTypes.func.isRequired
+  }
+
+  rootRef = React.createRef()
+
+  componentDidMount() {
+    this.refreshLink = this.rootRef.current.querySelector('#error-refresh-link')
+    this.refreshLink.addEventListener('click', this.refresh)
+  }
```
ptbrowne: I am afraid this will not work for subsequent updates ? Maybe you have to removeEventListener and re-add also in the componentDidUpdate ?

PR #922, opened 2019-02-21
Select date overlaps history chart on movements page

drazik:
In this case, screenshots are better than words:

image

image

Live on https://testbanksmovementsgradient-banks.cozy.works

### DIFF
--- a/src/components/Chart/LineChart.jsx
+++ b/src/components/Chart/LineChart.jsx
@@ -285,14 +285,19 @@ class LineChart extends Component {
 
   updatePointLine() {
     const item = this.getSelectedItem()
-    const { height, margin, tickPadding } = this.props
+    const { height, margin, tickPadding, showAxis } = this.props
     const x = this.x(item.x)
+    let y2 = height - margin.top + tickPadding
+
+    if (showAxis) {
+      y2 -= margin.bottom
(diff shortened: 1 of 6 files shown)


src/components/Chart/LineChart.jsx
```diff
@@ -130,6 +131,7 @@ class LineChart extends Component {
     }
 
     this.y.domain(yDomain)
+    this.y.domain(yDomain)
```
ptbrowne: You can remove this I guess :)

src/components/Chart/LineChart.jsx
```diff
@@ -180,11 +182,27 @@ class LineChart extends Component {
...
-        .attr('width', this.getInnerWidth())
-        .attr('height', this.getInnerHeight())
+        if (!this.props.showAxis) {
+          min += this.props.margin.bottom
+        }
+
+        return min
+      })
+
+      let height = this.getInnerHeight()
```
ptbrowne:
```
const innerHeight  = this.getInnerHeight()
const height = innerHeight + (this.props.showAxis ? 0 : this.props.marginBottom)
```
This lets us get rid of the let var.

innerHeight can be used directly in the y0 function.

PR #940, opened 2019-02-26
feat(TransactionRow): Add show-transactions-ids flag

drazik:
By enabling this flag, the id of each transaction is shown. This is
useful for debug purposes.

Particularly, since demo fixtures transactions dates are now dynamic, it can be hard to mentally do the link between a transaction shown in the app and a transaction in the JSON file. With this, we can link them using the ID, which is more straightforward.

image

image


kosssi:
After discussion we would like simplify it, to not have css just for debug mode

drazik:
LGTM (can't approve since I opened the PR)

PR #961, opened 2019-02-28
fix(TransactionsPage): Update current month when changing filtering doc

drazik:
When changing the filtering doc while on a specific month, we didn't
update `limitMin` and `limitMax`. But it could be wrong since the number
of transactions changed. Now, on update we trigger `handleChangeMonth`,
which recomputes the limits and the current month.

https://testbanksfixtransactions-banks.cozy.works

Do the following to test:

* Go to an account details
* Select "all accounts"
* Go back one month
* Select "awaiting reimbursements"
* You see the transactions (before the fix, you couldn't see it)

### DIFF
--- a/src/ducks/filters/index.js
+++ b/src/ducks/filters/index.js
@@ -14,8 +14,9 @@ const FILTER_BY_DOC = 'FILTER_BY_DOC'
 const RESET_FILTER_BY_DOC = 'RESET_FILTER_BY_DOC'
 
 // selectors
-export const getPeriod = state => state.filters.period
-export const getFilteringDoc = state => state.filters.filteringDoc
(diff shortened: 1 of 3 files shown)


ptbrowne:
Nice, is there already tests on handleChangeMonth ? Maybe we can add a test checking that handleChangeMonth is correctly called when filtering doc is changed ?

drazik:
There's a problem with the tests. As soon as I call `root.setProps()`, I get a `TypeError: Cannot read property 'params' of undefined` at line 140 of `TransactionHeader.jsx`, even if this component is wrapped in a `withRouter` HOC, therefore it should receive the `router` prop. If you have an idea please let me know because I can't figure it out.

PR #971, opened 2019-03-01
Use the color from injected CSS variables

drazik:
We were using custom variable names but the good one are the ones used in [cozy-ui](https://github.com/cozy/cozy-ui/blob/ff96bcb4ccab339a053e12029e652630a1f56c9c/stylus/settings/palette.styl#L150-L152). So now we use these names, making the app more customizable.

See https://testbankscolorvars-banks.cozy.works


drazik:
I merge as test as failing for a totally unrelated reason.

PR #974, opened 2019-03-01
feat(bar): Handle bar theme overrides

drazik:
For customization needs, we need to be able to pass overrides to
`cozy.bar.setTheme`. Do be able to override it easily, I added a map in
its own file. This way, we can override this single file in every
customization. In the original Banks we don't want any overrides, so it
is an empty object.

See cozy/cozy-bar#473 for the implementation of
the overrides in the cozy-bar side.

### DIFF
--- a/src/ducks/mobile/utils.js
+++ b/src/ducks/mobile/utils.js
@@ -1,5 +1,6 @@
 /* global cozy, __TARGET__ */
 import { setTheme as setStatusBarTheme } from './statusBar'
+import barOverrides from 'ducks/bar/overrides'
 
 const getLang = () =>
   navigator && navigator.language ? navigator.language.slice(0, 2) : 'en'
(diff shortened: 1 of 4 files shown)


src/ducks/mobile/utils.js
```diff
@@ -41,6 +42,7 @@ export const setBarTheme = theme => {
   }
 
   if (cozy.bar && cozy.bar.setTheme) {
-    cozy.bar.setTheme(theme)
+    const overrides = barOverrides[theme]
+    cozy.bar.setTheme(theme, overrides)
```
ptbrowne: If the theme does not exist, it should throw ? What happens ?

drazik: IMHO it should set the default theme but not fail. We can add a warning in the console.

ptbrowne: OK for me.

drazik: But it should be implemented in the cozy-bar since it knows its available themes.

PR #979, opened 2019-03-04
Use CSS variables for header

drazik:
By introducing CSS variables for each part of the header, this is much easier to apply another theme to it by creating an override of the `src/styles/palette.styl` file.

### DIFF
--- a/src/styles/main.styl
+++ b/src/styles/main.styl
@@ -1,10 +1,10 @@
 @require 'generic/normalize.styl'
 @require 'elements/defaults.styl'
-@require 'utilities.styl'
-@require 'text.styl'
+@require '~styles/utilities.styl'
+@require '~styles/text.styl'
 @require 'settings/breakpoints'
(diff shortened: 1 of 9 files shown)


src/styles/main.styl
```diff
@@ -1,10 +1,10 @@
 @require 'generic/normalize.styl'
 @require 'elements/defaults.styl'
-@require 'utilities.styl'
-@require 'text.styl'
+@require '~styles/utilities.styl'
+@require '~styles/text.styl'
 @require 'settings/breakpoints'
 @require '~styles/variables.styl'
-@require 'palette.styl'
+@require '~styles/palette.styl'
```
ptbrowne: To me ~ meant node_modules ? Is it the case here ?

ptbrowne: Explained via mattermost : ~ mean any directory in resolve.modules since we use webpack resolver.

edas:
:+1: pour la méthode

PR #980, opened 2019-03-05
Optimistic update of BalancePanels toggling

ptbrowne:
Since cozy-client does not do optimistic update, we have
to implement it in the component, otherwise, when toggling
a BalancePanel, it takes some time to update (waiting for
the server to respond).


package.json
```diff
@@ -20,6 +20,7 @@
     "prebuild:mobile": "npm run clean:mobile",
     "prewatch:mobile": "npm run clean:mobile",
     "lint": "npm-run-all --parallel 'lint:*'",
+    "lint:js:modified": "eslint --fix $(git ls-files -m | grep -E 'jsx?$')",
```
drazik: What about a precommit hook using [lint-staged](https://github.com/okonet/lint-staged)? Still skippable via `--no-verify` for cases in which we want to commit anyway.

src/ducks/balance/components/GroupPanel.jsx
```diff
@@ -93,6 +98,26 @@ class GroupPanel extends React.PureComponent {
     e.stopPropagation()
   }
 
+  componentDidUpdate(prevProps) {
+    // We have to revert optimisticExpanded if for example
+    // the network has failed
+    if (this.props.expanded !== prevProps.expanded) {
+      this.setState({ optimisticExpanded: this.props.expanded })
+    }
```
drazik: It means that if I expand a panel it actually expands, but will automatically collapse few seconds after if the network failed ?

ptbrowne: You're right, and this is not actually the behavior we want. Since the data criticity is not high, I removed this bit of code. This is implemented (and tested 🎉 ) in https://github.com/cozy/cozy-banks/pull/980/commits/e53e6c9fd266d77fd07ba29351ae58ba7e70b123.

package.json
ptbrowne: True I forgot about lint-staged. I'm removing this from the PR.

PR #981, opened 2019-03-05
chore: Add lint-staged

ptbrowne:
Not yet added as a husky hook

Linting of partially staged files seem to work well.

Since lint-staged is still quite slow, I am unconfortable
to add it as a husky hook yet and would prefer to have it
optional (see if it works well for us).

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "imports-loader": "0.8.0",
     "jest": "23.6.0",
     "json-loader": "0.5.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -119,6 +119,7 @@
     "imports-loader": "0.8.0",
     "jest": "23.6.0",
     "json-loader": "0.5.7",
+    "lint-staged": "^8.1.5",
```
kosssi: can you pined version?

PR #982, opened 2019-03-05
Use CSS variable in header

drazik:
Use CSS variable in header to be able to customize it via overrides. I also did some refactoring to remove unneeded custom styles.

See https://testbanksheadervariables-banks.cozy.works


src/ducks/transactions/header/TableHead.jsx
```diff
@@ -21,7 +20,7 @@ class TableHead extends React.PureComponent {
     }
 
     return (
-      <Table className={styles[`TableHead_${color}`]}>
+      <Table color={color}>
```
ptbrowne: TIL color is supported by Table

PR #991, opened 2019-03-06
refactor: Split cozy-ui imports

drazik:
This makes it easier to override components in customizations. Without
that, the customization would have to also override the `react/index.js`
file from cozy-ui. But doing this would force the customization to
maintain this file so it always reflects cozy-ui's original one.

You can see https://testbankssplitcozyuiimports-banks.cozy.works if you want to check that nothing breaks with these changes.


src/ducks/account/AccountSwitch.jsx
```diff
@@ -6,15 +6,10 @@ import { connect } from 'react-redux'
...
-import {
-  translate,
-  withBreakpoints,
-  Icon,
-  Media,
-  Bd,
-  Img,
-  Overlay
-} from 'cozy-ui/react'
+import { translate, withBreakpoints } from 'cozy-ui/react'
```
kosssi: finalement on gagne des lignes ^^

src/ducks/balance/components/BalanceAccounts.jsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
-import { translate, Button } from 'cozy-ui/react'
+import { translate } from 'cozy-ui/react'
+import Button from 'cozy-ui/react'
```
ptbrowne:
```suggestion
import Button from 'cozy-ui/react/Button'
```

src/ducks/transactions/actions/HealthLinkAction.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
-import { translate, ButtonAction } from 'cozy-ui/react'
+import { translate } from 'cozy-ui/react'
+import ButtonAction from 'cozy-ui/react'
```
ptbrowne:
```suggestion
import ButtonAction from 'cozy-ui/react/ButtonAction'
```

src/ducks/balance/components/BalanceAccounts.jsx
drazik: Hawk eyes :eye:

PR #992, opened 2019-03-06
fix(deps): Update cozy-stack-client to v6.6.0

drazik:


drazik:
Actually since this dependency is not required anymore, we just remove it. See https://github.com/cozy/cozy-banks/pull/994

PR #998, opened 2019-03-07
fix: Better perf for toggling expansion panels

ptbrowne:
Debounce has been introduced not to re-render the whole
BalancePage when toggling expansion panels.

https://testbanksdebouncedpanels-banks.cozy.works

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -1,5 +1,6 @@
 import React, { PureComponent, Fragment } from 'react'
-import { flowRight as compose, get, sumBy, set } from 'lodash'
+import { flowRight as compose, get, sumBy, set, debounce } from 'lodash'
+
 import { queryConnect, withMutations } from 'cozy-client'
 import flag from 'cozy-flags'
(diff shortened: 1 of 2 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -72,10 +72,10 @@ class Balance extends PureComponent {
       set(nextState.panels, path, expanded)
 
       return nextState
-    }, this.onPanelsStateChange)
+    }, this.savePanelState)
   }
 
-  onPanelsStateChange() {
+  savePanelState() {
```
drazik: :+1:

src/ducks/balance/Balance.jsx
```diff
@@ -29,8 +30,15 @@ class Balance extends PureComponent {
...
 
     setBarTheme('primary')
+    this.handlePanelChange = this.handlePanelChange.bind(this)
+    this.debouncedHandlePanelChange = debounce(this.handlePanelChange, 3000, {
+      leading: false,
+      trailing: true
+    }).bind(this)
   }
 
+  componentDidMount() {}
```
drazik: We can remove this if it does nothing.

kosssi: is unnecessary?

PR #1008, opened 2019-03-10
fix: Use NODE_ENV==test to deactivate icons in jest

ptbrowne:
`require.context` does not exist in runtime so icons where not displayed. `require.context` was used to detect if we were in tests, so NODE_ENV==test can be used also.


ptbrowne:
Yes I meant to include it into the release branch then I saw that it was not yet included.

PR #1010, opened 2019-03-11
fix(HistoryChart): Scroll to end of chart only if possible

drazik:
On some old browsers, `Element.scrollTo` is not supported. See
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo for
more informations.


drazik:
Closing in favor of #1011

PR #1012, opened 2019-03-11
test: Added a column identifying the cozy concerned by the fixtures

francoisWeber:
Added a column in the resulting `.csv` allowing to identify to which cozy instance belongs an operation/categorization.

### DIFF
--- a/src/ducks/categorization/services-fixtures.spec.js
+++ b/src/ducks/categorization/services-fixtures.spec.js
@@ -81,7 +81,8 @@ const setCsvWriter = () => {
       { id: 'amount', title: 'Amount' },
       { id: 'label', title: 'Label' },
       { id: 'catNameDisplayed', title: 'Category displayed' },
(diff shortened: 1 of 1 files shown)


src/ducks/categorization/services-fixtures.spec.js
```diff
@@ -426,7 +418,7 @@ xOrDescribe('Chain of predictions', () => {
         // Format results
         const fmtedResults = fmtResults(results)
         // Format results for the historized CSV
-        const fixtureCSV = fmtResultsCSV(results)
+        const fixtureCSV = fmtResultsCSV(results, bank)
```
ptbrowne: It's written bank here but in the signature of fmtResultsCSV, it's written cozyId  ?

francoisWeber: You're right. Actually the `banks/bank` naming had no meaning at all in this context. I replaced it with `cozyInstances/cozyId`

PR #1015, opened 2019-03-11
No account

kosssi:
This PR should display new onboarding. This allow to navigate on all application without data bank

You can check [demo here](https://testbanksnoaccount-banks.cozy.works)


src/ducks/balance/components/NoAccount.styl
```diff
@@ -0,0 +1,27 @@
+.NoAccount
+    height calc(50vh - 48px)
```
ptbrowne: ? What is this from ? Is No Account really the half height of the screen ? Shouldn't we have an absolute size ?

kosssi: On the onboarding page, we want the header make 50%, this allow don't have a white space too big after.

PR #1031, opened 2019-03-13
Importing accounts

kosssi:
Le but de cette PR est d'affiche un écran spécifique sur la page Solde Actuel avec les comptes entrain d'être importé.
On peut voir une démo [ici](https://testbanksimportingaccounts-banks.cozy.works/#/balances). 

Il me reste 3 choses à faire:
- [x] Supprimer le clique sur les `ExpansionPanel`
- [x] Vérifier une traduction du bouton du bas qui est en anglais (Add a Bank)...
- [x] Implémenter la fonction all pour TriggerCollection dans cozy-client [ici](https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/TriggerCollection.js#L32)

Vu qu'il n'est pas possible d'injecter des fixtures j'ai ajouté directement dans le code ceci:
```js
    const triggers = [
      {
        _id: 'd8690a8947ebc934f81df23ec52395a1',
        domain: 'gangsters.cozy.rocks',
        prefix: 'gangsters.cozy.rocks',
        type: '@cron',
        worker: 'konnector',
        arguments: '0 52 2 * * *',
        debounce: '',
        options: null,
        message: {
          konnector: 'societegenerale',
          banking: true,
          account: 'e02702daad059578a50cf9f78e24ebc8'
        }
      },
      {
        _id: 'd8690a8947ebc934f81df23ec52395a1',
        domain: 'gangsters.cozy.rocks',
        prefix: 'gangsters.cozy.rocks',
        type: '@cron',
        worker: 'konnector',
        arguments: '0 52 2 * * *',
        debounce: '',
        options: null,
        message: {
          konnector: 'bnpparibas82',
          banking: true,
          account: 'e02702daad059578a50cf9f78e24ebc8'
        }
      }
    ]
```


src/ducks/balance/Balance.jsx
```diff
@@ -123,19 +128,21 @@ class Balance extends PureComponent {
...
-    if (
-      isCollectionLoading(accountsCollection) ||
-      isCollectionLoading(groupsCollection)
-    ) {
+    const collections = [
+      accountsCollection,
+      groupsCollection,
+      triggersCollection
+    ]
+    if (some(collections, isCollectionLoading)) {
```
drazik: Nit: I would use [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here as it's the same thing, but native:

```suggestion
    if (collections.some(isCollectionLoading)) {
```

src/ducks/balance/components/AccountsImporting.jsx
```diff
@@ -0,0 +1,115 @@
...
+          </Padded>
+        </VerticalBox>
+      </Header>
+      <Padded className="u-pt-0">
+        <BalancePanels
+          groups={groups}
+          panelsState={panelsState}
+          warningLimit={0}
+          onSwitchChange={emptyFct}
+          onPanelChange={emptyFct}
```
drazik: What about checking in `BalancePanels` if the props is a function? So if we pass nothing, it's not called. It avoids an empty functions + function calls to this empty functions.

src/ducks/balance/components/AccountsImporting.styl
```diff
@@ -0,0 +1,24 @@
+@require 'settings/breakpoints'
+
+$header-margin-bottom = -28px
+
+.content
+    height 245px
+    color white
```
drazik: Can you use a CSS variable here? Not necesarily a new one, but something like `primaryContrastTextColor`?

src/ducks/balance/components/GroupPanel.jsx
```diff
@@ -154,7 +154,7 @@ class GroupPanel extends React.PureComponent {
             disableRipple: true
           }}
           className={cx({
-            [styles['GroupPanelSummary--unchecked']]: !checked
+            [styles['GroupPanelSummary--unchecked']]: !checked && !group.loading
```
drazik: Maybe this class name is not relevant anymore if it's used not only to show that a panel is unchecked, but also when it's loading.

src/locales/en.json
```diff
@@ -65,7 +65,6 @@
     "solde": "Balance",
     "account_number": "Account number",
     "account_name": "Account name",
-    "add_bank": "Add a bank",
```
drazik: It was already at the right place in french translations?

src/ducks/balance/components/AccountRowLoading.jsx
```diff
@@ -8,7 +9,7 @@ import styles from './AccountRow.styl'
 import stylesLoading from './AccountRowLoading.styl'
 
 const AccountRowLoading = ({ t, institutionSlug }) => (
-  <li className={styles.AccountRow}>
+  <li className={cx(styles.AccountRow, stylesLoading.noLink)}>
```
drazik: The base component should have the `u-clickable` class, and this one should remove it. This way we don't add a class for this purpose.

src/locales/en.json
kosssi: yes I clean it [here](https://github.com/cozy/cozy-banks/commit/891b25b54897dfd4fd20911cfda9ec7ea80403e8). I will fixup this commit ;)

src/ducks/balance/components/GroupPanel.jsx
kosssi: I create a variable `const isUncheckable = !group.loading` which allow to keep a coherence

PR #1037, opened 2019-03-14
Remove categorization

drazik:
We now do the categorization in banking connectors. For that, we extracted the categorization logic to cozy-konnector-libs (see https://github.com/konnectors/libs/pull/429).

The goal is to avoid a time interval during which the user has imported his banking data, but we did'nt categorized it via our `onOperationOrBillCreate` service (since there is a 3mn debounce between data creation and service execution). This led to bad behavior in banks like:

* The user imports his data, then open banks
* He sees his transactions with some categories
* When he opens a transaction, the category may change, since the service categorized it meanwhile

The only thing related to categorization that remains in cozy-banks (at least for now) is the test dashboard, because it uses some cozy-banks internal things (like the category tree) that I didn't want to duplicate in cozy-konnector-libs. We may want to change this someday.


kosssi:
You should update `cozy-konnector-libs` when the version is on NPM to fix test ;)

drazik:
Yes I forgot to say that it's the only remaining task. Thanks

PR #1039, opened 2019-03-15
docs: Add doc about categorization dashboard

drazik:
Add some documentation about categorization tests dashboard.

### DIFF
--- a/docs/categorization-dashboard.md
+++ b/docs/categorization-dashboard.md
@@ -1,0 +1,50 @@
+# Categorization dashboard
+
+To check that the categorization works well, we have tests that generate a dashboard showing global and local models performances.
+
+## Run tests
+
+Since these tests are executed on real data, the data is encrypted. To be able to run the tests, you have to decrypt it:
+
+```
+yarn decrypt-banking-tests
+```
+
+The global categorization model tests also needs the current and latest model parameters. These parameters should be downloaded:
+
+```
+yarn download-banking-tests
+```
+
+Now you have everything to run the tests.
+
+```
+env BACKUP_DIR=/path/to/dir yarn jest src/ducks/categorization/services-fixtures.spec.js
+```
+
+* `BACKUP_DIR` environment variable is the path to the directory in which you want to write the result of the tests (csv and txt files)
+* You can add the jest flags you need: `-u` to update snapshots and `--watch` to run the tests in watch mode, for example
+
+## Work on models
+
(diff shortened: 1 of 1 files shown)


docs/categorization-dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Run tests
+
+Since these tests are executed on real data, the data is encrypted. To be able to run the tests, you have to decrypt it:
+
+```
+yarn decrypt-banking-tests
+```
+
+The global categorization model tests also needs the model parameters. These parameters should be downloaded:
```
francoisWeber: To avoid any question like "why don't you embed the model's parameter in the repo", i would have been slightly more precise : "The global categorization model tests also needs the **current and latest** model parameters

docs/categorization-dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+
+```
+yarn download-banking-tests
+```
+
+Know you have everything to run the tests.
```
francoisWeber: Know/Now ;)

docs/categorization-dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+```
+env BACKUP_DIR=/path/to/dir yarn jest src/ducks/categorization/services-fixtures.spec.js
+```
+
+* `BACKUP_DIR` environment variable is the path to the directory in which you want to write the result of the tests (csv and txt files)
+* You can add the jest flags you need: `-u` to update snapshots and `--watch` to run the tests in watch mode, for example
+
+## Work on models
+
+The models are extracted in [cozy-konnector-libs](https://github.com/konnectors/libs/tree/master/packages/cozy-konnector-libs/src/libs/categorization). So if you need to update them and check their performance with the dashboard, you have to work with a linked cozy-konnector-libs.
```
francoisWeber: I think "models implementations" rather than just "models"  could avoid a confusion between "model parameter" and "model".

docs/categorization-dashboard.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+Then, link you local cozy-konnector-libs:
+
+```
+# Inside cozy-konnector-libs
+yarn link
+
+# Inside cozy-banks
+yarn link cozy-konnector-libs
```
francoisWeber: `yarn link` shows an outputs that uses quoted importation, like 
`yarn link "cozy-konnector-libs"`
Dunno if it is necessary or not but ... just in case

drazik: It also works without quotes, but thanks ;)

PR #1053, opened 2019-03-19
feat: Auto user agent

ptbrowne:
Automatically ensures the user agent in config.xml corresponds
to the version in package.json

Done via a Cordova hook before preparing.


ptbrowne:
Hooks are executed in alphabetical order. The number is here in case we need more. I've seen an example with this and kept it.

PR #1058, opened 2019-03-20
[epic] Pin epic

ptbrowne:
Epic PR.

- [x] After 30s of inactivity, the app is locked on mobile
- [x] It can be disabled via fingerprint
- [ ] It can be disabled via pin
- [ ] The pin can be set up in the settings


ptbrowne:
Closing since we favor flags instead of long running branches.

PR #1069, opened 2019-03-21
Add maintenance property in brand dictionnary and KonnectorAction

drazik:
The KonnectorAction (« My invoices » CTA) now checks if the brand is in maintenance mode or not. This state can be specified in the brand dictionnary by adding the `maintenance: true` property to a brand.

### DIFF
--- a/src/ducks/brandDictionary/index.spec.js
+++ b/src/ducks/brandDictionary/index.spec.js
@@ -1,4 +1,4 @@
-import { getBrands, matchBrands, findMatchingBrand } from '.'
+import { getBrands, matchBrands, findMatchingBrand, isMatchingBrand } from '.'
 import brands from './brands'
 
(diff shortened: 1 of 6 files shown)


src/ducks/brandDictionary/index.spec.js
```diff
@@ -13,6 +13,18 @@ describe('brandDictionary', () => {
...
+    const [cpam] = brands.filter(brand => brand.name === 'Ameli')
+
+    it('should return true if the brand and the label match', () => {
+      expect(isMatchingBrand(cpam, 'Remboursement CPAM')).toBe(true)
+    })
+
+    it("should return false if the brand and the label don't match", () => {
+      expect(isMatchingBrand(cpam, 'Pouet pouet')).toBe(false)
+    })
+  })
```
ptbrowne: <3

PR #1079, opened 2019-03-22
fix: Check for revocation would trigger false positives

ptbrowne:
Shame, shame...

I forgot the indexOf() > -1

### DIFF
--- a/src/ducks/client/mobile.js
+++ b/src/ducks/client/mobile.js
@@ -48,12 +48,12 @@ export const getManifestOptions = manifest => {
   }
 }
 
-const isRevoked = async client => {
+export const isRevoked = async client => {
   try {
     await client.stackClient.fetchInformation()
     return false
   } catch (err) {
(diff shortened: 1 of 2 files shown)


src/ducks/client/mobile.js
```diff
@@ -48,12 +48,12 @@ export const getManifestOptions = manifest => {
...
 }
 
-const isRevoked = async client => {
+export const isRevoked = async client => {
   try {
     await client.stackClient.fetchInformation()
     return false
   } catch (err) {
-    if (err.message && err.message.indexOf('Client not found')) {
+    if (err.message && err.message.indexOf('Client not found') > -1) {
```
nono: By the way, when the code is `if (condition) { return true } else { return false }`, I prefer `return conditon` (or `return !!condition` for forcing a boolean).

kosssi:
```
return includes(err.message, 'Client not found')
```

PR #1081, opened 2019-03-22
Travis build perf

drazik:
The yarn cache seems to be very big. Disabling it makes the cache goes
from ~1GB to ~200MB. The Travis doc warns about big cache archives:

"If you store archives larger than a few hundred megabytes in the cache,
it is unlikely that you’ll see a big speed improvement." (https://docs.travis-ci.com/user/caching/#how-does-caching-work)

I think the time to download/unpack this cache was longer than
installing dependencies without it, since the build time went from ~12mn
with it to ~8mn without it.

Before: https://www.travis-ci.org/cozy/cozy-banks/builds/509975813
After: https://www.travis-ci.org/cozy/cozy-banks/builds/509994019 / https://www.travis-ci.org/cozy/cozy-banks/builds/510004629


ptbrowne:
Woot 30% improvement !

PR #1087, opened 2019-03-25
refactor(AddAccountButton): Create button component

kosssi:
To facilitate customization


src/ducks/categories/AddAccountButton.jsx
```diff
@@ -0,0 +1,18 @@
+import React, { memo } from 'react'
+import Bouton from 'cozy-ui/react/Button'
```
drazik: Can you rename it to `Button`?

PR #1101, opened 2019-03-27
Inject CozyClient in Bar

gregorylegarec:
Some custom versions of the CozyBar are needing a injected CozyClient.

### DIFF
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -66,6 +66,7 @@ const setupApp = async persistedState => {
   if (__TARGET__ !== 'mobile') {
     cozy.bar.init({
       appName: data.cozyAppName,
+      cozyClient: client,
       iconPath: data.cozyIconPath,
       lang: data.cozyLocale,
       replaceTitleOnMobile: true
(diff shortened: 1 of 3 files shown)


ptbrowne:
> Some custom versions of the CozyBar

Is banks using a custom version ? I do not understand.

src/main.jsx
```diff
@@ -66,6 +66,7 @@ const setupApp = async persistedState => {
   if (__TARGET__ !== 'mobile') {
     cozy.bar.init({
       appName: data.cozyAppName,
+      cozyClient: client,
```
drazik: We want to do this only for web app?

gregorylegarec: No, is there somewhere else to inject cozy Client ?

gregorylegarec: Thanks for pointing out this, I've just updated the commit.

drazik: Yes, here: https://github.com/cozy/cozy-banks/blob/master/src/ducks/mobile/utils.js#L24

PR #1102, opened 2019-03-27
feat(Balance): Fetch triggers or accounts when is necessary

kosssi:
Nous utilisons actuellement la lib cozy-realtime directement dans le dépot. Lorsque la nouvelle API sera sorti on pourra utiliser cozy-realtime directement depuis npm.

Relevant PR for realtime new API : https://github.com/cozy/cozy-libs/pull/355

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -3,7 +3,13 @@ import { flowRight as compose, get, sumBy, set, debounce } from 'lodash'
 
 import { queryConnect, withMutations, withClient } from 'cozy-client'
 import flag from 'cozy-flags'
-import { groupsConn, settingsConn, triggersConn, accountsConn } from 'doctypes'
+import {
+  groupsConn,
+  settingsConn,
+  triggersConn,
+  accountsConn,
+  ACCOUNT_DOCTYPE
+} from 'doctypes'
 import cx from 'classnames'
 
 import { connect } from 'react-redux'
@@ -27,6 +33,7 @@ import BalancePanels from './BalancePanels'
 import { getPanelsState } from './helpers'
 import BarTheme from 'ducks/mobile/BarTheme'
 import { filterByAccounts } from 'ducks/filters'
+import { CozyRealtime } from 'cozy-realtime'
 
 const INTERVAL_DURATION = 5000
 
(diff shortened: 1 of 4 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -136,6 +143,54 @@ class Balance extends PureComponent {
     })
   }
 
+  startRealtime(type, callback) {
+    if (!this.realtime[type]) {
+      const cozyClient = this.props.client
+      const url = cozyClient.client.uri
+      const token = cozyClient.client.token.token
+      this.realtime[type] = new CozyRealtime({ url, token })
+      this.realtime[type].subscribe({ type }, 'created', callback)
```
ptbrowne: I think we need only one CozyRealtime and not one CozyRealtime per doctype.

Ideally it should go like this

cozyClient.realtime.subscribe(doctype, token)

We should not care about cozyClient.client.uri  and .token.token inside the component, the initialisation of CozyRealtime from cozy-client should be extracted from the component as it will surely move and be refactored. The component does not care about that.

src/ducks/balance/Balance.jsx
```diff
@@ -136,6 +143,54 @@ class Balance extends PureComponent {
...
+    }
+  }
+
+  stopRealtime(type, callback) {
+    if (this.realtime[type]) {
+      this.realtime[type].unsubscribe({ type }, 'created', callback)
+      delete this.realtime[type]
```
ptbrowne: Is the websocket closed when every unsubscribe has been made ?

src/ducks/balance/Balance.jsx
```diff
@@ -136,6 +143,54 @@ class Balance extends PureComponent {
...
+    }
+  }
+
+  fetchTriggers() {
+    const client = this.props.client
+    client.query(triggersConn.query(client))
+  }
+
+  startFetchTriggersInterval() {
```
ptbrowne: I think interval is no longer the right word.

src/ducks/balance/Balance.jsx
```diff
@@ -136,6 +147,63 @@ class Balance extends PureComponent {
...
+    this.createRealtime()
+    if (!this.realtimeStatus[type]) {
+      this.realtimeStatus[type] = true
+      this.realtime.subscribe({ type }, 'created', callback)
+    }
+  }
+
+  stopRealtime(type, callback) {
+    if (this.realtimeStatus[type]) {
+      this.realtimeStatus[type] = false
```
ptbrowne: Put `false` when you're sure unsubscribe has worked, below th next line ;)

src/ducks/balance/Balance.jsx
kosssi: not at the moment

kosssi: I'm writing a todo list with this issue on https://github.com/cozy/cozy-libs/pull/355

PR #1103, opened 2019-03-27
Pin part iv

ptbrowne:
Add a page to the pin onboarding to let the user choose if (s)he wants to use fingerprint recognition to unblock the app.


src/ducks/pin/PinEditView.jsx
```diff
@@ -14,6 +16,35 @@ import { pinSetting } from 'ducks/pin/queries'
...
+const FingerprintChoice = translate()(({ t, onChoice }) => {
+  return (
+    <div className={styles.Pin__FingerprintChoice}>
+      <div className={styles.Pin__FingerprintChoice__top}>
+        <Icon size="4rem" icon={fingerprint} />
+        <br />
+        {t('Pin.use-fingerprint-question')}
+        <br />
+      </div>
+      <div style={{ flexGrow: 0 }}>
```
drazik: Is it possible to avoid inline styles ? I'm not totally against it, but I'd rather not mix classnames and inline styles

src/locales/en.json
```diff
@@ -547,6 +547,7 @@
       "toggle-title": "PIN code",
       "toggle-description": "Will lock your app when switching apps or after 1m of inactivity."
     },
-    "fingerprint-text": "Or your fingerprint"
+    "fingerprint-text": "Or your fingerprint",
+    "use-fingerprint-question": "Do you want to use your fingerprint to unlock Banks ?"
```
drazik: It seems to me that there is no space before punctuation in english. But I may be wrong

ptbrowne: Noice !

src/ducks/pin/PinEditView.jsx
ptbrowne: Done via utility class in https://github.com/cozy/cozy-banks/pull/1103/commits/41374e2a30ac7ccb74bcb54aa2f406f955102168 . Thanks !

PR #1120, opened 2019-04-01
refactor: Simplify webpack scripts and configuration

ptbrowne:
I tried to simplify our webpack scripts and configuration.

We use `--env` flags instead of environment variable.
This lets us get rid of our idiosyncratic `target:mode` 
NODE_ENV variable.

Hopefully it will also makes it easier to switch to
cozy-scripts.


webpack.config.js
```diff
@@ -2,14 +2,13 @@
...
 const {
-  production,
   target,
   hotReload,
   analyze
 } = require('./config/webpack.vars')
 
 module.exports = (env = {}) => {
-  env.target = env.target || target
+  env.target = env.target || 'browser'
```
drazik: Maybe we should have a default env and use it like this?

```js
const defaultEnv = {
  target: 'browser'
}

module.exports = (env = defaultEnv) => {
...
}
```

ptbrowne: Yes, though I would not use the default argument syntax since otherwise, inside the function you are changing the defaultEnv object and this is not something you want.

Done in https://github.com/cozy/cozy-banks/pull/1120/commits/4bef509bb4f421b9fd71be810255afcf19453c71

ptbrowne:
Merging since the failing test is not about that.

PR #1121, opened 2019-04-01
fix: Clear timeout in PinGuard

ptbrowne:
We have to clear timeouts when a component is unmounted.


ptbrowne:
Merging since the red build is not because of this.

PR #1125, opened 2019-04-02
Find apps suggestions in the service

drazik:
This add a new step to the onOperationOrBillCreate service. In this step, we look for available connectors that the user didn't installed but that can be useful for him, based on his banking transactions.

We create one `io.cozy.apps.suggestions` document for each connector we want to suggest. Each of these documents have relationships to the transactions that made us think this connector could be useful for the user.

You can find more on the [`io.cozy.apps.suggestions` documentation](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.apps.suggestions.md).


babel.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = function(api) {
   api.cache(true)
 
   return {
+    sourceType: 'unambiguous',
```
ptbrowne: Link to doc : https://babeljs.io/docs/en/options#sourcetype

> sourceType
> Type: "script" | "module" | "unambiguous"
> Default: "module"
 >   "script" - Parse the file using the ECMAScript Script grammar. No import/export statements allowed, and files are not in strict mode.
 >   "module" - Parse the file using the ECMAScript Module grammar. Files are automatically strict, and import/export statements are allowed.
    "unambiguous" - Consider the file a "module" if import/export statements are present, or else consider it a "script".


On avait quoi comme problème ?

src/ducks/appsSuggestions/AppsSuggestion.js
```diff
@@ -0,0 +1,45 @@
+import { Document } from 'cozy-doctypes'
+import { head } from 'lodash'
+import { TRANSACTION_DOCTYPE } from 'doctypes'
+
+class AppsSuggestion extends Document {
```
ptbrowne: I would prefer to name it AppSuggestion since we have only one app in each document right ?

src/ducks/appsSuggestions/AppsSuggestion.js
```diff
@@ -0,0 +1,45 @@
...
+  static fetchBySlug(slug) {
+    return this.queryAll({ slug }).then(head)
+  }
+
+  static linkTransaction(suggestion, transaction) {
```
ptbrowne: I don't know why but it seems weird to have `link` here where the original suggestion is not modified.

src/ducks/appsSuggestions/AppsSuggestion.js
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      ...suggestion,
+      relationships: {
+        ...suggestion.relationships,
+        transactions: {
+          data: [
+            ...suggestion.relationships.transactions.data,
```
ptbrowne: I think we should have a little helper that does the nested push.

`push('relationships.transactions.data', { _id: transaction._id, _type: TRANSACTION_DOCTYPE })`

Here I find it easy to erase data that we are not interested in. For example if `transactions` has other attributes than data, they will be erased.

```
const pushAtPath = (obj, path, value) => {
  const arr = get(obj, path) || []
  arr.push(value)
  set(obj, path, arr)
}
```

src/ducks/appsSuggestions/AppsSuggestion.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+  }
+}
+
+AppsSuggestion.doctype = 'io.cozy.apps.suggestions'
+AppsSuggestion.idAttributes = ['slug']
+AppsSuggestion.createdByApp = 'banks'
+
+export default AppsSuggestion
```
ptbrowne: AppSuggestion ?

src/ducks/appsSuggestions/Trigger.js
```diff
@@ -0,0 +1,7 @@
+import { Document } from 'cozy-doctypes'
+import { TRIGGER_DOCTYPE } from 'doctypes'
+
+class Trigger extends Document {}
+Trigger.doctype = TRIGGER_DOCTYPE
+
+export default Trigger
```
ptbrowne: Add it to cozy-doctypes ? Or open an issue there and a comment here.

src/ducks/appsSuggestions/services.js
```diff
@@ -0,0 +1,101 @@
...
+import { BankTransaction } from 'cozy-doctypes'
+import AppsSuggestion from './AppsSuggestion'
+import Trigger from './Trigger'
+import { groupBy, flatMap } from 'lodash'
+
+const log = logger.namespace('apps-suggestions')
+
+AppsSuggestion.registerClient(cozyClient)
+Trigger.registerClient(cozyClient)
+BankTransaction.registerClient(cozyClient)
```
ptbrowne:
```
import { Document } from 'cozy-doctypes'

Document.registerClient(client)
```

Document.registerClient does the job.

src/ducks/appsSuggestions/Trigger.js
ptbrowne: Opened an issue not to forget https://github.com/cozy/cozy-doctypes/issues/82

PR #1131, opened 2019-04-03
WIP: feat(store): Combine app store and cozy client store

drazik:
The goal is to be able to get documents from the redux store instead of cozy-client queries. This way the store acts like a local cache.

It's not fully working because the `cozy` property of the state contains empty `documents` and `queries` properties.


nono:
Ping @ptbrowne and @kosssi for review

ptbrowne:
Thanks @nono. I read that, I think we should add WIP to this PR.

nono:
Ping

PR #1138, opened 2019-04-05
refactor(KonnectorIcon): Style

kosssi:
Remove inline style

### DIFF
--- a/src/ducks/balance/components/KonnectorIcon.jsx
+++ b/src/ducks/balance/components/KonnectorIcon.jsx
@@ -2,12 +2,7 @@ import React from 'react'
 import UIAppIcon from 'cozy-ui/react/AppIcon'
 import { withClient } from 'cozy-client'
 
-const style = {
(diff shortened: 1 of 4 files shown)


src/ducks/balance/components/KonnectorIcon.jsx
```diff
@@ -23,10 +18,13 @@ class KonnectorIcon extends React.PureComponent {
...
   render() {
+    const { className } = this.props
+
     return (
-      <span style={style}>
-        <UIAppIcon fetchIcon={this.fetchIcon} />
-      </span>
+      <UIAppIcon
+        fetchIcon={this.fetchIcon}
+        className={className ? className : styles.KonnectorIcon_default}
```
drazik: We really want to totally override default style as soon as we pass a `className`?

ptbrowne: I don't think so. If urgent, put a comment and an issue :/

PR #1142, opened 2019-04-08
feat(amout): Blur all amount

kosssi:
On demo we need to hide the amounts

Capture d’écran 2019-04-08 à 11 30 38


kosssi:
I merged it on local

PR #1160, opened 2019-04-15
feat(AddAccountLink): Open a web browser on mobile

drazik:
The enrollment flow is better if we open the user's web browser instead
of an inapp browser: the user is already logged in in his browser.

### DIFF
--- a/src/ducks/settings/AddAccountLink.jsx
+++ b/src/ducks/settings/AddAccountLink.jsx
@@ -1,5 +1,3 @@
-/* global __TARGET__ */
-
 import React, { Component } from 'react'
 import { withClient } from 'cozy-client'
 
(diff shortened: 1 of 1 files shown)


src/ducks/settings/AddAccountLink.jsx
```diff
@@ -26,11 +24,7 @@ class SameWindowLink extends Component {
       }
     )
 
-    if (__TARGET__ === 'browser') {
-      window.location = redirectionURL
-    } else if (__TARGET__ === 'mobile') {
-      open(redirectionURL)
-    }
+    window.location = redirectionURL
```
ptbrowne: I think it would be worth commenting why we do not use open here. Apparently it is because in the onboarding flow, the user clicks on email links, which open the "native" browser (external Firefox for example) instead of the in-app one. This means login cookies are stored in the external browser. To prevent asking the user to login again, we have to use an external browser instead of the in app browser.

ptbrowne:
LGTM

PR #1167, opened 2019-04-18
refactor(KonnectorAction): Split match function

drazik:
So we can easily override it.


src/ducks/transactions/actions/KonnectorAction/match.js
```diff
@@ -0,0 +1,23 @@
+import { findMatchingBrand } from 'ducks/brandDictionary'
```
kosssi: this file is not imported. Are you sure you don't forget anything?

PR #1171, opened 2019-04-26
Updates on services webpack config

drazik:
This make it possible to use the customizations overrides in services too. Currently, we had to build the services without overrides, which is not good because we want some overrides to be used in services sometimes.


drazik:
Merging without review because it's urgent and nobody is available at the moment.

PR #1174, opened 2019-04-28
feat: Support universal link

Crash--:
-  [x] Tested on iOS (& entitlements created)
- [ ] Tested on Android => Works with http not https, but Android is not our priority since deeplinks are working for now. Will need to fix it later.

### DIFF
--- a/src/main.jsx
+++ b/src/main.jsx
@@ -79,8 +79,18 @@ const setupApp = async persistedState => {
         Alerter.info('Token refreshed')
       }
     })
-
+    const openUniversalLink = (/*eventData*/) => {
+      /* 
+      banks just need to be waked up by an universal link. 
+      We currently do not manage any path
+    */
+    }
     document.addEventListener('deviceready', onStartOrResume)
+    document.addEventListener('deviceready', () => {
+      if (window.universalLinks) {
+        window.universalLinks.subscribe('openUniversalLink', openUniversalLink)
+      }
+    })
     document.addEventListener('resume', onStartOrResume)
   }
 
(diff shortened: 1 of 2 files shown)


src/targets/mobile/config.xml
```diff
@@ -13,6 +13,12 @@
     <allow-intent href="geo:*" />
     <preference name="AppendUserAgent" value="io.cozy.banks.mobile-1.0.4" />
     <hook src="scripts/customizeConfigXML.js" type="before_prepare" />
+    <universal-links>
+        <host name="links.mycozy.cloud">
```
Crash--: This link should be overridden for customization

src/targets/mobile/config.xml
```diff
@@ -13,6 +13,12 @@
...
+            <path event="openUniversalLink" scheme="https" url="/banks/*" />
```
Crash--: This one not. The idea behind is to have : 
links.mycozy.cloud/{slug} 

The 'event' is the event to listen.

src/targets/mobile/config.xml
```diff
@@ -99,4 +105,5 @@
     </plugin>
     <plugin name="cordova-blur-app-privacy-screen" spec="https://github.com/lifeofcoding/cordova-blur-app-privacy-screen.git" />
     <engine name="ios" spec="4.5.5" />
+    <plugin name="cordova-universal-links-plugin" spec="https://github.com/walteram/cordova-universal-links-plugin.git" />
```
Crash--: This fork fix the installation on android

src/main.jsx
```diff
@@ -79,8 +79,18 @@ const setupApp = async persistedState => {
         Alerter.info('Token refreshed')
       }
     })
-
+    const openUniversalLink = (/*eventData*/) => {
+      /* 
+      banks just need to be waked up by an universal link. 
+      We currently do not manage any path
+    */
+    }
```
Crash--: The only place where banks need to resolve UL is during the onboarding process. It'll be done in cozy-auth

ptbrowne: So this is not necessary ?

Crash--: "ok laisse le" :D

PR #1177, opened 2019-05-02
Integrate new authentication

ptbrowne:
The updated cozy-authentication library takes care of most of the work done initially in the MobileRouter. This and the fact that cozy-client has learned to emit events let us organize the code in a more logical and decoupled way.

For example, it is now the responsibility of cozy-authentication to save the credentials and reconnect the user if the app is reopened after a first login. This enables us to completely remove the mobile store that was used to store user credentials.

Thanks to events, it is possible to decouple the code used by the bar to init from the initialisation of the client. Same for the push notification.


src/ducks/mobile/bar.js
```diff
@@ -0,0 +1,45 @@
...
+    if (document.getElementById('coz-bar')) {
+      document.getElementById('coz-bar').remove()
+      document.body.setAttribute('style', '')
+    }
+  })
+
+  client.on('refreshToken', () => {
+    cozy.bar.updateAccessToken(client.stackClient.token.accessToken)
+  })
+}
```
drazik: :+1:

src/ducks/bar/BarTheme.jsx
```diff
@@ -1,7 +1,21 @@
+/* global cozy, __TARGET__ */
+
 import React from 'react'
 import PropTypes from 'prop-types'
 import withSideEffect from 'react-side-effect'
-import { setBarTheme } from 'ducks/mobile/utils'
+import { setTheme as setStatusBarTheme } from './statusBar'
```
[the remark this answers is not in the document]
ptbrowne: Done thanks !

src/ducks/mobile/bar.js
ptbrowne: Not needed anymore since the bar is initaliazed with a CozyClient, we can completely remove this !

ptbrowne: Done in https://github.com/cozy/cozy-banks/pull/1177/commits/88b5895279ac215d97eb1dccd32e41d5dac7ecc5

package.json
```diff
@@ -174,6 +174,7 @@
     "geco": "0.11.1",
     "hammerjs": "2.0.8",
     "html-webpack-harddisk-plugin": "1.0.1",
+    "ios-sim": "8.0.1",
```
kosssi: why you import this dep?

src/ducks/pin/PinAuth.jsx
```diff
@@ -119,9 +118,8 @@ class PinAuth extends React.Component {
   }
 
   logout() {
-    const { router } = this.props
-    const { client, store } = this.context
-    onLogout(store, client, router)
+    const client = this.props.client
```
kosssi:
```suggestion
    const { client } = this.props
```

src/ducks/client/mobile.js
```diff
@@ -48,6 +54,7 @@ export const getManifestOptions = manifest => {
   }
 }
 
+// Should be moved to cozy-pouch
```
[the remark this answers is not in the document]
ptbrowne: OK for the discussion, I will update the comment to // Should be moved to a shared repo

src/ducks/pin/PinAuth.jsx
ptbrowne: Done

package.json
ptbrowne: yarn ios:run a planté si je faisais pas ca.

src/targets/mobile/fastlane/Fastfile
```diff
@@ -4,7 +4,8 @@ platform :ios do
   lane :devices do
     register_devices(
        devices: {
-         "Patrick iPhone 6s" => "dbb0aa403f6928368467d3f10b28a61336e52023"
+         "Patrick iPhone 6s" => "dbb0aa403f6928368467d3f10b28a61336e52023",
+         "Quentin iPhone X" => "f104c368769384e8f043ba497a74f704d27d6d3f"
```
drazik: What is it useful for? (just asking for my culture :))

ptbrowne: To reigster devices on itunes connect.

PR #1179, opened 2019-05-02
feat(account): Show the good account update date

drazik:
Use `BankAccount.getUpdatedAt` from cozy-doctypes to get the best update
date possible. See cozy/cozy-libs#423 for more
infos.

### DIFF
--- a/src/ducks/account/helpers.js
+++ b/src/ducks/account/helpers.js
@@ -3,6 +3,7 @@ import { getDate, getReimbursedAmount } from 'ducks/transactions/helpers'
 import { isHealthExpense } from 'ducks/categories/helpers'
 import { differenceInCalendarDays, isThisYear } from 'date-fns'
 import flag from 'cozy-flags'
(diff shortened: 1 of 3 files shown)


src/ducks/account/helpers.js
```diff
@@ -21,8 +21,21 @@ export const getAccountInstitutionLabel = account => {
...
+  }
+
+  const cozyUpdatedAt = get(account, 'cozyMetadata.updatedAt')
+
+  if (cozyUpdatedAt) {
+    return cozyUpdatedAt
+  }
+
+  return null
+}
```
ptbrowne: Could this be moved to cozy-libs/cozy-doctypes ? It's good to start using this even in the views I think. WDYT ?

drazik: Yep nice idea. See https://github.com/cozy/cozy-libs/pull/423

ptbrowne:
Top ! LGTM ! Thanks !

PR #1184, opened 2019-05-06
fix: Send balance-lower emails + overrides

ptbrowne:
Missing helper caused emails not to be sent.

Added tests. Surprisingly, mjml failed mysteriously on CSS classes with a double-dash. It caused a ParserError so I removed the double dashes and used camelCases.

I also used src-relative paths to be able to override. 

EDIT: Also added a way to override top logo


src/ducks/notifications/html/emails.spec.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable no-console */
```
drazik: It seems you can remove it since you don't use `console`

ptbrowne:
Thx drazik, I made a typo... it should be fine now... I added a commit to mock the date since the current date is used in the emails, and we do not want the snapshots to be dependent on the current date.

PR #1189, opened 2019-05-10
New realtime

ptbrowne:
- Updates the cozy-banks code to use the new realtime API.
- Re-fetch when on resume/online only if accounts.length === 0


src/ducks/balance/Balance.jsx
```diff
@@ -175,8 +178,13 @@ class Balance extends PureComponent {
...
   }
 
-  fetchTriggers() {
-    const client = this.props.client
+  async fetchTriggers() {
+    const { client } = this.props
+    if (__TARGET__ === 'mobile') {
+      const pouchManager = client.links[0].pouches
+      pouchManager.stopReplicationLoop()
+      await pouchManager.startReplicationLoop()
```
drazik: Is this the final solution or just a temporary one? It doesn't seem good to rely on pouch link internals. Shouldn't the link expose a method to do this?

src/ducks/balance/Balance.jsx
```diff
@@ -188,8 +196,13 @@ class Balance extends PureComponent {
...
   }
 
-  fetchAccounts() {
-    const client = this.props.client
+  async fetchAccounts() {
+    const { client } = this.props
+    if (__TARGET__ === 'mobile') {
+      const pouchManager = client.links[0].pouches
+      pouchManager.stopReplicationLoop()
+      await pouchManager.startReplicationLoop()
```
drazik: We repeat ourselves, we should extract these lines in a method.

src/targets/mobile/config.xml
```diff
@@ -105,6 +105,6 @@
     </plugin>
     <plugin name="cordova-blur-app-privacy-screen" spec="https://github.com/lifeofcoding/cordova-blur-app-privacy-screen.git" />
     <engine name="ios" spec="4.5.5" />
-    <plugin name="cordova-plugin-keyboard" spec="https://github.com/cozy/cordova-plugin-keyboard.git#4fa4850e6f0ea78e1e2ce573ccde5531a2353edb" />
     <plugin name="cordova-universal-links-plugin" spec="https://github.com/walteram/cordova-universal-links-plugin.git" />
+    <plugin name="cordova-plugin-network-information" spec="^2.0.1" />
```
drazik: Do we really use this plugin? I don't see any reference to it in the changes.

ptbrowne: C'est pour les événements online offline.

src/ducks/balance/Balance.jsx
ptbrowne: Done in https://github.com/cozy/cozy-banks/pull/1189/commits/76248a3889af03b5d73e95a40b91ebb5621aab41

src/targets/mobile/config.xml
drazik: Okay :+1:

src/ducks/balance/Balance.jsx
ptbrowne: Done in https://github.com/cozy/cozy-client/pull/434

PR #1193, opened 2019-05-13
feat: Simplify realtime logic in Balance

ptbrowne:
Tried to simplify realtime state
Transactions were not fetched again which left a timeline
that was incorrect.


drazik:
Just a lint issue : https://travis-ci.org/cozy/cozy-banks/jobs/531754249#L478

PR #1196, opened 2019-05-14
feat: Realtime fallback

ptbrowne:
It seems that on some Android device, the realtime does not work correctly. As a fallback, a simple setInterval loop is used to fetch data periodically.

### DIFF
--- a/src/ducks/balance/Balance.spec.jsx
+++ b/src/ducks/balance/Balance.spec.jsx
@@ -4,6 +4,7 @@ import getClient from 'test/client'
 const React = require('react')
 const { DumbBalance } = require('./Balance')
 const debounce = require('lodash/debounce')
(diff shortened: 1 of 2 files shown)


ptbrowne:
> But is there any plan to put it in cozy-realtime?

Not really right now. I keep this in mind though.

src/ducks/balance/Balance.spec.jsx
```diff
@@ -16,7 +17,7 @@ const fakeCollection = (doctype, data) => ({
 describe('Balance page', () => {
   let root, instance, saveDocumentMock, router, filterByAccounts
 
-  beforeEach(() => {
+  const setup = ({ accountsData } = {}) => {
```
kosssi: 👍 bonne idée

PR #1199, opened 2019-05-15
UI for transactions waiting for their categorization

drazik:
We now have a special status for the transactions: « categorization in progress ». This is shown as a « special » category in the UI, with an animated icon.

image
image

The user can still manually categorize these transactions, but this special category is obviously not selectable in the categorization choices.

I also fixed a long lasting bug: the « other » subcategory was always selected alongside the real selected subcategory. This is no longer the case.

### DIFF
--- a/src/components/PopupSelect/index.jsx
+++ b/src/components/PopupSelect/index.jsx
@@ -70,6 +70,8 @@ class PopupSelect extends Component {
     const { history } = this.state
     const current = history[0]
     const children = current.children || []
+    const selectedItem = children.find(this.props.isSelected)
+
     return (
       <Modal
         closeBtnClassName={this.props.closeBtnClassName}
@@ -91,7 +93,7 @@ class PopupSelect extends Component {
             {children.map(item => (
               <PopupRow
                 key={item.title}
-                isSelected={this.props.isSelected(item)}
+                isSelected={item === selectedItem}
                 icon={item.icon}
                 label={item.title}
                 onClick={() => this.handleSelect(item)}
--- a/src/ducks/categories/PendingCategoryIcon.jsx
+++ b/src/ducks/categories/PendingCategoryIcon.jsx
@@ -8,6 +8,8 @@ const PendingCategoryIcon = ({ size = 32 }) => {
     height: size
   }
 
+  // We don't externalize the SVG into its own file and use it in a cozy-ui
+  // Icon because it breaks the SVG animations
   return (
     <span className={styles.PendingCategoryIcon} style={style}>
       <svg
--- a/src/ducks/categories/PendingCategoryIcon.jsx
+++ b/src/ducks/categories/PendingCategoryIcon.jsx
@@ -1,0 +1,59 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from 'ducks/categories/PendingCategoryIcon.styl'
+
+const PendingCategoryIcon = ({ size = 32 }) => {
+  const style = {
+    width: size,
+    height: size
+  }
+
+  return (
+    <span className={styles.PendingCategoryIcon} style={style}>
+      <svg
+        version="1.1"
+        xmlns="http://www.w3.org/2000/svg"
+        xmlnsXlink="http://www.w3.org/1999/xlink"
+        x="0px"
+        y="0px"
+        viewBox="0 0 100 100"
+        enableBackground="new 0 0 0 0"
+        xmlSpace="preserve"
+      >
+        <circle fill="#fff" stroke="none" cx="25" cy="50" r="6">
+          <animate
+            attributeName="opacity"
+            dur="1s"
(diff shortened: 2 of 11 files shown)


src/ducks/categories/PendingCategoryIcon.jsx
```diff
@@ -0,0 +1,59 @@
...
+        <circle fill="#fff" stroke="none" cx="75" cy="50" r="6">
+          <animate
+            attributeName="opacity"
+            dur="1s"
+            values="0;1;0"
+            repeatCount="indefinite"
+            begin="0.3"
+          />
+        </circle>
+      </svg>
```
ptbrowne: Any particular reason to directly inline SVG ?

Disadvantage I see : Cannot be edited by tools that cna edit SVG, cannot be optimized ?

Not blocking.

src/ducks/categories/PendingCategoryIcon.jsx
```diff
@@ -0,0 +1,59 @@
...
+    </span>
+  )
+}
+
+PendingCategoryIcon.propTypes = {
+  size: PropTypes.number
+}
+
```
ptbrowne: Why not use an Icon ? It is the job of Icon to display an SVG centered horizontally and vertically.

src/ducks/categories/PendingCategoryIcon.jsx
drazik: When I put it in an external file and used it with cozy-ui's `Icon`, SVG animations did'nt work.

src/ducks/categories/PendingCategoryIcon.jsx
```diff
@@ -0,0 +1,59 @@
...
+          />
+        </circle>
+        <circle fill="#fff" stroke="none" cx="75" cy="50" r="6">
+          <animate
+            attributeName="opacity"
+            dur="1s"
+            values="0;1;0"
+            repeatCount="indefinite"
+            begin="0.3"
+          />
```
ptbrowne: Nice, I did not know it was so easy to animate an SVG.

src/ducks/categories/helpers.js
```diff
@@ -51,6 +51,10 @@ export const getCategoryId = transaction => {
     return transaction.cozyCategoryId
   }
 
+  if (!transaction.localCategoryId && !transaction.cozyCategoryId) {
+    return null
+  }
+
   return transaction.automaticCategoryId
```
ptbrowne: Maybe we can add a comment, "returned only if the automatic categorization by Cozy is not deemed not usable as per GLOBAL_MODEL_USAGE_THRESOLD" ?

src/ducks/categories/categoriesMap.js
```diff
@@ -120,4 +120,6 @@ Object.keys(tree).forEach(catId => {
   }
 })
 
+export const isAwaitingCategory = categoryId => categoryId === '-1'
```
ptbrowne: Shouldn't it be `null` here ?

ptbrowne: This helper is not used in this commit.

src/components/PopupSelect/index.jsx
```diff
@@ -70,6 +70,8 @@ class PopupSelect extends Component {
     const { history } = this.state
     const current = history[0]
     const children = current.children || []
+    const selectedItem = children.find(this.props.isSelected)
```
ptbrowne: So here, we are "lucky" because the "others" category comes in last ? What makes this.props.isSelected be true for others ? Hasn't it be fixed by the previous commit ?

src/ducks/categories/helpers.js
```diff
@@ -75,6 +75,11 @@ export const transactionsByCategory = transactions => {
     // Creates a map of categories, where each entry contains a list of
     // related operations and a breakdown by sub-category
     const catId = getCategoryId(transaction)
+
+    if (!catId) {
```
ptbrowne: I would be prefer to have an explicity `if (isAwaitingCategory(catId))` here, it would be more readable IMHO.

src/ducks/categories/PendingCategoryIcon.jsx
ptbrowne: Can you add a comment saying this ? :)

src/components/PopupSelect/index.jsx
drazik: Yes we are more or less lucky. The problem is that a category and its corresponding « others » subcategory have the same id. Since `this.props.isSelected` returns `true` for the parent category too, then it returns `true` for the « others » subcategory.

src/ducks/categories/PendingCategoryIcon.jsx
ptbrowne: Because Icon does not work with animated SVG.

PR #1203, opened 2019-05-17
Refactor apps suggestions

drazik:
The apps suggestions were crashing when there was a certain amount of transactions to analyze. There was two problems:

* A request to the DB was made for each transaction to find a potential corresponding app suggestion : now we fetch all existing suggestions in a single request
* The number of transactions listed in the relationships of a `io.cozy.apps.suggestions` document was not constrained, so it could grow very quickly. Writing big documents was  causing errors. We now save only the last 10 linked transactions

### DIFF
--- a/src/ducks/appSuggestions/AppSuggestion.js
+++ b/src/ducks/appSuggestions/AppSuggestion.js
@@ -1,21 +1,6 @@
 import { Document } from 'cozy-doctypes'
 import { head } from 'lodash'
 import { TRANSACTION_DOCTYPE } from 'doctypes'
-import get from 'lodash/get'
-import setWith from 'lodash/setWith'
-import clone from 'lodash/clone'
-
-/* Set a value inside an object, by path, immutably */
-const setIn = (obj, path, value) => {
-  return setWith(clone(obj), path, value, clone)
-}
-
-/* Push value into nested array, immutably */
-const pushAtPath = (obj, path, value) => {
-  const arr = get(obj, path, [])
-  arr.push(value)
-  return setIn(obj, path, arr)
-}
 
 class AppSuggestion extends Document {
   static fetchBySlug(slug) {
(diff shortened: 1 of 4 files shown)


src/ducks/appSuggestions/AppSuggestion.js
```diff
@@ -27,11 +12,13 @@ class AppSuggestion extends Document {
...
-      suggestion,
-      'relationships.transactions.data',
-      transactionRelationship
-    )
+
+    // We slice 9 existing transactions because we want to store only 10 items
+    // in the transactions relationship
+    suggestion.relationships.transactions.data = [
+      transactionRelationship,
+      ...suggestion.relationships.transactions.data.slice(0, 9)
```
ptbrowne: Here the latest suggestion is first. Is it wanted ?

src/ducks/appSuggestions/services.spec.js
```diff
@@ -8,28 +7,37 @@ describe('findSuggestionForTransaction', () => {
...
       }
-    })
+    ]
 
-    const suggestion = await findSuggestionForTransaction(transaction, brands)
+    const suggestion = await findSuggestionForTransaction(
+      transaction,
+      brands,
+      suggestions
+    )
```
ptbrowne: I could not see a test that was testing that we are using the originalSuggestion if it exists ? I think we need one if it does not exist.

src/ducks/appSuggestions/AppSuggestion.js
drazik: Yes it is wanted

src/ducks/appSuggestions/services.spec.js
drazik: That's what this one is testing: we have an existing suggestion, and `findSuggestionForTransaction` uses it and updates it to add the transaction to the suggestion's relationships.

PR #1212, opened 2019-05-24
Automatic reimbursement status on health expenses

drazik:
For health expenses, we want the reimbursement status to be controlled by the reimbursements we matched for a given transaction.

Here is the rule :

* If the user manually placed a reimbursement status, just show it
* If the user didn't manually placed a reimbursement status, check the reimbursements
  * If the transaction is fully reimbursed, then show "reimbursed"
  * If the transaction is not fully reimbursed, then show "pending"

This way, if we matched some reimbursements, we can show a status reflecting this, but the user can still manually override the status we infered.

By doing this on the client side, we avoid too many refactors and data duplication between the `reimbursementStatus` and the `reimbursements` array.

### DIFF
--- a/src/ducks/transactions/helpers.js
+++ b/src/ducks/transactions/helpers.js
@@ -123,8 +123,21 @@ export const isNew = transaction => {
   return parseInt(transaction._rev.split('-').shift(), 10) <= 2
 }
 
-export const getReimbursementStatus = transaction =>
-  transaction.reimbursementStatus || 'no-reimbursement'
+export const getReimbursementStatus = transaction => {
+  if (isHealthExpense(transaction)) {
+    return getHealthExpenseReimbursementStatus(transaction)
+  }
+
+  return transaction.reimbursementStatus || 'no-reimbursement'
+}
+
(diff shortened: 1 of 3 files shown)


src/ducks/transactions/helpers.js
```diff
@@ -123,8 +123,21 @@ export const isNew = transaction => {
...
+  return transaction.reimbursementStatus || 'no-reimbursement'
+}
+
+export const getHealthExpenseReimbursementStatus = transaction => {
+  if (transaction.reimbursementStatus) {
+    return transaction.reimbursementStatus
+  }
+
+  return isFullyReimbursed(transaction) ? 'reimbursed' : 'pending'
+}
```
ptbrowne: At some point, having this helpers in a reimbursement.js file would be good, it could then be shared in doctypes.

drazik: I don't understand what you mean by « it could then be shared in doctypes ». To me, this helper is very coupled to the `io.cozy.bank.operations` doctype.

PR #1219, opened 2019-05-28
Use Chip for all CTAs on transaction row

drazik:
Render a `Chip` instead of a `ButtonAction` for every CTA in the transaction row. This is behind the `reimbursement-tag` flag.

image
image
image
image

### DIFF
--- a/src/ducks/transactions/TransactionActions.jsx
+++ b/src/ducks/transactions/TransactionActions.jsx
@@ -58,7 +58,14 @@ export const SyncTransactionActions = ({
   compact,
   className
 }) => (
-  <span className={cx('u-dib', className)}>
+  <span
+    className={cx(
+      {
+        'u-dib': !isModalItem
+      },
+      className
+    )}
+  >
     {(displayDefaultAction || onlyDefault) && actions.default && (
       <MenuAction
         action={actions.default}
--- a/src/ducks/transactions/TransactionRow.jsx
+++ b/src/ducks/transactions/TransactionRow.jsx
@@ -204,17 +204,15 @@ class _RowMobile extends React.PureComponent {
           )}
         </Media>
         {flag('reimbursement-tag') && (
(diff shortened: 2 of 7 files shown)


src/ducks/transactions/TransactionActions.jsx
```diff
@@ -58,7 +58,14 @@ export const SyncTransactionActions = ({
   compact,
   className
 }) => (
-  <span className={cx('u-dib', className)}>
+  <span
+    className={cx(
+      {
+        'u-dib': !isModalItem
```
ptbrowne: Nit: When there is only one class i want to add conditionally, I prefer `cx(!isModalItem ? 'u-dib' : null, className)`

src/ducks/transactions/TransactionRow.jsx
```diff
@@ -204,17 +204,15 @@ class _RowMobile extends React.PureComponent {
...
-          </div>
+          <TransactionActions
+            transaction={transaction}
+            urls={urls}
+            brands={brands}
+            onlyDefault
+            compact
+            menuPosition="right"
+            className="u-mt-half u-ml-2-half"
+          />
```
ptbrowne: :tada:

src/ducks/transactions/actions/ReimbursementStatusAction/index.js
```diff
@@ -84,23 +69,37 @@ class ReimbursementStatusAction extends React.PureComponent {
...
+    const { isModalItem, transaction } = this.props
+    const status = getReimbursementStatus(transaction)
 
-    if (isModalItem) {
-      return this.renderModalItem()
-    } else {
-      return this.renderTransactionRow()
-    }
+    return (
+      <>
```
kosssi: C'est pas bizarre de faire une balise vide ?

PR #1221, opened 2019-05-29
Fixes on new actions

drazik:
Some things that I missed while implementing new actions styles:

* Use `file-outline` icon
* Use `normal` Chip theme when a reimbursement is in pending state
* Properly align transaction row content on mobile

### DIFF
--- a/src/ducks/transactions/Transactions.styl
+++ b/src/ducks/transactions/Transactions.styl
@@ -83,3 +83,13 @@ for category in categories
 .TransactionRowMobile
     flex-direction column
     align-items flex-start
+    justify-content center
+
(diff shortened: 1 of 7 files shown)


src/ducks/transactions/Transactions.styl
```diff
@@ -83,3 +83,13 @@ for category in categories
...
+    justify-content center
+
+.TransactionRowMobile__actions
+    display inline-block
+
+    // We may have an action that matches, but render nothing. In that case,
+    // this element exists, but is empty. We want to hide it to avoid extra
+    // margins
+    &:empty
+        display none
```
ptbrowne: TIL :empty .

PR #1223, opened 2019-05-31
fix: Use a more recent universal link plugin

Crash--:
This fork is up to date, and support the new iOS system for `entitlements`. 

Before, it was not possible to have universal links support by building the app from the cli. It worked from XCode, though. 

Now, with this forks, we can build from testflight... ;)

### DIFF
--- a/src/targets/mobile/config.xml
+++ b/src/targets/mobile/config.xml
@@ -105,6 +105,6 @@
     </plugin>
     <plugin name="cordova-blur-app-privacy-screen" spec="https://github.com/lifeofcoding/cordova-blur-app-privacy-screen.git" />
     <engine name="ios" spec="4.5.5" />
-    <plugin name="cordova-universal-links-plugin" spec="https://github.com/walteram/cordova-universal-links-plugin.git" />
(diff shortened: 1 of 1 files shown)


src/targets/mobile/config.xml
```diff
@@ -105,6 +105,6 @@
     </plugin>
     <plugin name="cordova-blur-app-privacy-screen" spec="https://github.com/lifeofcoding/cordova-blur-app-privacy-screen.git" />
     <engine name="ios" spec="4.5.5" />
-    <plugin name="cordova-universal-links-plugin" spec="https://github.com/walteram/cordova-universal-links-plugin.git" />
+    <plugin name="cordova-universal-links-plugin" spec="https://github.com/Mailbutler/cordova-universal-links-plugin"/>
```
ptbrowne: There is no .git at the end, is it ok ?

Crash--: Yep. It works better without the `.git`. In fact, if you delete the package.json and json-lock and try to install plugin with a `.git` it'll not works. Even weirder, If you copy/paste the line (with the `.git`) and try to install it from the cli, it'll works... Va comprendre

PR #1224, opened 2019-06-03
fix(transactions): Late reimbursement use real months difference

drazik:
Using calendar months difference was not the good solution because the
calendar months difference between `2019-05-31` and `2019-06-01` is `1`.
We want to have the difference in full months.

### DIFF
--- a/src/ducks/transactions/helpers.spec.js
+++ b/src/ducks/transactions/helpers.spec.js
@@ -1,5 +1,4 @@
 import configureStore from 'store/configureStore'
-import differenceInMonths from 'date-fns/difference_in_calendar_months'
 import {
   hydrateTransaction,
   getDate,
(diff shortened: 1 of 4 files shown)


src/ducks/transactions/helpers.spec.js
```diff
@@ -1,5 +1,5 @@
 import configureStore from 'store/configureStore'
-import differenceInCalendarMonths from 'date-fns/difference_in_calendar_months'
+import differenceInMonths from 'date-fns/difference_in_calendar_months'
```
ptbrowne: Il faut modifier la partie droite de l'import aussi. Et comprendre pourquoi ca fonctionnait alors que la partie droite etait toujours `difference_in_calendar_months`.

PR #1226, opened 2019-06-03
Handle reimbursement tag on balance and health reimbursements pages

drazik:
The reimbursement tag was not taken into account on these pages. This is now the case. Still behind the `reimbursement-tag` flag.

The health reimbursement virtual account's balance is computed by taking only pending reimbursements. And the reimbursements page split the transaction according to their reimbursement tag.

### DIFF
--- a/src/ducks/account/helpers.js
+++ b/src/ducks/account/helpers.js
@@ -1,5 +1,9 @@
-import { get, sumBy } from 'lodash'
-import { getDate, getReimbursedAmount } from 'ducks/transactions/helpers'
+import { get, sumBy, overEvery, flowRight as compose } from 'lodash'
+import {
+  getDate,
+  getReimbursedAmount,
+  hasPendingReimbursement
+} from 'ducks/transactions/helpers'
 import { isHealthExpense } from 'ducks/categories/helpers'
 import { differenceInCalendarDays, isThisYear } from 'date-fns'
 import flag from 'cozy-flags'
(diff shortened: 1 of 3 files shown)


src/ducks/account/helpers.js
```diff
@@ -123,7 +127,15 @@ export const getAccountUpdatedAt = account => {
...
-    return isHealthExpense(transaction) && isThisYear(getDate(transaction))
+    if (flag('reimbursement-tag')) {
+      return (
+        isHealthExpense(transaction) &&
+        isThisYear(getDate(transaction)) &&
+        getReimbursementStatus(transaction) === 'pending'
+      )
+    } else {
+      return isHealthExpense(transaction) && isThisYear(getDate(transaction))
+    }
```
ptbrowne: Nit: I think it would be clearer with a more functional/compositional logic. To show that the flag is only adding one more function and not a different logic.

```js
const filter = _.overEvery([
  isHealthExpense,
  compose(isThisYear, getDate),
  flag('reimbursement-tag') ? isPendingReimbursement : null
].filter(Boolean))
const healthExpenses = transactions.filter(filter)
```

src/ducks/reimbursements/HealthReimbursements.jsx
```diff
@@ -3,24 +3,42 @@ import { connect } from 'react-redux'
...
+    const groupedTransactions = groupBy(
+      filteredTransactions,
+      reimbursementTagFlag ? getReimbursementStatus : isFullyReimbursed
+    )
+
+    const reimbursedTransactions = reimbursementTagFlag
+      ? groupedTransactions.reimbursed
+      : groupedTransactions.true
+
+    const awaitingTransactions = reimbursementTagFlag
```
ptbrowne: Nit: pendingTransactions to be coherent with the group name.

src/ducks/reimbursements/HealthReimbursements.jsx
```diff
@@ -3,24 +3,42 @@ import { connect } from 'react-redux'
...
+      ? groupedTransactions.pending
+      : groupedTransactions.false
```
ptbrowne: I am not a fan of doing groupBys in the render. Would it be easy to hide that behind a selector (easily memoizable if we need it).

src/ducks/account/helpers.js
drazik: TIL `overEvery`, thanks

src/ducks/reimbursements/HealthReimbursements.jsx
drazik: Unfortunately I can't do this now, so I added a comment above the grouping code to say that it should be extracted to a selector. Sorry

PR #1227, opened 2019-06-03
feat: Use AppLinker helper to generate universal link

ptbrowne:
Links in emails bring to native if possible now.


package.json
```diff
@@ -140,6 +140,7 @@
     "stylus": "0.54.5",
     "stylus-loader": "3.0.2",
     "svg-sprite-loader": "3.9.2",
+    "universal-url": "^2.0.0",
```
kosssi: can you fix version

PR #1228, opened 2019-06-03
Update reimbursement status modal UI

drazik:
Add a header and missing border on the choices list.

image
image


src/components/List/index.jsx
```diff
@@ -3,7 +3,21 @@ import styles from 'components/List/List.styl'
...
 
-export const List = props => <div>{props.children}</div>
+export const List = props => {
+  const { className, withOuterBorders, ...rest } = props
+
+  return (
+    <div
+      className={cx(
+        {
+          [styles['List--withOuterBorders']]: withOuterBorders
```
ptbrowne: `bordered` could be simpler than `withOuterBorders` ?

src/ducks/transactions/actions/ReimbursementStatusAction/ReimbursementStatusModal.jsx
```diff
@@ -1,34 +1,50 @@
...
+    const { transaction, onChange, t, className, ...rest } = this.props
     const choices = ['pending', 'reimbursed', 'no-reimbursement']
+    const status = getReimbursementStatus(transaction)
 
     return (
-      <PageModal {...rest}>
-        <form>
-          <List>
+      <Modal mobileFullscreen className={cx('u-pt-2', className)} {...rest}>
+        <header className="u-ta-center">
```
ptbrowne: Are you sure you do not want a `ModalHeader` here ?

drazik: I could use a `ModalHeader`, but I would have to do custom styles around it because we are not 100% in the use case of this component.

PR #1229, opened 2019-06-03
fix(Analyse): Display all account when virtual account is selected

kosssi:


src/ducks/filters/index.js
```diff
@@ -22,7 +22,7 @@ export const getFilteringDoc = state =>
 export const getFilteredAccountIds = state => {
   const availableAccountIds = getAccounts(state).map(x => x._id)
   const doc = getFilteringDoc(state)
-  if (!doc) {
+  if (!doc || !doc._type) {
```
drazik: Nit: I would add a warning in the console in the `!doc._type` case, because this is not a normal situation. Here, this is because the account it a virtual account, but I'm pretty sure someday we will want to fix this.

PR #1232, opened 2019-06-03
fix(deps): Update cozy-bar to v6.28.5

drazik:
The main goal is to have a cozy-bar with a fresh cozy-ui (20.7.0). This
indirectly fixes an issue when building for mobile: cozy-bar's cozy-ui
stylesheet is overriding the app's cozy-ui stylesheet. Which means that
if the cozy-bar has an old cozy-ui, we can have weird bugs on the mobile
app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,7 @@
     "classnames": "2.2.6",
     "cordova": "7.1.0",
     "cozy-authentication": "1.13.0",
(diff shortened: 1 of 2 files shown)


drazik:
> Are you sure about all theses changes in the yarn.lock file? I know we added some dep to the bar recently but :o

Yes, all these changes happen when I update the cozy-bar :/

PR #1234, opened 2019-06-04
feat: Don't display reimbursements accounts on categories

kosssi:
When categories page is display, it reset filter if filtering doc is reimbursements accounts

### DIFF
--- a/src/ducks/account/helpers.js
+++ b/src/ducks/account/helpers.js
@@ -8,6 +8,7 @@ import { isHealthExpense } from 'ducks/categories/helpers'
 import { differenceInCalendarDays, isThisYear } from 'date-fns'
 import flag from 'cozy-flags'
 import { BankAccount } from 'cozy-doctypes'
+import { ACCOUNT_DOCTYPE } from 'doctypes'
 
 const PARTS_TO_DELETE = ['(sans Secure Key)']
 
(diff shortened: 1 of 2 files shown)


src/ducks/account/helpers.js
```diff
@@ -146,6 +146,7 @@ export const buildHealthReimbursementsVirtualAccount = transactions => {
 
   const account = {
     _id: 'health_reimbursements',
+    _type: 'io.cozy.bank.accounts',
```
drazik: You can do

```js
import ACCOUNT_DOCTYPE from 'doctypes'
```

then

```js
const account = {
  // ...
  _type: ACCOUNT_DOCTYPE
}
```

src/ducks/categories/CategoriesPage.jsx
```diff
@@ -25,6 +29,17 @@ import flag from 'cozy-flags'
...
+    super(props)
+    const { filteringDoc, resetFilterByDoc } = this.props
+    if (
+      filteringDoc &&
+      includes(['Reimbursements', 'health_reimbursements'], filteringDoc._id)
+    ) {
+      resetFilterByDoc()
+    }
+  }
+
```
drazik: @ptbrowne had a good example of a component that can be constructed but never mounted (which I can't remember, sorry). With this in mind, I think this logic should be put in a `componentDidMount` method.

kosssi:
```jsx
const a = <A />
if (b) { reutrn null}
return a
```

le composant A est instancié mais pas monté

PR #1240, opened 2019-06-04
fix: Remove warnings

ptbrowne:
Remove React warnings

### DIFF
--- a/src/ducks/categories/Categories.jsx
+++ b/src/ducks/categories/Categories.jsx
@@ -276,8 +276,8 @@ class Categories extends Component {
                 />
               </Img>
             </Media>
+            <PercentageLine value={percentage} color={category.color} />
           </td>
-          <PercentageLine value={percentage} color={category.color} />
         </tr>
       ),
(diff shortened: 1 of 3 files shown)


src/components/KonnectorUpdateInfo/index.js
```diff
@@ -69,9 +69,8 @@ class KonnectorUpdateInfo extends React.PureComponent {
           <SubTitle className="u-monza">
             {t('KonnectorUpdateInfo.title')}
           </SubTitle>
-          <Text
-            tag="p"
-            className="u-mt-half"
+          <p
+            className="u-mt-half u-text u-ellipsis"
```
drazik: Shouldn't we open an issue on cozy-ui for that? I think that `Text` should not set `children` prop as required. Since there is no potential error in the component if the prop is undefined, then the prop is not required for the component to work.

ptbrowne: Removed this commit since https://github.com/cozy/cozy-ui/pull/1030 has been merged. Thx.

PR #1246, opened 2019-06-05
Empty health reimbursements page

drazik:
* Show some messages when there is no pending reimbursements and/or reimbursed transactions
* Show a link to the store filtered on insurance konnectors when there is no reimbursed transactions
* Dinstinguish the sections visually
* Some minor style updates

Live on https://testbanksemptyreimbursementspage-banks.cozy.works

### DIFF
--- a/src/components/Section/Section.jsx
+++ b/src/components/Section/Section.jsx
@@ -1,0 +1,11 @@
+import React from 'react'
+import cx from 'classnames'
+import styles from 'components/Section/Section.styl'
+
+const Section = props => {
+  const { className, ...rest } = props
+
+  return <section className={cx(styles.Section, className)} {...rest} />
+}
+
+export default Section
--- a/src/components/StoreLink/StoreLink.jsx
+++ b/src/components/StoreLink/StoreLink.jsx
@@ -1,0 +1,36 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { pickBy } from 'lodash'
+import { withClient } from 'cozy-client'
+
+class DumbStoreLink extends React.Component {
+  static propTypes = {
+    type: PropTypes.oneOf(['konnector', 'webapp']),
+    category: PropTypes.string
+  }
+
+  redirect = async () => {
+    const { client: cozyClient, type, category } = this.props
+    const redirectionURL = await cozyClient.intents.getRedirectionURL(
+      'io.cozy.apps',
+      pickBy({ type, category }, Boolean)
+    )
+
+    // We use `window.location` because on desktop we want to stay in the same tab/window
+    // and on mobile we want to open the user's browser instead of an inapp browser
+    // because in the onboarding flow, the user clicks on email links, which open
+    // the "native" browser (external Firefox for example) instead of the in-app one.
+    // This means login cookies are stored in the external browser.
+    // To prevent asking the user to login again, we have to use an external browser
+    // instead of the in app browser.
+    window.location = redirectionURL
+  }
+
+  render() {
+    return React.cloneElement(this.props.children, { onClick: this.redirect })
+  }
+}
+
+const StoreLink = withClient(DumbStoreLink)
+
+export default StoreLink
--- a/src/components/StoreLink/StoreLink.jsx
+++ b/src/components/StoreLink/StoreLink.jsx
@@ -9,12 +9,32 @@ class DumbStoreLink extends React.Component {
     category: PropTypes.string
   }
 
-  redirect = async () => {
-    const { client: cozyClient, type, category } = this.props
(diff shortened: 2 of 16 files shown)


src/ducks/reimbursements/HealthReimbursements.jsx
```diff
@@ -62,25 +62,37 @@ class DumbHealthReimbursements extends Component {
...
+        {pendingTransactions ? (
+          <TransactionsWithSelection
+            transactions={pendingTransactions}
+            brands={this.props.brands}
+            urls={this.props.urls}
+            withScroll={false}
+            className={styles.HealthReimbursements__transactionsList}
+          />
+        ) : (
+          <Padded className="u-pv-0">
```
ptbrowne: Why do we need a u-pv-0 ? Because the `p` has a padding ?

src/ducks/reimbursements/HealthReimbursements.styl
```diff
@@ -13,3 +13,8 @@
 
     +small-screen()
         margin-top 16px
+
+.Caption
+    font-size 14px
+    line-height 1.7
+    color var(--coolGrey)
```
ptbrowne: @gooz do we have already a Text class for this ?

src/components/Section/Section.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import cx from 'classnames'
+import styles from 'components/Section/Section.styl'
+
+const Section = props => {
```
ptbrowne: TODO send this to cozy-ui ?

ptbrowne: ping @y-lohse @GoOz I do not think we have a component like this in cozy-ui yet.

src/components/StoreLink/StoreLink.jsx
```diff
@@ -0,0 +1,36 @@
...
+    type: PropTypes.oneOf(['konnector', 'webapp']),
+    category: PropTypes.string
+  }
+
+  redirect = async () => {
+    const { client: cozyClient, type, category } = this.props
+    const redirectionURL = await cozyClient.intents.getRedirectionURL(
+      'io.cozy.apps',
+      pickBy({ type, category }, Boolean)
+    )
```
ptbrowne: Nit: The getRedirectionURL could be done sooner than the click so that the user does not have to wait.

src/ducks/reimbursements/HealthReimbursements.jsx
drazik: Yeah. Maybe we should add `horizontal` and `vertical` props to `Padded`, so we don't have to use utilities when we need padding in only one direction?

src/components/StoreLink/StoreLink.jsx
```diff
@@ -0,0 +1,36 @@
...
+    // because in the onboarding flow, the user clicks on email links, which open
+    // the "native" browser (external Firefox for example) instead of the in-app one.
+    // This means login cookies are stored in the external browser.
+    // To prevent asking the user to login again, we have to use an external browser
+    // instead of the in app browser.
+    window.location = redirectionURL
+  }
+
+  render() {
+    return React.cloneElement(this.props.children, { onClick: this.redirect })
```
ptbrowne: If we reuse this pattern, we could do something like the AppLinker.

```
<RedirectionLink doctype='io.cozy.apps' options={options} target={{ mobile: '_blank' }}>{
  (onClick, href) => <a href={href} onClick={onClick}>Open </a>
} />
```

src/ducks/settings/AddAccountLink.jsx
```diff
@@ -1,69 +1,12 @@
...
 
 const AddAccountLink = props => {
-  // For now we redirect on collect on both mobile app and browsers
-  // since this is not possible to show a waiting message
-  // const Link = __TARGET__ === 'mobile' ? NewWindowLink : IntentLink
-  return <SameWindowLink {...props} />
+  return (
+    <StoreLink type="konnector" category="banking">
+      {props.children}
+    </StoreLink>
```
ptbrowne: :+1:

src/ducks/reimbursements/HealthReimbursements.styl
GoOz: We do but the font-size is 12px and the line-height is 1.2

drazik: I didn't find a Text class with these styles. I asked the design team if it's a new style of text that we should put in cozy-ui. I think we can merge as is and refactor when we have their answer.

src/components/Section/Section.jsx
drazik: I asked the design team if this is something generic that will land in other apps

PR #1266, opened 2019-06-11
Update raven-js to v3.27.2

drazik:
This is not the reason why communication with sentry failed (see https://github.com/cozy/cozy-doctypes/pull/92 for that), but this way we use the latest raven-js version. We must keep in mind that this SDK is now deprecated and only receives bugfixes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
     "number-to-locale-string": "1.2.0",
     "piwik-react-router": "0.8.2",
     "pouchdb-adapter-cordova-sqlite": "2.0.5",
-    "raven-js": "3.27.0",
+    "raven-js": "3.27.2",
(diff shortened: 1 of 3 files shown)


ptbrowne:
> We must keep in mind that this SDK is now deprecated and only receives bugfixes.

What is the recommended way of using Sentry then ?

drazik:
> What is the recommended way of using Sentry then ?

They have a new SDK : `@sentry/browser` and `@sentry/node`. See https://docs.sentry.io/error-reporting/quickstart/?platform=browsernpm and https://docs.sentry.io/error-reporting/quickstart/?platform=node

Migration path for basic usage is at https://github.com/getsentry/sentry-javascript/releases/tag/4.0.0. Everything we use does not seem to be included in it.

PR #1269, opened 2019-06-12
feat: Transfer page flow

ptbrowne:
Revamp transfer page, use swipeable views.

- Bottom button in a Portal
- Text cards can show choices of the user in a text, can be clicked to
change the value (going back to the slide)


src/components/BackButton/index.jsx
```diff
@@ -37,17 +51,13 @@ const BackButton = ({
...
       ? getCssVariableValue('primaryContrastTextColor')
       : getCssVariableValue('coolGrey')
+  const handleClick = (onClick = onClick || (() => to && router.push(to)))
   return isMobile ? (
-    <BarLeft>
-      <button className="coz-bar-btn coz-bar-burger" onClick={handleClick}>
-        <Icon icon={arrowLeft} color={arrowColor} />
-      </button>
-    </BarLeft>
+    <BackButtonMobile onClick={this.handleClick} theme={theme} />
```
kosssi: You don't used the const `handleClick` ?

ptbrowne: Fixed in a next PR, thx

src/components/Stepper.jsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+/**
+ * Controlled component responsible for containing a list of views like
+ * a carousel. Progression bar at the top and back button displayed when
+ * not on first view.
+ */
+class Stepper extends React.Component {
+  render() {
+    const { current, children, onBack } = this.props
```
drazik: `currentIndex` would make it clearer that it's an index

src/components/List/index.jsx
```diff
@@ -9,13 +10,16 @@ export const Radio = props => {
...
-  const { className, bordered, ...rest } = props
+  const { className, border, paper, ...rest } = props
 
   return (
     <div
       className={cx(
         {
-          [styles['List--bordered']]: bordered
+          [styles['List--bordered']]: border === true,
+          [styles['List--papered']]: paper === true,
```
drazik: I think we should set `papered` class for any truthy value. The user of the component can do something like

```jsx
<List papered={something && somethingElse}>
	{/* ... */}
</List>
```

Where `something` and `somethingElse` can be truthy value, so the result of the expression will be a truthy value too.

src/ducks/transfers/TransferPage.jsx
```diff
@@ -79,15 +98,58 @@ const ChooseRecipientCategory = translate()(
   }
 )
 
+const BeneficiaryRow = ({ beneficiary, onSelect }) => {
+  return (
+    <Row
+      className="u-clickable"
+      onClick={onSelect.bind(null, beneficiary)}
+      key={beneficiary._id}
```
drazik: Why do you use a `key` outside of a `map` ?

src/ducks/transfers/TransferPage.jsx
```diff
@@ -79,15 +98,58 @@ const ChooseRecipientCategory = translate()(
...
+    >
+      <Media className="u-w-100">
+        <Img />
```
drazik: Why do we need this empty `Img`?

src/ducks/transfers/TransferPage.jsx
```diff
@@ -79,15 +98,58 @@ const ChooseRecipientCategory = translate()(
...
+        <Bd>
+          <Text>{beneficiary.label}</Text>
+          <Caption>{beneficiary.iban}</Caption>
+        </Bd>
+        <Img>{beneficiary.balance ? <Bold>{beneficiary.balance}</Bold> : null}</Img>
```
drazik: Shouldn't the condition wrap the `Img`?

src/components/Stepper.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+const styles = {
+  percBackground: {
+    backgroundColor: 'var(--silver)'
+  },
+  percLine: {
+    height: 5,
+    marginBottom: '0.5rem'
+  }
+}
```
drazik: These styles are not dynamic so they should be extracted in a CSS file, no?

src/ducks/transfers/TransferPage.jsx
ptbrowne: Forgotten key, thanks

src/components/Stepper.jsx
ptbrowne: Done

src/ducks/transfers/TransferPage.jsx
ptbrowne: Done

src/components/List/index.jsx
ptbrowne: I am not sure about that, I prefer to be precise about the types. What should I put in proptypes to support `papered={something && somethingElse}` ?

ptbrowne:
> il n'y a aucun test

C'est vrai je vais en rajouter, merci du rappel.

src/components/List/index.jsx
drazik: Since any type is either truthy or falsy, there's nothing more precise than `PropTypes.any`. I understand your point of view, we can merge as is, that's not so important.

PR #1270, opened 2019-06-12
Use late health reimbursement notification value to show status

drazik:
Before, we showed that a health reimbursement is late if it was pending for more than 1 month. Now, we want to use the late health reimbursement notification value instead of the hard coded `1 month`.

To do that, we need the setting. Since we don't want to fetch the settings everytime we want to show a reimbursement status, and we will need it in services too, we store the value in a flag just like we did for the local categorization model override setting.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -23,6 +23,10 @@ const ReactHint = ReactHintFactory(React)
 const App = props => {
   const settings = getDefaultedSettingsFromCollection(props.settingsCollection)
   flag('local-model-override', settings.community.localModelOverride.enabled)
+  flag(
+    'late-health-reimbursement-limit',
+    settings.notifications.lateHealthReimbursement.value
+  )
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/App.jsx
```diff
@@ -23,6 +23,10 @@ const ReactHint = ReactHintFactory(React)
 const App = props => {
   const settings = getDefaultedSettingsFromCollection(props.settingsCollection)
   flag('local-model-override', settings.community.localModelOverride.enabled)
+  flag(
+    'late-health-reimbursement-limit',
+    settings.notifications.lateHealthReimbursement.value
```
ptbrowne: Should we have  `|| 30` here ?

drazik: The settings collection has default values thanks to the line 24:

```js
const settings = getDefaultedSettingsFromCollection(props.settingsCollection)
```

In these default values, we have `30`.

PR #1275, opened 2019-06-13
Add late health reimbursement notification

drazik:
This PR adds the logic of late health reimbursement notification sending (mail + push). The logic is the following:

We fetch the transactions between `now - n days` (where `n` is the value the user configured in his settings) and 6 months before this date. Then we keep only the transactions that are health reimbursements that are in a pending state and that have not already been notified. We then send the notification and update these transactions, adding a `notifications` object to their `cozyMetadata`:

```js
notifications: {
  lateHealthReimbursement: ['2019-06-13T...']
}
```


manifest.webapp
```diff
@@ -99,6 +99,14 @@
       "multiple": false,
       "default_priority": "normal",
       "templates": {}
+    },
+    "late-health-reimbursement": {
+      "description": "Alert the user when a health reimbursement is late",
+      "collapsible": false,
+      "stateful": false,
+      "multiple": false,
+      "defualt_priority": "normal",
```
ptbrowne: default

src/ducks/notifications/LateHealthReimbursement.js
```diff
@@ -0,0 +1,210 @@
...
+import { Bill } from 'models'
+import { getReimbursementBillId, getReimbursementBillIds } from './helpers'
+
+const log = logger.namespace('lateHealthReimbursement')
+
+const ACCOUNT_SEL = '.js-account'
+const DATE_SEL = '.js-date'
+const TRANSACTION_SEL = '.js-transaction'
+
+const toText = cozyHTMLEmail => {
```
ptbrowne: Can you add a doc on what this does ?

src/ducks/notifications/LateHealthReimbursement.js
```diff
@@ -0,0 +1,210 @@
...
+    log('info', `Fetching transactions between ${gt} and ${lt}`)
+    const transactionsInDateRange = await BankTransaction.queryAll({
+      date: {
+        $gt: gt,
+        $lt: lt
+      }
+    })
+    log(
+      'info',
+      `Fetched ${
```
ptbrowne: Number at the beginning of the log to keep the same convention as we discussed ;)

src/ducks/notifications/LateHealthReimbursement.js
```diff
@@ -0,0 +1,210 @@
...
+  return utils.toText(cozyHTMLEmail, getContent)
+}
+
+class LateHealthReimbursement extends Notification {
+  constructor(config) {
+    super(config)
+    this.interval = config.value
+  }
+
+  isAlreadyNotified(transaction) {
```
ptbrowne: Can you extract this to transaction helpers ? ducks/transactions/helpers

src/ducks/notifications/Notification.js
```diff
@@ -54,12 +54,16 @@ class Notification {
 
       log('info', `Send notifications with category: ${attributes.category}`)
       const cozyClient = this.cozyClient
-      return cozyClient.fetchJSON('POST', '/notifications', {
+      await cozyClient.fetchJSON('POST', '/notifications', {
         data: {
           type: 'io.cozy.notifications',
           attributes
         }
       })
```
ptbrowne: Are we sure that the result was not used by the caller ?

src/ducks/notifications/LateHealthReimbursement.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  getPushContent(transactions) {
+    return this.t(
+      'Notifications.when_late_health_reimbursement.notification.content.message',
+      { smart_count: transactions.length }
+    )
+  }
+
+  async onSendNotificationSuccess() {
```
ptbrowne: docstring ? What should it do ?

src/ducks/notifications/helpers.js
```diff
@@ -1,7 +1,27 @@
...
+        return (
+          transaction.reimbursements &&
+          transaction.reimbursements.map(getReimbursementBillId)
+        )
+      })
+    )
+  ).filter(Boolean)
+
+  return billIds
+}
```
ptbrowne: I think those helpers should be in ducks/transactions/helpers

src/locales/fr.json
```diff
@@ -443,7 +443,17 @@
     "when_late_health_reimbursement": {
       "settingTitle": "Remboursement de santé en retard",
       "description": "Vous recevrez une notification lorsqu'un remboursement de santé est en attente depuis plus de",
-      "unit": "jours"
+      "unit": "jours",
+      "email": {
+        "innerTitle": "Notification de remboursement en retard"
```
ptbrowne: remboursement(s) ?

src/ducks/notifications/Notification.js
drazik: Yes

PR #1295, opened 2019-06-19
feat: Override file

ptbrowne:
Override file can contain fullVersion. This is used to dynamically insert
the version inside config.xml via the xsl stylesheet.


src/targets/mobile/fastlane/Appfile
```diff
@@ -1,5 +1,5 @@
 # ios
-app_identifier "io.cozy.banks"
+app_identifier "io.cozy.orange.assistantfinancier"
```
Crash--: :-°

PR #1311, opened 2019-06-25
fix: Remove keyboard plugin since it works on iOS 10 & 12 but not on 11

Crash--:
Actually the plugin works great on iOS 10 and iOS 12 but not on iOS 11. So removing it ATM.


ptbrowne:
Thanks !!

PR #1320, opened 2019-06-26
Delayed debit alert UI

drazik:
This PR introduces the settings UI for the new delayed debit alert.

The user can select a credit card and a checkings account. The association is then saved in the settings and will be used (in another PR) to alert the user when the credit card account coming balance is greater than the checkings account balance.

For now, only one association can be selected. But we already store an array of associations so it will be easy to allow the user to select multiple associations in near future.

The translations are obviously not the good ones, but this is behind the `delayed-debit-alert` flag for now. I will update it as soon as I have it.

Live on https://testbanksdelayeddebitalert-banks.cozy.works/#/settings


src/ducks/settings/DelayedDebitAlert.styl
```diff
@@ -0,0 +1,17 @@
...
+        margin-right 0.25rem
+
+        &:first-child
+            margin-left 0
+
+        &:last-child
+            margin-right 0
+
+    & + &
+        margin-top 0.5rem
```
ptbrowne: Looks like a stack layout here ;)

src/ducks/settings/DelayedDebitAlert.jsx
```diff
@@ -0,0 +1,125 @@
...
+    association,
+    creditCardOptions,
+    checkingsOptions,
+    enabled,
+    onChange,
+    ...rest
+  } = props
+
+  return (
+    <div className={styles.AccountsPicker} {...rest}>
```
ptbrowne: It's nice when the component and style have the same name, can you pick one ?

src/ducks/settings/DelayedDebitAlert.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+    const accountsByType = pick(groupBy(accounts, getAccountType), [
+      'CreditCard',
+      'Checkings'
+    ])
+
+    const {
+      CreditCard: creditCardOptions,
+      Checkings: checkingsOptions
+    } = mapValues(accountsByType, accounts =>
```
ptbrowne: `mapValues(accountsByType, makeSelectValue)` would make it a bit clearer and `makeSelectValue` could be extracted from `render`.

src/ducks/settings/DelayedDebitAlert.jsx
```diff
@@ -0,0 +1,125 @@
...
+        value: account._id,
+        label: getAccountLabel(account)
+      }))
+    )
+
+    return (
+      <ToggleRowWrapper>
+        <ToggleRow
+          title={t('Notifications.delayed_debit.settingTitle')}
+          description={t('Notifications.delayed_debit.description')}
```
ptbrowne: snake_case then camelCase ?

src/ducks/settings/DelayedDebitAlert.jsx
```diff
@@ -0,0 +1,125 @@
...
+    } = this.props
+
+    const accountsByType = pick(groupBy(accounts, getAccountType), [
+      'CreditCard',
+      'Checkings'
+    ])
+
+    const {
+      CreditCard: creditCardOptions,
+      Checkings: checkingsOptions
```
ptbrowne: It seems that `checkingsOptions` and `creditCardOptions` are used even if they are undefined in `AccountsAssociationSelect`. This would break I think since the `find` method is called on them to find the current selected value.

src/ducks/settings/DelayedDebitAlert.jsx
drazik: Yeah, we have a lot of inconsistencies in translations keys. I tried to stay consistent with inconsistencies here. I opened an issue about this: https://github.com/cozy/cozy-banks/issues/1322

src/ducks/settings/DelayedDebitAlert.jsx
```diff
@@ -95,34 +100,47 @@ class DumbDelayedDebitAlert extends React.Component {
...
+        {hasNoAccount && (
+          <Text>{t('Notifications.delayed_debit.noAccount')}</Text>
+        )}
+        {!hasNoAccount && hasNoCreditCard && (
+          <Text>{t('Notifications.delayed_debit.noCreditCard')}</Text>
+        )}
+        {!hasNoAccount && hasNoCheckings && (
+          <Text>{t('Notifications.delayed_debit.noCheckings')}</Text>
+        )}
+        {!hasNoAccount && !hasNoCreditCard && !hasNoCheckings && (
```
ptbrowne: Nit: Having a boolean variable as a negative is a bit confusing IMHO. I prefer hasAccount and !hasAccount.

PR #1335, opened 2019-06-27
feat(DelayedDebitAlert): Use bank accounts relationships

drazik:
Previously the delayed debit alert UI was using an object like `{ creditCardAccount: id, checkingsAccount: id  }` stored in `io.cozy.bank.settings`. But after thinking about what could happen when a user removes an account that is part of the associations, I thought this solution was not good enough because we would have to manually update the object everytime the user removes an account.

I think using a relationship on the account is better because it automatically handles bank accounts removal:

* If the credit card account is removed, then the relationship is also removed
* If the checkings account is removed, then the relationship is null

There are two remaining things to do:

* ~~Update cozy-client after https://github.com/cozy/cozy-client/pull/473 has been merged (we are using `HasOne::set` and `HasOne::unset` here)~~ (done in https://github.com/cozy/cozy-banks/pull/1335/commits/bb17d512df6fa2d744e9b68a3edd625a5e06dc47)
* Add tests

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "cozy-authentication": "1.14.0",
     "cozy-bar": "6.28.6",
     "cozy-ci": "0.2.0",
-    "cozy-client": "6.46.0",
+    "cozy-client": "6.48.0",
     "cozy-client-js": "0.16.4",
     "cozy-device-helper": "1.7.2",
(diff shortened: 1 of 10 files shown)


drazik:
We need it right now and nobody can review it. I'm merging it but please feel free to make any comment and changes suggestion :pray:

PR #1358, opened 2019-07-01
test(TransactionSelectDates): Don't fail on each new month

drazik:
Mock the date so we don't have problem on each new month.

This fixes the issues we currently have with tests. See https://travis-ci.org/cozy/cozy-banks/jobs/552657202#L642


kosssi:
👍

PR #1391, opened 2019-07-08
Configure auto merge

ptbrowne:
I want to add this for very small PRs : https://github.com/bobvanderlinden/probot-auto-merge

The idea being that for very small PRs, we can label when creating the issue with "merge" and
as soon as it is reviewed, it is automatically merged.

### DIFF
--- a/auto-merge.yml
+++ b/auto-merge.yml
@@ -1,0 +1,4 @@
+min-approvals:
+  MEMBER: 1
+requiredLabels:
+- merge


auto-merge.yml
```diff
@@ -0,0 +1,4 @@
+min-approvals:
+  COLLABORATOR: 1
+requireLabes:
```
Crash--:
```suggestion
requiredLabels
```

auto-merge.yml
```diff
@@ -0,0 +1,4 @@
...
+- merg
```
Crash--:
```suggestion
- merge
```

Crash--:
I like the idea! 

I also like the `deleteBranchAfterMerge` option!

ptbrowne:
Thanks for the fixes Quentin 😅

> `deleteAfterMerge` option

I already added https://probot.github.io/apps/delete-merged-branch/ so I think it will do the job :)

auto-merge.yml
```diff
@@ -0,0 +1,4 @@
+min-approvals:
+  COLLABORATOR: 1
```
Crash--: I would be more confortable with `member`

auto-merge.yml
```diff
@@ -0,0 +1,4 @@
...
+requiredLabels
```
Crash--:
```suggestion
requiredLabels:
```

sorry :'(

auto-merge.yml
ptbrowne: Done, you're right, it's better.

auto-merge.yml
ptbrowne: My bad.. what is wrong with me !

PR #1393, opened 2019-07-09
Stats service

drazik:
This service is executed when a new `io.cozy.bank.operations` document is created (with a 3mn debounce). It fetches the transactions of the last 3 completed months and aggregates the following data:

* Mean activity income
* Mean complementary income
* Mean mortgage expense
* Mean total loans expenses

The data is stored in [`io.cozy.bank.accounts.stats` documents](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.bank.md#iocozybankaccountsstats) (one for each bank account).

The result looks like:

```json
{
  "_id": "e574571309e5f486aebf284cf70176d1",
  "_rev": "3-cdcd0fcdf646b1c8233dab311871807d",
  "additionalIncome": 0,
  "cozyMetadata": {
    "createdAt": "2019-07-09T08:21:35.192Z",
    "metadataVersion": 1,
    "updatedAt": "2019-07-09T08:26:01.139Z",
    "updatedByApps": [
      {
        "date": "2019-07-09T08:26:01.139Z"
      }
    ]
  },
  "currency": "EUR",
  "id": "e574571309e5f486aebf284cf70176d1",
  "income": 2576.9933333333333,
  "loans": 410.3333333333333,
  "mortgage": 410.3333333333333,
  "periodEnd": "2019-06-30T21:59:59.999Z",
  "periodStart": "2019-02-28T23:00:00.000Z",
  "relationships": {
    "account": {
      "data": {
        "_id": "compteisa1",
        "_type": "io.cozy.bank.accounts"
      }
    }
  }
}
```


config/webpack.config.dev.js
```diff
@@ -25,9 +25,7 @@ module.exports = {
     new webpack.DefinePlugin({
       __STACK_ASSETS__: false,
       __DEV__: true,
-      'process.env': {
-        NODE_ENV: JSON.stringify('development')
-      }
+      'process.env.NODE_ENV': JSON.stringify('development')
```
ptbrowne: I dont understand well the difference ? What is the purpose ?

src/ducks/stats/services.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    MockDate.set(new Date('2019-07-08'))
+
+    const period = getPeriod()
+
+    expect(period).toMatchSnapshot()
+  })
+})
+
+describe('fetchTransactionsForPeriod', () => {
+  it('should fetch transaction for the given period', async () => {
```
ptbrowne: s/transactions/transactions

src/ducks/stats/services.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+jest.mock('cozy-doctypes')
+
+describe('getPeriod', () => {
+  afterEach(() => {
+    MockDate.reset()
+  })
+
+  it('should return the period', () => {
+    MockDate.set(new Date('2019-07-08'))
```
ptbrowne: <3

src/targets/services/stats.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+const client = new CozyClient({
+  uri: process.env.COZY_URL.trim(),
+  schema,
+  token: process.env.COZY_CREDENTIALS.trim()
+})
+
+BankTransaction.registerClient(client)
```
ptbrowne: I think we should do it on Document, since we can run into errors later when we find out that it was only registered on BankTransaction.

src/targets/services/stats.js
```diff
@@ -0,0 +1,103 @@
...
+  )
+
+  Object.entries(transactionsByAccount).forEach(([accountId, transactions]) => {
+    // TODO get `local-model-override` setting and put it in a flag
+    const transactionsByCategory = groupBy(transactions, getCategoryId)
+
+    const stats = {
+      periodStart: period.start,
+      periodEnd: period.end,
+      income: getMeanOnPeriod(transactionsByCategory['200110'], period),
```
ptbrowne: '200110' could be INCOME_CATEGORY

src/targets/services/stats.js
```diff
@@ -0,0 +1,103 @@
...
+        transactionsByCategory['200180'],
+        period
+      ),
+      mortgage: getMeanOnPeriod(transactionsByCategory['401010'], period),
+      loans: getMeanOnPeriod(
+        [
+          ...(transactionsByCategory['401010'] || []),
+          ...(transactionsByCategory['400120'] || []),
+          ...(transactionsByCategory['400930'] || []),
+          ...(transactionsByCategory['400210'] || [])
```
ptbrowne: Its a bit hard to understand which categories are taken into account here, constants could be used ?

src/targets/services/stats.js
```diff
@@ -44,9 +46,15 @@ const client = new CozyClient({
...
-BankTransaction.registerClient(client)
-
 const main = async () => {
+  Document.registerClient(client)
+
+  log('info', 'Fetching settings...')
+  let setting = await Settings.fetchWithDefault()
+
+  // The flag is needed to use local model when getting a transaction category ID
+  flag('local-model-override', setting.community.localModelOverride.enabled)
```
ptbrowne: Are we using a flag for a user preference here ? I think flags should not be used for that. Maybe we could have an issue about this ?

config/webpack.config.dev.js
drazik: The previous form was the same as doing

```js
process.env = { NODE_ENV: 'development' }
```

We lost all other `process.env` properties (including `COZY_*` provided by the stack). Now we just provide the `NODE_ENV` property without touching other ones.

src/targets/services/stats.js
drazik: The `getCategoryId` needs to know if the user enabled a setting (local model override) or not. This helper is used in multiple contexts and also in other helpers. So we would have to fetch and propagate the settings  in a lot of places. We decided to put this setting in a flag to avoid that. But yeah maybe we can find a better solution

src/targets/services/stats.js
drazik: You're right, I will use constants.

config/webpack.config.dev.js
ptbrowne: Nice catch, thanks for the explanation.

src/ducks/stats/services.js
```diff
@@ -0,0 +1,33 @@
...
+  const end = endOfMonth(subMonths(today, 1))
+  const start = startOfMonth(subMonths(end, 2))
+
+  return { start, end }
+}
+
+export const fetchTransactionsForPeriod = period => {
+  return BankTransaction.queryAll({
+    date: {
+      $gt: period.start,
```
edas: Shouldn't it be $gte ?

src/ducks/stats/services.js
```diff
@@ -0,0 +1,33 @@
...
+  endOfMonth,
+  subMonths,
+  differenceInMonths
+} from 'date-fns'
+import { BankTransaction } from 'cozy-doctypes'
+import { sumBy } from 'lodash'
+
+export const getPeriod = () => {
+  const today = new Date()
+  const end = endOfMonth(subMonths(today, 1))
```
edas: This will give you a locale-based end of month. For a server in UTC+2, you will get a time 23:59:59 UTC+2, so 22:00:00Z. Are you sure this is want you want ?

src/ducks/stats/services.js
drazik: Yes :+1:

src/ducks/stats/services.js
drazik: I refactored it using moment so we manipulate UTC dates.

src/ducks/stats/services.js
```diff
@@ -0,0 +1,36 @@
...
+    .startOf('month')
+
+  return { start, end }
+}
+
+export const fetchTransactionsForPeriod = period => {
+  return BankTransaction.queryAll({
+    date: {
+      $gte: period.start,
+      $lte: period.end
```
edas: I didn't see that before. You are doing "start of this month" -> "end of this month", both included.

If someone put fractions of seconds in the date, you may exclude a few documents. This is more a theorical case, but in the future you're better to do "start of this month" (included) -> "start of next month" (excluded)

src/ducks/stats/services.js
```diff
@@ -0,0 +1,36 @@
+import moment from 'moment'
+import { BankTransaction } from 'cozy-doctypes'
+import { sumBy } from 'lodash'
+
+export const getPeriod = () => {
+  const end = moment()
+    .utc()
```
edas: I've done my example in UTC yesterday but Patrick tolds me you have all your transactions written un Europe/Paris so that's probably what you want to ask here

src/ducks/stats/services.spec.js
```diff
@@ -0,0 +1,64 @@
+import {
+  getPeriod,
+  fetchTransactionsForPeriod,
+  getMeanOnPeriod
+} from './services'
+import MockDate from 'mockdate'
+import { BankTransaction } from 'cozy-doctypes'
+import { endOfDay } from 'date-fns'
```
edas: If you use moment.js in your code, you'd probably best to do the same here

src/ducks/stats/services.js
drazik: I just tried to use moment-timezone to set the Europe/Paris timezone, but it doesn't work well with webpack: https://github.com/moment/moment-timezone/issues/449

I tried the solution of this thread but none of them worked. I'm a bit stuck

PR #1394, opened 2019-07-09
feat: Attempt at VersionPlugin

ptbrowne:
Collects versions of packages and uses DefinePlugin to define a variable
containing those versions so we can use it in the code. The idea would be to have those versions in the Debug view.

### DIFF
--- a/config/VersionPlugin.js
+++ b/config/VersionPlugin.js
@@ -1,0 +1,55 @@
+const webpack = require('webpack')
+const { CachedInputFileSystem, ResolverFactory } = require('enhanced-resolve')
+const fs = require('fs')
+
+const pluginName = 'PackageVersionPlugin'
+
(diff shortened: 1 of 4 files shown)


config/VersionPlugin.js
```diff
@@ -0,0 +1,55 @@
...
+class VersionPlugin {
+  constructor(options) {
+    this.varName = options.varName || '__VERSIONS__'
+    this.packages = options.packages
+  }
+  async apply(compiler) {
+    // Create a resolver with the same options as the compiler
+    const resolver = ResolverFactory.createResolver({
+      fileSystem: new CachedInputFileSystem(fs, 4000),
+      extensions: ['.js', '.json'],
```
Crash--: Do we need to look at a `js` file since we just want `json`?

PR #1400, opened 2019-07-10
Compute additional income using all income subcategories except activity income

drazik:
We previously used only the additional income subcategory, but we actually want to include all incomes except activity incomes in this sum.


src/targets/services/stats.js
```diff
@@ -82,34 +82,44 @@ const main = async () => {
       const accountStats = statsByAccountId[accountId] || {}
       const transactionsByCategory = groupBy(transactions, getCategoryId)
 
+      const getTransactions = categoryName => {
```
y-lohse: nit: `getTransactionsByCategoryName`

PR #1413, opened 2019-07-12
feat: Trying to validate bundle

ptbrowne:
It is not the first time that we have errors on old browsers (ios 10)
for syntax errors. It is costly to understand where those errors come
from when there are on the wild. IMHO it'd be better be caught as soon
as possible, for example during CI.

2 experiments here :

- Trying to validate the bundle via eslint but could not 
  - parsing error with the default parser (espree) (async function)
  - too slow with babel-eslint (10m and counting)
  - out of memory with esprima 
- Custom validation via esquery

Details in the commit messages.

### DIFF
--- a/scripts/bundle-eslintrc
+++ b/scripts/bundle-eslintrc
@@ -1,0 +1,8 @@
+{
+  "env": {
+    es6: true
+  },
+  "parser": 'babel-eslint',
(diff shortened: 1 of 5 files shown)


ptbrowne:
Closing in favor of #1423

PR #1416, opened 2019-07-15
Upgrade node10

Crash--:
libxslt is not compatible with node 10 (https://github.com/albanm/node-libxslt/issues/68) so I changed to `xsltproc` so I needed to rewrite some part of `customizeConfig.js`

Lot of `yarn.lock` changes since I had to remove a lot of package before being able to install the right package versions.

Seems that I had to regenerate a few snapshots because of some indents 

The idea behind was to be able to use this webpack plugin https://www.npmjs.com/package/webpack-plugin-ramdisk


ptbrowne:
> so I changed to xsltproc

Woot, this means that I will not have the native library always rebuilding.

src/targets/mobile/scripts/customizeConfigXML.js
```diff
@@ -64,29 +40,31 @@ const customizeConfigXML = function(context) {
...
+        .replace('$IOS_VERSION_CODE', overrideConfig.fullVersion)
+        .replace('$USER_AGENT_VERSION', 'io.cozy.banks.mobile-' + versionCode)
 
       fs.writeFileSync(configPath, output)
-      console.log(
-        `config.xml successfully transformed using ${transformFilePath}`
-      )
-    })
+    }
+    fs.unlinkSync(tmpPath)
```
ptbrowne: The temp file is not unlinked if there is a problem. It is not particularly important here but when there are resources, it is good to have the habit of 

```
initializeResource()
try {
 doStuffThatMightThrow()
} finally {
  destroyResource()
}
```

In python, this is built into the language with the `with` statement. A similar approach could be used in JS.

```
withTempFile(tempFile => {
 // do stuff
})
```

ptbrowne:
:+1: Thanks !

src/targets/mobile/scripts/customizeConfigXML.js
Crash--: Is it better now?

ptbrowne: Yes :) Thanks !

PR #1437, opened 2019-07-30
Repair bills matching visualizer

drazik:
The visualizer was not working after updates on the bills matching code. This repairs it and update the documentation.

### DIFF
--- a/scripts/visualizer/src/server/index.js
+++ b/scripts/visualizer/src/server/index.js
@@ -1,13 +1,11 @@
-#!/usr/bin/env node
-
 /**
  * Visualizer to see links between transactions and bills.
  *
  * Launch with `cozy-run-dev visualizer/index.js`
  */
-const Linker = require('../../src/ducks/billsMatching/Linker/Linker')
+const Linker = require('ducks/billsMatching/Linker/Linker').default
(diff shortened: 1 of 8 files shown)


scripts/visualizer/src/server/index.js
```diff
@@ -17,6 +15,7 @@ class DryLinker extends Linker {
 }
 
 const generate = async options => {
+  // return 'pouet'
```
ptbrowne: :)

scripts/visualizer/src/server/index.js
```diff
@@ -5,9 +5,11 @@
  */
 const Linker = require('ducks/billsMatching/Linker/Linker').default
 const { cozyClient } = require('cozy-konnector-libs')
-const { fetchAll } = require('ducks/billsMatching/utils')
+const { Bill } = require('models')
 const path = require('path')
 
+Bill.registerClient(cozyClient)
```
ptbrowne: You can use Document here, it will be the same but it will be less confusing. Here another class extending Document will also be registered which might be surprising, it is less surprising if you use Document since it is the mother class.

scripts/visualizer/src/server/index.js
drazik: :rofl:

PR #1465, opened 2019-08-21
Handle bills matching criterias in matching algorithm

drazik:
The `io.cozy.bills` documents can now carry some matching criterias (see https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.bills.md#optional-attributes-but-some-are-important-depending-the-context). This PR updates the matching algorithm so these criterias are taken into account if they exist. If a bill doesn't have its own matching criterias, the previous defaults criterias apply.

This PR misses the combined bills with criterias case. I'm going to handle it in its own PR so this one is not delayed in its entirety.

### DIFF
--- a/src/ducks/billsMatching/Linker/billsToOperation/helpers.js
+++ b/src/ducks/billsMatching/Linker/billsToOperation/helpers.js
@@ -21,9 +21,22 @@ const getIdentifiers = options => options.identifiers
 
 const getDateRangeFromBill = (bill, options) => {
   const date = getOperationDateFromBill(bill, options)
+
+  const pastWindow = get(
+    bill,
+    'matchingCriterias.dateLowerDelta',
+    options.pastWindow
+  )
+
+  const futureWindow = get(
+    bill,
+    'matchingCriterias.dateUpperDelta',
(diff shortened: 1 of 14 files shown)


src/ducks/billsMatching/Linker/billsToOperation/helpers.js
```diff
@@ -93,12 +94,26 @@ const sortedOperations = (bill, operations) => {
...
 }
 
+const getBillRegexp = bill => {
+  let regexpStr = bill.matchingCriterias && bill.matchingCriterias.labelRegex
+
+  if (!regexpStr) {
+    const [brand] = getBrands(
+      brand => brand.name === bill.vendor || brand.konnectorSlug === bill.vendor
+    )
+    regexpStr = brand ? brand.regexp : `\\b${bill.vendor}\\b`
```
ptbrowne: Should we check here that bill.vendor is defined ? Otherwise the regex would match everytime right ?

drazik: Yes good catch, I'm going to add this check

PR #1466, opened 2019-08-21
refactor: Use JSON for data used in tests

ptbrowne:
Since those test cases are data, I transformed them into JSON, they can
be more easily manipulated this way.

Evil one liner to transform them:

```fish
for file in (ls | grep -v index); set module (echo $file | sed 's/\.js$//'); node -e "data=require(\"./$module\");fs.writeFileSync(`$module-\${data.description.replace(/ /g,\"-\")}.json`, JSON.stringify(data, null, 2))"; end
```

EDIT, it was also necessary to convert undefined values to null values.

`transform.js`

```
const fs = require('fs')
fs.readdirSync(__dirname)
  .filter(x => x.endsWith('json'))
  .forEach(filename => {
    const data = JSON.parse(fs.readFileSync(filename))
    const expectedResult = data.expectedResult
    for (let key of Object.keys(expectedResult)) {
      const bill = expectedResult[key]
      if (!bill.debitOperation) { bill.debitOperation = null }
      if (!bill.creditOperation) { bill.creditOperation = null }
    }
    fs.writeFileSync(filename, JSON.stringify(data, null, 2))
  })
```


drazik:
Actually in the beginning they were JSON files but I found it's useful to be able to test that something didn't match by expecting the `creditOperation` and / or the `debitOperation` to be undefined, which is impossible using JSON.

PR #1475, opened 2019-08-22
Handle matching criterias in combined bills

drazik:
This was the last missing piece in https://github.com/cozy/cozy-banks/pull/1465. Now we also handle matching criterias when we try to combine bills by making the union of the different matching criterias. At the moment, we can't find a real application to this, but it ensures that the matching doesn't crash if it happens.

### DIFF
--- a/src/ducks/billsMatching/Linker/Linker.spec.js
+++ b/src/ducks/billsMatching/Linker/Linker.spec.js
@@ -542,5 +542,64 @@ describe('linker', () => {
         expect(combinedBill.originalDate).toBe('2018-03-10T00:00:00Z')
       })
     })
+
+    describe('mergeMatchingCriterias', () => {
+      it('should merge criterias', () => {
+        const bills = [
+          {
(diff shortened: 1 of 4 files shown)


src/ducks/billsMatching/Linker/Linker.spec.js
```diff
@@ -542,5 +542,58 @@ describe('linker', () => {
...
+              amountUpperDelta: 10,
+              dateLowerDelta: 20,
+              dateUpperDelta: 20
+            }
+          }
+        ]
+
+        const criterias = linker.mergeMatchingCriterias(bills)
+
+        expect(criterias.labelRegex).toBe('(\\binulogic\\b|(impot|impots))')
```
ptbrowne: Est ce quon peut rajouter un test pour vérifier que la regex créée marche bien sur inulogic et impot ?

ptbrowne:
LGTM

PR #1498, opened 2019-08-27
fix(Linker): Don't use vendor to check bill type

drazik:
Until now, we maintained a vendors list to check if a bill has the
health type or not. This is not very scalable because we have to add new
vendors when new connectors are released, and we can do typos. Using the
bill `type` property is much more reliable.

### DIFF
--- a/src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.js
+++ b/src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.js
@@ -1,4 +1,3 @@
-const includes = require('lodash/includes')
 const sumBy = require('lodash/sumBy')
 const isWithinRange = require('date-fns/is_within_range')
 const { log } = require('../../utils')
@@ -11,19 +10,6 @@ const {
 } = require('./helpers')
 
 // constants
-
-const HEALTH_VENDORS = [
-  'Ameli',
-  'Harmonie',
-  'Malakoff Mederic',
-  'MGEN',
-  'Generali',
-  'ascoreSante',
-  'EoviMCD',
-  'Humanis',
-  'Alan',
-  'lamutuellegenerale'
-] // TODO: to import from each konnector
 const HEALTH_EXPENSE_CAT = '400610'
 const HEALTH_INSURANCE_CAT = '400620'
 const UNCATEGORIZED_CAT_ID_OPERATION = '0' // TODO: import it from cozy-bank
@@ -45,7 +31,7 @@ const isUncategorizedOperation = operation => {
 }
 
 const isHealthBill = bill => {
-  return includes(HEALTH_VENDORS, bill.vendor)
+  return bill.type === 'health_costs'
 }
 
 // filters
@@ -58,7 +44,7 @@ const filterByBrand = bill => {
     }
 
     const label = operation.label.toLowerCase()
-    return label.match(regexp)
+    return Boolean(label.match(regexp))
   }
 
   brandFilter.name = 'byBrand'
--- a/src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
+++ b/src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
@@ -13,18 +13,18 @@ describe('operations filters', () => {
       const bill = { vendor: 'Trainline' }
       const fByBrand = filterByBrand(bill)
 
-      expect(fByBrand({ label: 'Trainline !!!' })).toBeTruthy()
-      expect(fByBrand({ label: 'Yes Trainline' })).toBeTruthy()
-      expect(fByBrand({ label: 'CapitainTrain' })).toBeFalsy()
+      expect(fByBrand({ label: 'Trainline !!!' })).toBe(true)
+      expect(fByBrand({ label: 'Yes Trainline' })).toBe(true)
+      expect(fByBrand({ label: 'CapitainTrain' })).toBe(false)
     })
 
     it("should generate a regexp with the bill vendor if the brand doesn't exist in the dictionnary", () => {
       const bill = { vendor: 'Tartanpion' }
       const fByBrand = filterByBrand(bill)
 
-      expect(fByBrand({ label: 'Chez Tartanpion !!!' })).toBeTruthy()
-      expect(fByBrand({ label: 'tartanpion 17/09' })).toBeTruthy()
-      expect(fByBrand({ label: 'Rien à voir' })).toBeFalsy()
+      expect(fByBrand({ label: 'Chez Tartanpion !!!' })).toBe(true)
+      expect(fByBrand({ label: 'tartanpion 17/09' })).toBe(true)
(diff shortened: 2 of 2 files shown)


src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.js
```diff
@@ -45,7 +31,7 @@ const isUncategorizedOperation = operation => {
 }
 
 const isHealthBill = bill => {
-  return includes(HEALTH_VENDORS, bill.vendor)
+  return bill.type === 'health_costs'
```
ptbrowne: Are you sure there is no risk to have false negatives here ? I mean that we must be sure that every vendor here correctly sets the bill.type. Shouldn't we do 

```
  return  bill.type === 'health_costs' || includes(HEALTH_VENDORS, bill.vendor)
```

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
```diff
@@ -31,9 +31,9 @@ describe('operations filters', () => {
       const bill = { matchingCriterias: { labelRegex: 'Bidule' } }
       const fByBrand = filterByBrand(bill)
 
-      expect(fByBrand({ label: 'Chez Bidule !!!' })).toBe(true)
-      expect(fByBrand({ label: 'bidule 20/08' })).toBe(true)
-      expect(fByBrand({ label: 'Ca match paaaas du tout' })).toBe(false)
+      expect(fByBrand({ label: 'Chez Bidule !!!' })).toBeTruthy()
+      expect(fByBrand({ label: 'bidule 20/08' })).toBeTruthy()
+      expect(fByBrand({ label: 'Ca match paaaas du tout' })).toBeFalsy()
```
ptbrowne: Truthy and falsy are not the same as true/false, it seems we are lowering the bar of the test here, what is the gain ?

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
```diff
@@ -78,8 +78,8 @@ describe('operations filters', () => {
   const HEALTH_INSURANCE_CAT = '400620'
 
   describe('filtering by category', () => {
-    fit('should only match bills with the right categoryId when the vendor is known to be a health insurance provider', () => {
-      const fByCategory = filterByCategory({ vendor: 'Ameli' })
+    it('should only match bills with the right categoryId when the bill type is health_costs', () => {
```
ptbrowne: fit -> it. oups :)

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
```diff
@@ -140,14 +140,9 @@ describe('operations filters', () => {
...
       const fByCategory = filterByCategory({ vendor: 'SFR' })
       expect(fByCategory({ manualCategoryId: HEALTH_EXPENSE_CAT })).toBeFalsy()
-      expect(
-        fByCategory({ automaticCategoryId: HEALTH_EXPENSE_CAT })
-      ).toBeFalsy()
       expect(fByCategory({ manualCategoryId: '0' })).toBeTruthy()
-      expect(fByCategory({ automaticCategoryId: '0' })).toBeTruthy()
       expect(fByCategory({})).toBeTruthy()
       expect(fByCategory({ manualCategoryId: '400611' })).toBeTruthy()
-      expect(fByCategory({ automaticCategoryId: '400611' })).toBeTruthy()
```
ptbrowne: For my information, why did you remove those lines ?

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.js
drazik: @doubleface checked that all health connectors sets the `bill.type` to `health_costs`, so I think it is safe

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
```diff
@@ -177,14 +172,14 @@ describe('operations filters', () => {
         label: 'Visite chez le médecin',
         _id: 'o1',
         date: new Date(2017, 11, 13),
-        automaticCategoryId: '400610'
+        manualCategoryId: '400610'
```
ptbrowne: Question: why do we change from automatic to manual.

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
drazik: Because the case in which we only have `automaticCategoryId` means that our categorization models have not been applied, and in that case the `getCategoryId` helper returns `null`. That's how we represent the "categorization pending" state when a transaction have been improted but not yet categorized by our categorization service.

In tests, I think it's better to use only `manualCategoryId` because when it is present, it is always used. So we don't have to think about all the specific cases implemented by `getCategoryId` in every test that uses categories. By always using `manualCategoryId`, we are certain that this category will be used.

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
drazik: See https://github.com/cozy/cozy-banks/pull/1498#discussion_r317960624

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
drazik: Actually I don't know why but these tests were passing and then they were failing. I don't know how they passed before since `fByBrand` returns the result of `label.match(regex)`, so it returns `null` or an array, but never a boolean. That's why I  used `toBeTruthy` and `toBeFalsy` to fix this. But another way would be to `return Boolean(label.match(regex))` and still use `toBe(false)` and `toBe(true)` in tests. What do you think?

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
drazik: This is on purpose

src/ducks/billsMatching/Linker/billsToOperation/operationsFilters.spec.js
ptbrowne: I think it is better to be fairly strict on what we return, so Boolean(label.match(regex)) is good to me.

> Be conservative in what you send, be liberal in what you accept

https://en.wikipedia.org/wiki/Robustness_principle

ptbrowne: Otherwise, at some point, someones sees that you return the regex result, and starts using it instead of considering it as a boolean, then you want to change the behavior and not use a regex engine, and you break the implicit API for the other person.

drazik: :+1:

PR #1508, opened 2019-08-27
feat(onOperationOrBillCreate): Find bills for all updated transactions

drazik:
When a user manually categorizes a transaction, we want to apply the
matching algorithm to it. For example, if a transaction is recategorized
to health expense category, maybe we can find a related bill that didn't
match previously because the category was not good.

I see a possible problem with that though: the service will be run twice when a transaction matched with a bill, because the matching updates the transaction. I don't know how I can do better than that actually. The only solution I have in mind is to compare the document and its previous revision to see precisely what changed and skip some parts of the service if needed. But the service will still be run twice.

I think it shows that we need to split the `onOperationOrBillCreate` service into multiple ones. Any thoughts?

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -66,7 +66,7 @@
     "onOperationOrBillCreate": {
       "type": "node",
       "file": "onOperationOrBillCreate.js",
-      "trigger": "@event io.cozy.bills:CREATED",
+      "trigger": "@event io.cozy.bills:CREATED io.cozy.bank.operations:UPDATED:!=:manualCategoryId",
       "debounce": "75s"
     },
     "categorization": {
--- a/src/targets/services/onOperationOrBillCreate.js
+++ b/src/targets/services/onOperationOrBillCreate.js
@@ -9,7 +9,6 @@ import matchFromBills from 'ducks/billsMatching/matchFromBills'
 import matchFromTransactions from 'ducks/billsMatching/matchFromTransactions'
 import { logResult } from 'ducks/billsMatching/utils'
 import { findAppSuggestions } from 'ducks/appSuggestions/services'
-import { isNew as isNewTransaction } from 'ducks/transactions/helpers'
 import { fetchChangesOrAll, getOptions } from './helpers'
 
 const log = logger.namespace('onOperationOrBillCreate')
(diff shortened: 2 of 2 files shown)


ptbrowne:
> The only solution I have in mind is to compare the document and its previous revision to see precisely what changed and skip some parts of the service if needed. But the service will still be run twice.

Yes it is a problem, I wonder if it would be a good idea to deactivate modifications from a service
to trigger other services. I am a bit worried about infinite loops of triggers between services. I think we need to keep this in mind, have you already thought about that @nono ? It reminds me of the trick you did on the account deletion not to trigger the service twice ?

> I think it shows that we need to split the onOperationOrBillCreate service into multiple ones. Any thoughts?

I fully agree with that, I think it will also be good for monitoring. We need to split it so that the common JS between the two is not duplicated though.

ptbrowne:
Another solution when the services are splitted would be to support the trigger of a service only on modification of some attributes.

```patch
-       "trigger": "io.cozy.bank.operations:UPDATED",
+       "trigger": "io.cozy.bank.operations:category:UPDATED",
```

ptbrowne:
> The only solution I have in mind is to compare the document and its previous revision to see precisely what changed and skip some parts of the service if needed.

Do you think it should be done in this PR ? I think yes but it's been a long time since I touched the service and I do not know if it is feasible easily.

drazik:
> > The only solution I have in mind is to compare the document and its previous revision to see precisely what changed and skip some parts of the service if needed.
> 
> Do you think it should be done in this PR ? I think yes but it's been a long time since I touched the service and I do not know if it is feasible easily.

I also think this PR shouldn't be merged without addressing this issue. But I don't think the solution I gave is good enough: the service will still be executed and use resources for nothing.

nono:
> Yes it is a problem, I wonder if it would be a good idea to deactivate modifications from a service
to trigger other services. I am a bit worried about infinite loops of triggers between services. I think we need to keep this in mind, have you already thought about that @nono ?

Yes, I'm also worried that triggers can easily lead to loops. But my worries are not limited to services, it also include sharings and interactions with the desktop client. I don't have good answers for that.

> It reminds me of the trick you did on the account deletion not to trigger the service twice (I think it was with a manual attribute ?

Yes, it is a similar issue, but in my case, it was the stack that could trigger the service twice, not the service that would trigger its-self.

> Another solution when the services are splitted would be to support the trigger of a service only on modification of some attributes.

We have the notion of selector for triggers, but it is a "static" notion: does the new document matches the selector or not? And here, it would be a "dynamic" notion: does the diff between the previous version of the document and the new version include this field? Maybe, it is something we can add without too much work on the stack, but I need to think about it to say clearly what it means.

drazik:
The new feature of the stack is working (see https://github.com/cozy/cozy-stack/pull/2049). But since we can't hide the manifest update behind a flag, I think it is better to wait for this stack feature to be deployed on production environement before merging this PR.

PR #1530, opened 2019-08-28
Check locales

ptbrowne:
I modified greg scripts to sort keys and use git to have colored diff.

Now travis will fail if we have differences betweeen keys in en.json or fr.json.


ptbrowne:
Yes.

ptbrowne:
OK I think I went a bit too far with the last 3 commits, I honestly though that I would detect more unused keys.

PR #1551, opened 2019-09-04
Use COZY_TIME_LIMIT env variable in categorization service instead of hard coded time

drazik:
While doing a totally unrelated exploration on services, I found that they receive a`COZY_TIME_LIMIT` environment variable that can be used in the `categorization` service instead of the hard coded time limit we had.

### DIFF
--- a/src/ducks/categorization/services.js
+++ b/src/ducks/categorization/services.js
@@ -30,7 +30,7 @@ export const fetchTransactionsToCategorize = () => {
 /**
  * Fetch transactions to categorize and chunk them
  *
- * @return {Object[][]} The chunks to categorize
+ * @return {io.cozy.bank.operations[][]} The chunks to categorize
  */
 export const fetchChunksToCategorize = async () => {
   const toCategorize = await fetchTransactionsToCategorize()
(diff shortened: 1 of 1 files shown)


src/ducks/categorization/services.js
```diff
@@ -12,16 +12,26 @@ BankTransaction.registerClient(cozyClient)
...
 export const fetchTransactionsToCategorize = () => {
   return BankTransaction.queryAll({
     toCategorize: true
   })
 }
 
+/**
+ * Fetch transactions to categorize and chunk them
+ *
+ * @return {Object[][]} The chunks to categorize
```
ptbrowne:
```suggestion
 * @return {Transaction[][]} The chunks to categorize
```

?

src/ducks/categorization/services.js
```diff
@@ -30,6 +40,14 @@ export const fetchChunksToCategorize = async () => {
   return chunks
 }
 
+/**
+ * Apply global and local categorization models to a chunk
+ *
+ * @param {Object} categorizer - A categorizer (see https://docs.cozy.io/en/cozy-konnector-libs/api/#categorization)
+ * @param {Object[]} chunk - An array of io.cozy.bank.operations to categorize
```
ptbrowne:
```suggestion
 * @param {io.cozy.bank.operations[]} chunk - Operations to categorize
```

?

ptbrowne:
👍 Thanks for the docstrings !

PR #1552, opened 2019-09-04
feat(onOperationOrBillCreate): Get options from env and CLI

drazik:
With this we can now pass arguments to the `onOperationOrBillCreate` service. For example by doing `cozy-stack --domain default.cozy.tools:8080 jobs run service --json '{"slug": "banks", "name": "onOperationOrBillCreate", "file": "onOperationOrBillCreate.js", "arguments": {"transactionsMatching": false, "billsMatching": { "lastSeq": "0"  } }}'` we can run the service and make it try to match all bills but bypass the matching from transactions.

### DIFF
--- a/src/targets/services/helpers.js
+++ b/src/targets/services/helpers.js
@@ -1,0 +1,57 @@
+import last from 'lodash/last'
+import get from 'lodash/get'
+
+/**
+ * If lastSeq is 0, it's more efficient to fetch all documents.
+ */
+export const fetchChangesOrAll = async (Model, lastSeq) => {
+  if (lastSeq === '0') {
+    const documents = await Model.fetchAll()
+    // fetch last change to have the last_seq
+    const lastChanges = await Model.fetchChanges('', {
+      descending: true,
+      limit: 1
+    })
+    return { documents, newLastSeq: lastChanges.newLastSeq }
+  } else {
+    return Model.fetchChanges(lastSeq)
+  }
+}
+
+const getJobIdFromEnv = env => {
+  const cozyJobId = env.COZY_JOB_ID
+  const jobId = last(cozyJobId.split('/'))
+
+  return jobId
+}
+
+const getOptionsFromEnv = async (client, env) => {
+  const jobId = getJobIdFromEnv(env)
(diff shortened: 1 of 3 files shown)


src/targets/services/helpers.js
```diff
@@ -0,0 +1,57 @@
...
+import get from 'lodash/get'
+
+/**
+ * If lastSeq is 0, it's more efficient to fetch all documents.
+ */
+export const fetchChangesOrAll = async (Model, lastSeq) => {
+  if (lastSeq === '0') {
+    const documents = await Model.fetchAll()
+    // fetch last change to have the last_seq
+    const lastChanges = await Model.fetchChanges('', {
```
nono: [nit] It would be better to first fetch the `last_seq`, and only after that to get all the documents. It avoids a race condition where a document is created between the two calls, and will never be analyzed for matching.

drazik: Thanks @nono, I did the change. See https://github.com/cozy/cozy-banks/pull/1552/commits/b2f73f29882311bb73ad95e6bd3c3e151c8ea28c

src/targets/services/helpers.js
```diff
@@ -0,0 +1,71 @@
+import last from 'lodash/last'
+import get from 'lodash/get'
+
+/**
+ * Returns the changes or all documents if it's more efficient
```
ptbrowne: What does efficient mean here ? Based on what ?

ptbrowne: I think this could be moved to cozy-doctypes right ? Maybe not in this PR but a TODO could be good.

drazik: I updated the docstring.

ptbrowne:
Ping, can we merge this ?

src/targets/services/helpers.js
ptbrowne: Thanks for updating the docstring. I wonder if it could be moved to cozy-stack-client::DocumentCollection, do you think it would have its place there ?

drazik: It definitely has its place there. I'm going to open a PR.

drazik:
I merge it. I thought it was already merged actually. Thank you.

PR #1556, opened 2019-09-05
Perf: Improvements of Balance perf at mount

ptbrowne:
- Delay rendering of GroupPanels (biggest performance gain)

I could not see why GroupPanels where so long to render. I plan to put here more benchmarks. Stay tuned...

Visible on https://testbanksperf-banks.cozy.works/#/balances

### DIFF
--- a/src/components/Chart/LineChart.jsx
+++ b/src/components/Chart/LineChart.jsx
@@ -381,9 +381,14 @@ class LineChart extends Component {
       this.mask.attr('opacity', 0)
     }
 
+    // We do not know where exactly this comes from but without it a bit of the end
+    // of the line is visible just before the animation as if the line was too offsetted.
+    // This is why we need to remove a bit of the offset.
+    const OFFSET = -60
+
     this.line
       .attr('stroke-dasharray', lineTotalLength)
-      .attr('stroke-dashoffset', lineTotalLength)
+      .attr('stroke-dashoffset', lineTotalLength + OFFSET)
       .transition()
       .duration(this.props.enterAnimationDuration)
       .ease(d3.easeExpInOut)
--- a/src/ducks/balance/BalancePanels.jsx
+++ b/src/ducks/balance/BalancePanels.jsx
@@ -8,6 +8,7 @@ import { withRouter } from 'react-router'
 import AddAccountLink from 'ducks/settings/AddAccountLink'
 import { translateAndSortGroups } from 'ducks/groups/helpers'
 import styles from 'ducks/balance/BalancePanels.styl'
+import Delayed from 'components/Delayed'
 
 class BalancePanels extends React.PureComponent {
   static propTypes = {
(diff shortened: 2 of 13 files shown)


src/ducks/balance/Balance.jsx
```diff
@@ -432,6 +432,8 @@ class Balance extends PureComponent {
   }
 }
 
+const Balance = React.memo(_Balance)
```
Crash--: Is there really a gain to memo a PureComponent ?

ptbrowne: Je vais sortir des benchmarks, je pensais que ca ne pouvait pas faire de mal. Mais en fait React.Memo est fait pour les composants à base de fonctions... Je vais revoir ma copie pour voir si ca a vraiment du sens.

src/components/Delayed.jsx
```diff
@@ -2,15 +2,15 @@ import { useState, useEffect } from 'react'
 
 const useDelay = delay => {
   // ok is true when delay has elapsed
-  const [ok, setOK] = useState(false)
+  const [ok, setOK] = useState(delay === 0 ? true : false)
```
drazik: Nit: `useState(delay === 0)` should be sufficient

src/ducks/balance/BalancePanels.jsx
```diff
@@ -43,18 +44,20 @@ class BalancePanels extends React.PureComponent {
...
-            warningLimit={warningLimit}
-            expanded={panelsState[group._id].expanded}
-            checked={panelsState[group._id].checked}
-            switches={panelsState[group._id].accounts}
-            onSwitchChange={onSwitchChange}
-            onChange={onPanelChange}
-            withBalance={withBalance}
-          />
+        {groupsSorted.map((group, i) => (
+          <Delayed key={group._id} delay={i < 2 ? 0 : i * 300}>
```
drazik: This is very strict to state that the third panel and next ones will by below the fold. This will not be the case in a lot of situations (folded panels, panels with few items, big screens).

I'm afraid it will introduce a weird effect by rendering the first two instantly, then adding next ones after some time. I will test it.

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -131,11 +131,11 @@ class AccountRow extends React.PureComponent {
             totalClassName={styles.AccountRow__figure}
             currencyClassName={styles.AccountRow__figure}
           />
+          {/* color: Do not deactivate interactions with the button,
+            only color it to look disabled */}
```
drazik: Where was syntax color broken?

src/ducks/balance/components/BalanceHeader.jsx
```diff
@@ -39,7 +40,9 @@ const BalanceHeader = ({
         onClickBalance={onClickBalance}
       />
       {accounts && (
-        <History accounts={accounts} transactions={transactionsCollection} />
+        <Delayed delay={500}>
+          <History accounts={accounts} transactions={transactionsCollection} />
+        </Delayed>
```
drazik: I more 100% sure it's related, but when testing, I get the bug that was previously resolved by https://github.com/cozy/cozy-banks/pull/1510. Maybe the fix is not strong enough :(

src/ducks/balance/BalancePanels.jsx
drazik: Just tested: I think it's weird to have the first two rendering instantly, then the others appearing later :thinking:

src/ducks/balance/components/AccountRow.jsx
ptbrowne: In sublime text.

drazik: Oh ok, I didn't understand since it looked OK in Github :sweat_smile:

src/ducks/balance/BalancePanels.jsx
ptbrowne: Now, there is no separation between the first and the others, an animation makes them appear fluidly.

src/ducks/balance/Balance.jsx
ptbrowne: Finalement les React.memo ne servaient à rien, merci de m'avoir challengé.

src/ducks/balance/components/BalanceHeader.jsx
ptbrowne: I fixed the line bug in https://github.com/cozy/cozy-banks/pull/1556/commits/a318c8383f3649b48abf250aaa150acba8407d6b (but it is hacky...). The tooltip placement is fixed in #1564 (less hacky).

src/components/Chart/LineChart.jsx
```diff
@@ -381,9 +381,14 @@ class LineChart extends Component {
       this.mask.attr('opacity', 0)
     }
 
+    // We do not know where exactly this comes from but without it a bit of the end
+    // of the line is visible just before the animation as if the line was too offsetted.
+    // This is why we need to remove a bit of the offset.
+    const OFFSET = -60
```
drazik: :(

PR #1564, opened 2019-09-10
Tooltip placement

ptbrowne:
Tooltip placement was a bit off only after creation of the graph. It used a
different coordinate system and a different update system as the rest of the
graph since it was rendered in React.

The fix here is to render the tooltip in SVG and rely on the same update system
and coordinates as the rest of the graph.


src/components/Chart/LineChart.jsx
```diff
@@ -3,7 +3,11 @@ import PropTypes from 'prop-types'
 import * as d3 from 'utils/d3'
 import { max, minBy, keyBy, memoize } from 'lodash'
 import styles from 'components/Chart/LineChart.styl'
-import Tooltip from 'components/Chart/Tooltip'
+
+const getCSSPropertyValue = memoize(varName => {
```
drazik: We already have a [`getCssVariableValue`](https://github.com/cozy/cozy-ui/blob/master/react/utils/color.js#L1) function in cozy-ui. But it's not memoized, it can be a good improvement.

PR #1573, opened 2019-09-17
Display connection problems in Banks

ptbrowne:
When one of the connection has a problem, we tell the user inside Banks so that they can fix that in the Home.

- Display connection status in Balance page
- Display more information and a button to the Home in Transaction page
  - The connection problem banner is displayed in function of the account filter; if we are seeing a revolut account, we will not have a banner signalling the the caisse d'epargne account has a problem
  - Since it is possible that groups contain one or more accounts with problems, we can display more than 1 error at a time. To reduce visual clutter, we display the errors in a carrousel.

image

image

### DIFF
--- a/src/components/Carrousel.jsx
+++ b/src/components/Carrousel.jsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import compose from 'lodash/flowRight'
+import SwipeableViews from 'react-swipeable-views'
+import { Media, Bd, Img } from 'cozy-ui/transpiled/react/Media'
+import { withBreakpoints } from 'cozy-ui/transpiled/react'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import range from 'lodash/range'
+
+const hidden = {
+  visibility: 'hidden'
+}
+
+class Carrousel extends React.PureComponent {
+  constructor(props, context) {
+    super(props, context)
+    this.state = { currentIndex: 0 }
+    this.handleChange = this.handleChange.bind(this)
+    this.handlePrev = this.handlePrev.bind(this)
+    this.handleNext = this.handleNext.bind(this)
+  }
+
+  handleChange(i) {
+    this.setState({ currentIndex: i })
+  }
+
+  handlePrev() {
+    this.setState({ currentIndex: Math.max(0, this.state.currentIndex - 1) })
+  }
+
+  handleNext() {
+    this.setState({
+      currentIndex: Math.min(
+        this.props.children.length - 1,
+        this.state.currentIndex + 1
+      )
+    })
+  }
+
+  render() {
+    const { currentIndex } = this.state
+    const {
+      children,
+      className,
+      breakpoints: { isMobile }
+    } = this.props
+    return (
+      <Media className={className}>
+        <Img
+          style={currentIndex === 0 ? hidden : null}
+          className="u-c-pointer u-slateGrey"
+          onClick={this.handlePrev}
+        >
+          <Icon className={isMobile ? 'u-mh-half' : 'u-mh-1'} icon="left" />
+        </Img>
+        <Bd>
+          <SwipeableViews animateHeight disabled index={currentIndex}>
+            {React.Children.map(children, (child, i) => {
+              return React.cloneElement(child, { active: i === currentIndex })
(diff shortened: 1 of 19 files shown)


src/components/effects.js
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react'
+import { Intents } from 'cozy-interapp'
+
+export const useRedirectionURL = (client, doctype, options) => {
+  const [fetched, setFetched] = useState(null)
+  useEffect(() => {
+    const fetch = async () => {
+      const intents = new Intents({ client: client })
+      const fetchedRes = await intents.getRedirectionURL(doctype, options)
```
drazik: If it contains a redirectionURL, can we rename `fetched`, `setFetched` and `fetchedRes` to something like `redirectionURL`, `setRedirectionURL` and `fetchedRedirectionURL`?

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import CozyClient, { queryConnect } from 'cozy-client'
 import { withBreakpoints, translate } from 'cozy-ui/react'
 import Icon from 'cozy-ui/react/Icon'
 import cx from 'classnames'
+import get from 'lodash/get'
```
drazik: This can be grouped with the next line

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -16,32 +18,30 @@ import {
 import styles from 'ducks/balance/components/AccountRow.styl'
 import { HealthReimbursementsIcon } from 'ducks/balance/components/HealthReimbursementsIcon'
 import AccountIcon from 'components/AccountIcon'
+import { triggersConn } from 'doctypes'
+import { isErrored } from 'utils/triggers'
```
drazik: Not part of this PR, but maybe we should extract these utils to cozy-doctypes

src/components/Carrousel.jsx
```diff
@@ -0,0 +1,83 @@
...
+    const {
+      children,
+      className,
+      breakpoints: { isMobile }
+    } = this.props
+    return (
+      <Media className={className}>
+        <Img
+          className={cx('u-c-pointer u-slateGrey', {
+            'u-o-0': currentIndex === 0
```
drazik: This doesn't prevent from clicking on the element. It is not visible but the user can still interact with it. If we want the element to keep its dimensions so nothing jumps, we can add `visibility: hidden` to it

src/components/Carrousel.jsx
```diff
@@ -0,0 +1,83 @@
...
+            {range(children.length).map((x, i) => (
+              <span key={i} onClick={this.handleChange.bind(this, i)}>
+                {i === currentIndex ? '●' : '○'}
+              </span>
+            ))}
+          </div>
+        </Bd>
+        <Img
+          className={cx('u-c-pointer u-slateGrey', {
+            'u-o-0': currentIndex === children.length - 1
```
drazik: Same here with visibility

src/components/ErrorCard/index.jsx
```diff
@@ -13,18 +13,23 @@ const ErrorCard = ({
   content,
   buttonLabel,
   buttonHref,
-  buttonIcon
+  buttonIcon,
+  style
 }) => {
   return (
-    <div className={styles.ErrorCard}>
+    <div className={styles.ErrorCard} style={style}>
```
drazik: This can be done when extracting it to cozy-ui, but I think we should spread props, so the component can receive events, custom classname, etc

ptbrowne: The ErrorCard will be removed in favor of using the Infos component from cozy-ui.

Will be done in https://github.com/cozy/cozy-banks/pull/1574

src/components/effects.js
ptbrowne: Done.

src/ducks/balance/components/AccountRow.jsx
ptbrowne: Done

PR #1615, opened 2019-10-11
fix: Don't put accountType: null on a group that did not have a…

ptbrowne:
- Otherwise virtualGroups would disappear when creating a group
- Added tests

### DIFF
--- a/src/ducks/groups/helpers.spec.js
+++ b/src/ducks/groups/helpers.spec.js
@@ -1,7 +1,8 @@
 import {
   buildAutoGroups,
   getGroupLabel,
-  translateAndSortGroups
+  translateAndSortGroups,
+  renamedGroup
 } from './helpers'
 import { associateDocuments } from 'ducks/client/utils'
 import { ACCOUNT_DOCTYPE } from 'doctypes'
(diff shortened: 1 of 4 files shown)


src/ducks/groups/helpers.spec.js
```diff
@@ -158,3 +159,21 @@ describe('translateAndSortGroups', () => {
...
+      accountType: 'checkings'
+    }
+    expect(renamedGroup(group, 'My super group')).toEqual({
+      label: 'My super group'
+    })
+    expect(renamedGroup(autogroup, 'My super group')).toEqual({
+      label: 'My super group',
+      accountType: null
+    })
+  })
```
drazik: I suggest

```js
describe('when the given group has an accountType', () => {
  it('should not set accountType to null', () => {
    const group = { accountType: 'checkings' }

    expect(renamedGroup(group, 'My super group')).toEqual({
      label: 'My super group',
      accountType: null
    })
  })
})

describe('when the given group does not have an accountType', () => {
  it('should not set accountType to null', () => {
    const group = { label: 'My group' }

    expect(renamedGroup(group, 'My super group')).toEqual({
      label: 'My super group'
    })
  })
})
```

I agree this is more verbose, but I think having everything under a `should work` test case is not descriptive enough.

ptbrowne: ok, thanks for the code I copy pasted it.

PR #1617, opened 2019-10-11
test: Fix TransactionPageErrors tests

drazik:
Since https://github.com/cozy/cozy-banks/commit/77783752c2dcc52e2b572e0babb6da78f4baa5b3, tests were broken because the flag check that was added is falsy in test env.

It makes our build red on master and all new PRs.


ptbrowne:
Thanks !

PR #1627, opened 2019-10-16
feat: Show harvest error as title of error shown in transaction…

ptbrowne:
Used a function from cozy-harvest-lib to interpret the error from the konnector.

Moved TransactionErrors inside Transactions for the scroll to be OK. Since
the scrolling container is Transactions in desktop, if the errors were
displayed outside, they would be fixed

### DIFF
--- a/src/ducks/transactions/TriggerErrorCard.spec.jsx
+++ b/src/ducks/transactions/TriggerErrorCard.spec.jsx
@@ -1,0 +1,32 @@
+/* global mount */
+import React from 'react'
+import { DumbTriggerErrorCard as TriggerErrorCard } from './TriggerErrorCard'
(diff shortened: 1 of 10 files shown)


src/ducks/transactions/TriggerErrorCard.spec.jsx
```diff
@@ -0,0 +1,32 @@
+/* global mount */
```
drazik: Why don't you get it from enzyme?

ptbrowne: I think it's everywhere in Banks. mount is global. It is not very useful since we have to put this comment... I'll make a pass on test files to replace with an import.

PR #1630, opened 2019-10-18
fix: When LineChart data changes, the selected item key is the…

ptbrowne:
Previously the logic was, if the itemKey cannot be found in the current data,
let's update it. It was not matching what we wanted : on data update, we want
the selected item to be the latest item available.

EDIT: Also fixed a small bug for error in transaction page.


ptbrowne:
No it is not a mistake, I should have edited the description after adding it. Since it was small I put it there also ;)

PR #1632, opened 2019-10-18
fix(Nav): Show selected item on nested routes

drazik:
When we navigated to pages with nested routes (like clicking on "manage my accounts" on the home page to navigate to `/settings/groups`), the selected element of the nav as not updating properly.

I first added the regular expressions for each route. It was not OK when navigating, but it was OK when refreshing or coming on the app on a specific route directly.

Removing the `React.memo` call around `NavLink` fixes the issue while navigating. Actually, a `NavLink` props don't change when navigating to another page without clicking directly on a `NavLink`, so it seems we can't use memo here.

### DIFF
--- a/src/ducks/commons/Nav.jsx
+++ b/src/ducks/commons/Nav.jsx
@@ -28,32 +28,33 @@ const navLinkMatch = props =>
  * to have a faster change of active (not waiting for the route to completely
  * change).
  */
-export const NavLink = React.memo(
-  withRouter(props => {
-    const {
-      children,
-      to,
-      clickState: [lastClicked, setLastClicked]
-    } = props
-    return (
-      <a
-        style={{ outline: 'none' }}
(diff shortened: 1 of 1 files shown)


src/ducks/commons/Nav.jsx
```diff
@@ -28,32 +28,33 @@ const navLinkMatch = props =>
...
+        UINavLink.className,
+        (!lastClicked && navLinkMatch(props)) || lastClicked === to
+          ? UINavLink.activeClassName
+          : null
+      )}
+    >
+      {children}
+    </a>
+  )
+})
```
ptbrowne: OK Can you just try to put the react memo *under* the withRouter ? If it works, good, otherwise LGTM :)

drazik: It doesn't work :(

PR #1643, opened 2019-10-28
Fix KonnectorAction CTA

ptbrowne:
Following the selector refactor, withAppUrls was broken, this prevented
KonnectorAction to correctly work.


src/ducks/settings/AccountsSettings.jsx
```diff
@@ -130,7 +130,7 @@ class AccountsSettings extends Component {
 }
 
 const mapStateToProps = state => ({
-  collectUrl: getAppUrlById(state, 'io.cozy.apps/collect')
+  collectUrl: getHomeURL(state)
```
drazik: Does it require lot of changes to rename this `collectUrl` to `homeUrl`?

src/ducks/apps/selectors.js
```diff
@@ -2,6 +2,8 @@ import keyBy from 'lodash/keyBy'
 import { createSelector } from 'reselect'
 import { queryDataSelector } from 'selectors'
 
+export const getAppURL = app => (app ? app.links.related : null)
```
drazik: We should be more safe with `app => (app && app.links ? app.links.related : null)`

src/ducks/settings/AccountsSettings.jsx
ptbrowne: This component did not use collectUrl anymore :tada:

src/ducks/apps/selectors.js
ptbrowne: Done thanks

PR #1646, opened 2019-11-04
UI tweaks

ptbrowne:
feat: Prevent layout computation during render

feat: Active style for mobile transaction row

fix: Add title to Modal to remove a11y warning

fix: Remove margin-bottom on konnector chip

### DIFF
--- a/src/ducks/transactions/scroll/InfiniteScroll.jsx
+++ b/src/ducks/transactions/scroll/InfiniteScroll.jsx
@@ -17,7 +17,11 @@ function elementInViewport(el, thresold) {
 class InfiniteScroll extends React.Component {
   constructor(props) {
     super(props)
-    this.checkForLimits = throttle(this.checkForLimits.bind(this), 100)
+    // We deactivate the leading call to avoid layout trashing during
+    // componentDidMount/componentDidUpdate
+    this.checkForLimits = throttle(this.checkForLimits.bind(this), 100, {
+      leading: false
+    })
     this.onWindowResize = debounce(this.onWindowResize.bind(this), 500)
   }
 
--- a/src/components/KonnectorChip/KonnectorChip.jsx
+++ b/src/components/KonnectorChip/KonnectorChip.jsx
@@ -8,7 +8,13 @@ const DumbKonnectorChip = props => {
   const { t, konnectorType, ...rest } = props
 
   return (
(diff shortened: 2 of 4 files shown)


src/ducks/transactions/scroll/InfiniteScroll.jsx
```diff
@@ -24,7 +24,7 @@ class InfiniteScroll extends React.Component {
   componentDidMount() {
     this.scrollingElement = this.getScrollingElement()
     this.listenToScroll()
-    this.checkForLimits()
+    setTimeout(() => this.checkForLimits(), 0)
```
Crash--: I didn't take a look at the `onReachTop` props for exemple, but are you doing some animation? Using `requestAnimationFrame` is not enough to put an event in the loop right after the render execution?

src/ducks/transactions/TransactionRow.jsx
```diff
@@ -149,7 +149,7 @@ class _RowMobile extends React.PureComponent {
     }
 
     return (
-      <List.Row onRef={onRef} {...rowRest}>
+      <List.Row onRef={onRef} {...rowRest} onClick={this.handleSelect}>
```
drazik: It doesn't cause any problem to open the categories modal when clicking on the category icon?

src/ducks/transactions/scroll/InfiniteScroll.jsx
drazik: Maybe add a comment to explain why we use a `setTimeout` here? I'm afraid in a few months we don't remember why.

ptbrowne: Yes, I added a comment, thanks !

Also, I went with another approach following crash's comment. Since checkForLimits is invoked in cDU *and* cDM, I changed the throttle options to deactivate the *leading* call, which prevents the call to happen during cDU *and* cDM.

ptbrowne: I went another way eventually (see https://github.com/cozy/cozy-banks/pull/1646#discussion_r341947599), but yes, requestAnimationFrame could also work.

src/ducks/transactions/TransactionRow.jsx
ptbrowne: I removed the commit, I should not have commited this, it is too risky.

PR #1650, opened 2019-11-06
feat: Budget category alerts settings UI

ptbrowne:
- 🖱 UI is functional : a new section in the settings allows to configure
an budget alert for a category. Alerts are shown as clickable cards that
open a modal to configure the alert.

- ✏️ Saves alerts into the io.cozy.bank.settings document under the `categoryBudgetAlerts` key

- 🚩 Section inside settings only shown if `settings.budget-category-alerts` flag is set.

Alert schema:

```
{
  accountOrGroup: {
    _id: 'c0ffeedeadbeef',
    _type: 'io.cozy.bank.accounts'
  }, // can be null if the alert concerns all accounts
  balanceMinThresold: 100,
  category: '400100'
}
```


src/ducks/settings/CategoryAlerts/CategoryAlertCard.jsx
```diff
@@ -0,0 +1,121 @@
...
+        />
+      ) : null}
+      {confirmingRemoval ? (
+        <Modal
+          primaryText={t('Settings.budget-category-alerts.remove.ok')}
+          primaryType="danger"
+          primaryAction={handleRemoveAlert}
+          secondaryText={t('Settings.budget-category-alerts.remove.cancel')}
+          secondaryAction={() => setConfirmingRemoval(false)}
+          dismissAction={() => setConfirmingRemoval(false)}
```
drazik: Nit: extract `setConfirmingRemoval(false)` to a function like `hideRemovalConfirm`

src/ducks/settings/CategoryAlerts/CategoryAlertSettingsPane.jsx
```diff
@@ -0,0 +1,158 @@
...
+    : [...arr, item]
+}
+
+const maxValue = (arr, fn) => {
+  const rs = arr.reduce((curMax, item) => {
+    const v = fn(item)
+    return typeof v === 'number' ? Math.max(v, curMax) : curMax
+  }, 0)
+  return rs
+}
```
drazik: Isn't Lodash's maxBy doing the same job?

src/ducks/settings/CategoryAlerts/CategoryAlertSettingsPane.jsx
```diff
@@ -0,0 +1,158 @@
...
+      <TogglePaneTitle>
+        {t('Settings.budget-category-alerts.pane-title')}
+      </TogglePaneTitle>
+      <TogglePaneText>
+        {t('Settings.budget-category-alerts.pane-description')}
+      </TogglePaneText>
+      <Stack spacing="s">
+        <div>
+          <Grid>
+            {!alerts
```
drazik: Nit: When possible, I prefer positive conditions

src/ducks/settings/CategoryAlerts/CategoryAlertCard.jsx
ptbrowne: Why ? Do you think the name is not clear enough ?

src/ducks/settings/CategoryAlerts/CategoryAlertSettingsPane.jsx
ptbrowne: maxBy will return the item, not the value. But I agree, it is simpler to use maxBy, I changed the code to use it.

PR #1653, opened 2019-11-06
fix: Android targetSDKVersion

Crash--:
Since the 1st November, android app needs to target at least SDK 28. We can update our cordova-android, but it's easier and safer to only specify this target in the config.


ptbrowne:
Thanks !

PR #1655, opened 2019-11-06
Loan details page

drazik:
This introduces a specific page to present the details of a loan account.

The flow is the following:

* If the user selects a loan account: the loan details page is shown
* If the user selects a group composed of loan accounts only: we show a list of the accounts that compose the group and the user can select one, then see the details
* If the user selects a group composed of loan and other account types: we show the traditional transactions page

I refactored the current routing logic using a `BalanceDetailsPage` that implements this logic and determine which page component we should show to the user when he navigates to `/balance/details` according to the currently filtering document.


src/ducks/account/AccountDetailsPage.jsx
```diff
@@ -0,0 +1,47 @@
...
+      return LoanDetailsPage
+    } else if (isReimbursementsVirtualGroup(filteringDoc)) {
+      return ReimbursementsPage
+    } else {
+      return TransactionsPageWithBackButton
+    }
+  } else {
+    return TransactionsPageWithBackButton
+  }
+}
```
ptbrowne: Nice to have all the business logic nicely tucked here.

src/ducks/balance/components/GroupPanel.jsx
```diff
@@ -84,14 +84,7 @@ class GroupPanel extends React.PureComponent {
...
     const { group, filterByDoc, router } = this.props
     filterByDoc(group)
-
-    const isReimbursementsVirtualGroup =
-      group._id === 'Reimbursements' && group.virtual
-    const route = isReimbursementsVirtualGroup
-      ? '/balances/reimbursements'
-      : '/balances/details'
-
-    router.push(route)
```
ptbrowne: Top

src/ducks/account/AccountDetailsPage.jsx
```diff
@@ -6,20 +6,21 @@ import LoanDetailsPage from 'ducks/account/LoanDetailsPage'
...
+    if (accountType === 'Loan' && flag('loan-details-page')) {
       return LoanDetailsPage
     } else if (accountType === 'Reimbursements') {
       return ReimbursementsPage
     } else {
       return TransactionsPageWithBackButton
     }
   } else if (filteringDoc._type === GROUP_DOCTYPE) {
-    if (isLoanGroup(filteringDoc)) {
+    if (isLoanGroup(filteringDoc) && flag('loan-details-page')) {
```
ptbrowne: The rationale here being that if the flag is not active, we will show the TransactionsPage which will show "No transactions here" right ?

drazik: Exactly

src/ducks/transactions/TransactionsPage.spec.jsx
```diff
@@ -91,25 +93,6 @@ describe('TransactionsPage', () => {
...
-    expect(root.find(TransactionsPageBar).length).toBe(1)
-  })
-
-  it('should not show the top balance on movements page on mobile if inside category', () => {
-    useMobile()
-    categoryName = 'dailyLife'
-    subcategoryName = 'supermarket'
-    setup()
-    expect(root.find(TransactionsPageBar).length).toBe(0)
-  })
```
ptbrowne: It seems we are losing those tests ?

src/ducks/account/LoanDetailsPage.jsx
```diff
@@ -1,11 +1,260 @@
...
+        value={get(account, 'loan.lastPaymentAmount')}
+        caption={'le ' + f(get(account, 'loan.lastPaymentDate'), DATE_FORMAT)}
+      />
+      <Row
+        type="amount"
+        title={t('Prochain paiement prévu')}
+        value={get(account, 'loan.nextPaymentAmount')}
+        caption={
+          t('le ') + f(get(account, 'loan.nextPaymentDate'), DATE_FORMAT)
+        }
```
ptbrowne: Careful you seem have untranslated chains here. In my latest works, I used {/* i18n /*} comments when I was building the page, not to forget to remove them afterwards.

src/ducks/account/LoanDetailsPage.jsx
```diff
@@ -1,11 +1,260 @@
...
+  )
+})
+
+const Section = props => {
+  const { title, children } = props
+
+  return <BaseSection title={title}>{children}</BaseSection>
+}
+
+const hasInfos = (account, paths) => {
```
ptbrowne: At first I thought you need all the infos to return true. What do you think of renaming the function to hasSomeInfos to make it clearer that only 1 piece of info is necessary.

ptbrowne: `return _.some(paths.map(path => get(account, path))` ;)

src/ducks/account/LoanDetailsPage.jsx
```diff
@@ -1,11 +1,260 @@
...
+      <Row
+        type="nb-payments"
+        title={t("Nb d'échéances encore prévues")}
+        value={get(account, 'loan.nbPaymentsLeft')}
+      />
+      <Row
+        type="nb-payments"
+        title={t("Nb d'échéances payées")}
+        value={get(account, 'loan.nbPaymentsDone')}
+      />
```
ptbrowne: Untranslated, I guess you know but I'd be more confortable with a big fat comment "// TODO i18n" ;)

src/ducks/account/LoanPercentageLine.jsx
```diff
@@ -0,0 +1,18 @@
...
+  const { className, ...rest } = props
+
+  return (
+    <PercentageLine
+      className={cx(styles.LoanPercentageLine, className)}
+      color="var(--emerald)"
+      {...rest}
+    />
+  )
+}
```
ptbrowne: Should be replaced by a PercentageBar from cozy-ui right ? Can you add a comment in this case ?

src/ducks/account/LoanProgress.jsx
```diff
@@ -0,0 +1,48 @@
...
+  return usedAmount - remaining
+}
+
+const getReimbursedPercentage = account => {
+  const reimbursedAmount = getReimbursedAmount(account)
+  const totalAmount = get(account, 'loan.usedAmount')
+  const percentage = (reimbursedAmount / totalAmount) * 100
+
+  return percentage
+}
```
ptbrowne: I wonder if these utility methods should be move to accounts/helpers.js. This would make
them easier to use in services for example. Also it keeps the separation between models and view clear.

src/ducks/loan/LoanListPage.jsx
```diff
@@ -0,0 +1,42 @@
...
+import CompositeRow from 'cozy-ui/transpiled/react/CompositeRow'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import { Bold } from 'cozy-ui/transpiled/react/Text'
+import AccountIcon from 'components/AccountIcon'
+import withFilters from 'components/withFilters'
+import AccountDetailsHeader from 'ducks/account/AccountDetailsHeader'
+
+const DumbLoanListPage = props => {
+  const { filteringDoc, filterByDoc } = props
+  const accounts = filteringDoc.accounts.data
```
ptbrowne: At this point we need to be sure that we have a filteringDoc, do you think defensive code against the case that filteringDoc is null would be necessary ?

src/components/AppRoute.jsx
```diff
@@ -30,7 +29,7 @@ const AppRoute = () => (
         <IndexRoute component={scrollToTopOnMount(Balance)} />
         <Route
           path="details"
-          component={scrollToTopOnMount(AccountDetailsPage)}
+          component={scrollToTopOnMount(BalanceDetailsPage)}
```
ptbrowne: Nit: Mmh I think AccountDetailsPage was pretty good actually, maybe you renamed it because it can concern a group ? In this case I think AccountOrGroupDetailsPage would be better than BalanceDetailsPage.

src/components/Spacing/Wrapper.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import cx from 'classnames'
+import styles from 'components/Spacing/Wrapper.styl'
+
+const Wrapper = props => {
```
ptbrowne: Nit: A comment explaining what this comment is about would be appreciated :)

src/ducks/loan/LoanDetailsPage.jsx
```diff
@@ -10,10 +11,11 @@ import cx from 'classnames'
 import styles from 'ducks/loan/LoanDetailsPage.styl'
 import LoanProgress from 'ducks/loan/LoanProgress'
 import { Padded, Wrapper } from 'components/Spacing'
+import { flowRight as compose } from 'lodash'
 
 const DATE_FORMAT = 'DD/MM/YY'
 
-const Row = translate()(props => {
+const DumbRow = translate()(props => {
```
ptbrowne: I always use _ as suffix when my component is not yet enhanced but Dumb is better 👍 

Careful: It seems DumbRow get `translate()`ed twice ;)

src/ducks/loan/LoanListPage.jsx
```diff
@@ -19,20 +33,22 @@ const DumbLoanListPage = props => {
...
-                </div>
-              }
-              right={<Icon icon="right" color="var(--coolGrey)" />}
-              actions={<LoanProgress account={account} />}
-              className="u-c-pointer"
-              onClick={() => {
-                filterByDoc(account)
-              }}
-            />
+            <PaddedOnDesktop>
```
ptbrowne: Mmh, I am getting confused, I thought the Wrapper was supposed to do the padding, but it is late. Anyway, Wrapper is a bit too generic :D

src/locales/en.json
```diff
@@ -683,5 +683,36 @@
...
+      "title": "Payments",
+      "lastPayment": "Last payment made:",
+      "nextPayment": "Next scheduled payment:"
+    },
+    "characteristics": {
+      "title": "Loan characteristics",
+      "subscriptionDate": "Subscription date:",
+      "maturityDate": "Repayment date:",
+      "nbPaymentsLeft": "Nb payments left:",
+      "nbPaymentsDone": "Nb payments done:"
```
ptbrowne: It seems No is the correct abbreviation for number: https://english.stackexchange.com/questions/43095/what-is-the-correct-abbreviation-for-the-word-numbers. I think # is also used.

src/locales/fr.json
```diff
@@ -695,5 +695,36 @@
...
+  },
+
+  "LoanProgress": {
+    "reimbursedAmount": "Montant remboursé",
+    "borrowedAmount": "Montant emprunté"
+  },
+
+  "LoanDetails": {
+    "keyInfos": {
+      "title": "Informations clés",
```
ptbrowne: Informations clées

ptbrowne:
Lots of small comments/suggestions but overall LGTM. Thanks, it was a pleasure to read ;)

src/ducks/loan/LoanListPage.jsx
drazik: Yes that's a good idea. But what do you think we should do if filteringDoc is null ?

src/components/AppRoute.jsx
drazik: Yes I struggled to decide how to name it and where to put it actually. If we name it AccountOrGroupDetailsPage, it's also difficult to determine in which duck it should go :(

ptbrowne: Ahah, I am not really sure it is a good way to decide for the name but fine by me.

src/ducks/transactions/TransactionsPage.spec.jsx
drazik: Oh yeah I forgot to move it after refacto

src/ducks/loan/LoanListPage.jsx
ptbrowne: Show something like "No accounts" ?

src/components/Spacing/Wrapper.jsx
drazik: Now using NarrowContent from cozy-ui

PR #1657, opened 2019-11-08
Fix loan details page bugs

drazik:
There was some problems with the loan details pages:

* The borrowed amount can be in `usedAmount` or `totalAmount` depending on the loan type. Extracted this in a helper
* The render conditions on the sections were not enough, we had cases in which only the title was displayed or the credit reserve section was rendered when it shouldn't
* The app was crashing when selecting a loan group because the filtering doc was not hydrated so the accounts relationship was not resolved. We now always return hydrated filtering doc


src/ducks/loan/LoanDetailsPage.jsx
```diff
@@ -114,69 +106,79 @@ const KeyInfosSection = translate()(props => {
...
+  const maturityDate = get(account, 'loan.maturityDate')
+  const nbPaymentsLeft = get(account, 'loan.nbPaymentsLeft')
+  const nbPaymentsDone = get(account, 'loan.nbPaymentsDone')
+
+  const shouldRender = [
+    subscriptionDate,
+    maturityDate,
+    nbPaymentsLeft,
+    nbPaymentsDone
+  ].some(isExistingData)
```
ptbrowne: Just checking, is some correctly supported by the browsers we target ?

src/ducks/loan/LoanDetailsPage.spec.jsx
```diff
@@ -0,0 +1,212 @@
...
+        />,
+        {
+          context: {
+            t: key => key
+          },
+          childContextTypes: {
+            t: PropTypes.func
+          }
+        }
+      )
```
ptbrowne: Instead of manually mocking the context, I thin using the I18n component passing a fake t is less brittle

src/ducks/loan/LoanDetailsPage.spec.jsx
```diff
@@ -0,0 +1,212 @@
...
+        <CreditReserveSection account={{ type: 'RevolvingCredit' }} />,
+        {
+          context: {
+            t: key => key
+          },
+          childContextTypes: {
+            t: PropTypes.func
+          }
+        }
+      )
```
ptbrowne: Between each example the majority of lines are duplicated, the setup method pattern could be useful

src/ducks/loan/LoanDetailsPage.jsx
drazik: Yes

PR #1665, opened 2019-11-12
feat(BillChip): Show vendor name

drazik:
We previously show the vendor name only for refunds. Now we show it for every bill CTA

### DIFF
--- a/src/ducks/transactions/actions/AttachedDocsAction/BillChip.jsx
+++ b/src/ducks/transactions/actions/AttachedDocsAction/BillChip.jsx
@@ -33,8 +33,18 @@ export class DumbBillChip extends React.PureComponent {
     return [doctype, id]
   }
 
+  getChipLabel(vendorName) {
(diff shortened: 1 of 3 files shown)


src/ducks/transactions/actions/AttachedDocsAction/BillChip.jsx
```diff
@@ -71,7 +71,9 @@ export class DumbBillChip extends React.PureComponent {
               <Figure total={bill.amount} coloredPositive signed symbol="€" />
             </>
           ) : (
-            t('Transactions.actions.attachedDocs.bill')
+            <>
+              {t('Transactions.actions.attachedDocs.bill')} {vendorName}
```
ptbrowne: Seems the vendorName should be part of the translation string `Blabla %{vendorName}` since in some languages, the vendorName should be inside/upfront of the string.

Nit

PR #1668, opened 2019-11-14
feat: Improve notification title for category budgets

ptbrowne:
- Title of email/push is different depending on whether only 1 alert was
   triggered or several
- Content of push also depends on whether only 1 alert was triggered or
   several

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -120,7 +120,7 @@ With this, the transactions you created will be categorized, then the
 
 ## Developing
 
-You can create manually an app token and launch the built service.
+You can manually create an app token and launch the built service.
 
 ```
 # Watch services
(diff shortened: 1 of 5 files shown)


docs/services.md
```diff
@@ -104,3 +117,22 @@ following permission to your `manifest.konnector`:
 With this, the transactions you created will be categorized, then the
 `onOperationOrBillCreate` service will be launched and do its work.
 
+
+## Developing
+
+You can create manually an app token and launch the built service.
```
y-lohse:
```suggestion
You can manually create an app token and launch the built service.
```

ptbrowne:
> I think I might have used the pluralization feature to accomplish this, but I'm not sure one approach is much better than the other.

I thought about this but I find that it is clearer with two different keys (clearer diffs for example since you do not have very long lines). Also, the "plural" form here takes different i18n options (alertCount vs categoryLabel).

PR #1670, opened 2019-11-15
Fix plugin initialisation

ptbrowne:
Plugin like Push and Sentry were not correctly initialized since the
register method was called instead of the registerPlugin method.
Unfortunately the register method exists on the client and does not
fail in case of wrong argument.

### DIFF
--- a/src/ducks/client/mobile/mobile.spec.js
+++ b/src/ducks/client/mobile/mobile.spec.js
@@ -1,0 +1,10 @@
+import { getClient } from './mobile'
+
+describe('get mobile client', () => {
+  it('should correctly register plugins', () => {
+    global.__APP_VERSION__ = '1.5.0'
+    const client = getClient()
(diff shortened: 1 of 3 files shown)


src/ducks/client/mobile/mobile.spec.js
```diff
@@ -0,0 +1,10 @@
+import { getClient } from './mobile'
+
+describe('get mobile client', () => {
+  it('should correctly register plugins', () => {
```
Crash--: it should correctly have excepted registered plugins ?

ptbrowne: I don't understand your question

ptbrowne: Done via private message, the name of test could be closer to the fact that we do not test that plugin are registered (testing implementation), but we test that they are there at the end. Fixed to "should have plugins correctly instantiated" in another commit.

PR #1674, opened 2019-11-19
Don't erase existing accounts when adding new accounts to an au…

drazik:
When adding accounts to an automatic group, we were erasing the previous
ones with the new ones. So if I already had a Checkings autogroup with one
account, adding a new checkings account would result in the autogroup
containing only the new account.


ptbrowne:
Oups, sorry :/ Could you add a test ? Thanks !

src/ducks/groups/services.spec.js
```diff
@@ -0,0 +1,82 @@
...
+      await createAutoGroups({ client })
+
+      expect(client.save).toHaveBeenCalledWith(
+        expect.objectContaining({
+          _id: 'checkings_auto',
+          _type: GROUP_DOCTYPE,
+          accountType: 'Checkings',
+          accounts: ['a1', 'a2']
+        })
+      )
```
ptbrowne: I think you're testing implementation here.

PR #1675, opened 2019-11-19
Two items selected in CategoryChoice

ptbrowne:
Since the others category is also the parent category, we have
to check at which level we are when reporting if the item is
selected.

Added tests on CategoryChoice.

### DIFF
--- a/src/components/MultiSelect/index.jsx
+++ b/src/components/MultiSelect/index.jsx
@@ -92,7 +92,7 @@ const ItemPropType = PropTypes.shape({
 
 MultiSelect.propTypes = {
   onSelect: PropTypes.func.isRequired,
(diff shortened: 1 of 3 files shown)


src/components/MultiSelect/index.jsx
```diff
@@ -16,7 +16,14 @@ class MultiSelect extends Component {
     }
   }
 
+  componentWillUnmount() {
+    this.unmounted = true
```
drazik: If you remember why we do a setState in a setTimeout, can you add a comment explaining it? I'm 100% sure future me will be like « meh, why? » in front of it :sweat_smile:

ptbrowne: > If you remember why we do a setState in a setTimeout

I don't :/

PR #1677, opened 2019-11-19
feat: Correctly handle parent categories in budget alerts

ptbrowne:
A flaw in the data model (ids for parent categories are the same as the
the one used for "others" subcategories) is now handled through the use
of an attribute on the alert. If the `categoryIsParent` boolean attribute
is true, all categories pertaining to this category will be included in
when checking for expenses. If it is false, it means that the user has
configured the alert with the "others" category of this parent category.

```
{
  categoryId: '400600'
}
// Others: health

{
  categoryId: '400600',
  categoryIsParent: true
}
// Health parent category
```

### DIFF
--- a/src/components/MultiSelect/index.jsx
+++ b/src/components/MultiSelect/index.jsx
@@ -1,7 +1,17 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import Row from 'components/Row'
+import Row, { RowBody } from 'components/Row'
+import { translate, Button } from 'cozy-ui/react'
 
+export const ChooseButton = translate()(({ onClick, t }) => (
+  <Button
+    className="u-m-0 u-ph-half"
+    theme="secondary"
+    label={t('General.choose')}
+    size="small"
+    onClick={onClick}
+  />
+))
 /**
  * Select like component to choose an option among a list of options.
  * Options can have children; selecting an option that has children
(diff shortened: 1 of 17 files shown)


src/components/MultiSelect/index.jsx
```diff
@@ -70,10 +88,16 @@ class MultiSelect extends Component {
...
+              onClick={() => this.handleClickItem(item)}
               hasArrow={item.children && item.children.length > 0}
-            />
+            >
+              <>
+                <RowBody>{item.title}</RowBody>
+                {item.children && canSelectParent ? (
+                  <ChooseButton onClick={() => this.handleSelect(item)} />
+                ) : null}
+              </>
```
drazik: Is the fragment really needed?

src/ducks/budgetAlerts/service.spec.js
```diff
@@ -221,7 +221,7 @@ describe('service', () => {
       )
     })
 
-    it('should have the correct content (multi)', async () => {
+    fit('should have the correct content (multi)', async () => {
```
drazik: s/fit/it

src/ducks/budgetAlerts/CategoryBudgetNotificationView.js
```diff
@@ -4,7 +4,7 @@ import keyBy from 'lodash/keyBy'
 import { ACCOUNT_DOCTYPE, GROUP_DOCTYPE } from 'doctypes'
 import NotificationView from 'ducks/notifications/BaseNotificationView'
 import { getCurrentDate } from 'ducks/notifications/utils'
-import { getParentCategory } from 'ducks/categories/helpers'
+import { getParentCategoryName } from 'ducks/categories/helpers'
```
drazik: It does not seem to be used :thinking:

src/components/MultiSelect/index.jsx
ptbrowne: No :) Removed, thanks

PR #1681, opened 2019-11-20
feat: When a new account is created, link myself to it

drazik:
First step in the account/contact link : when a new account is created, we link it to the « myself » contact if it exists.

### DIFF
--- a/src/ducks/account/services.js
+++ b/src/ducks/account/services.js
@@ -4,6 +4,7 @@ import { ACCOUNT_DOCTYPE, CONTACT_DOCTYPE } from 'doctypes'
 import get from 'lodash/get'
 import set from 'lodash/set'
 import mergeSets from 'utils/mergeSets'
+import { addOwnerToAccount } from './helpers'
 
 const log = logger.namespace('link-myself-to-accounts')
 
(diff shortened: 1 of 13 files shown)


src/ducks/account/services.js
```diff
@@ -0,0 +1,91 @@
...
+    log('info', 'No myself contact found, bailing out.')
+    return
+  }
+
+  for (const account of accountsToProcess) {
+    const currentOwners = get(account, 'relationships.owners.data', [])
+    const newOwners = [
+      ...currentOwners,
+      { _id: myself._id, _type: CONTACT_DOCTYPE }
+    ]
```
ptbrowne: If settings.linkMyselfToAccounts.processedAccounts is malformed (not correctly saved, or if settings is reset), we risk having the owner twice here. I think we should be a little more defensive here and make sure there are only unique _id here.

```
    const newOwners = uniqueBy([
      ...currentOwners,
      { _id: myself._id, _type: CONTACT_DOCTYPE }
    ], x => x._id)
```

BTW, this logic could be extracted into a `addOwnerToAccount` helper.

drazik: Thanks, done :+1:

PR #1682, opened 2019-11-21
UI tweaks for category alerts

ptbrowne:
- Change text of buttons for buttons of modal not to be on two lines on mobile
- Increase click area of remove alert button (area shown below)

image

### DIFF
--- a/src/ducks/balance/components/GroupPanel.jsx
+++ b/src/ducks/balance/components/GroupPanel.jsx
@@ -45,13 +45,7 @@ const GroupPanelSummary = withStyles(() => ({
 
 const GroupPanelExpandIcon = React.memo(function GroupPanelExpandIcon() {
   return (
-    <span className="u-click-xl">
-      <Icon
-        icon="bottom"
-        className={styles.GroupPanelSummary__icon}
-        width={12}
-      />
-    </span>
+    <Icon icon="bottom" className={styles.GroupPanelSummary__icon} width={12} />
   )
 })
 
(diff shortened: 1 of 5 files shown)


src/ducks/balance/components/GroupPanel.jsx
```diff
@@ -45,13 +45,7 @@ const GroupPanelSummary = withStyles(() => ({
 
 const GroupPanelExpandIcon = React.memo(function GroupPanelExpandIcon() {
   return (
-    <span className="u-click-xl">
```
drazik: u-click-xl should be replaced by u-click-200 ?

src/styles/utilities.styl
```diff
@@ -27,11 +27,16 @@
     .u-w-100
         width 100%
 
-    .u-click-xl::after
-      width 130%
-      height 130%
-      top -15%
-      left -15%
+    // TODO Move utility to cozy-ui
+    .u-click-200
```
drazik: I'm not sure u-click-200 is the good name to give to this class. If we want to increase the clickable area, it will not be representative anymore.

src/ducks/balance/components/GroupPanel.jsx
ptbrowne: In fact, I checked and it was useless to have this class here, the clickable area is already big.

src/styles/utilities.styl
ptbrowne: Replaced by `u-expanded-click-area`

PR #1685, opened 2019-11-25
Pin improvements

ptbrowne:
- Do not wait for pin doc to be fetched before showing pin (if already fetched once, use localStorage to store pinDoc)
- While app is mounting, do not flicker the screen without auth if last interaction
was too long ago (use localStorage to store last interaction)

- Clear values stored locally on client's logout

### DIFF
--- a/src/ducks/pin/PinAuth.jsx
+++ b/src/ducks/pin/PinAuth.jsx
@@ -68,11 +68,6 @@ const FingerprintParagraph = ({ t, onSuccess, onError, onCancel }) => (
  * - It automatically confirms when entered password's length is <props.maxLength>
  */
 class PinAuth extends React.Component {
-  static contextTypes = {
-    store: PropTypes.object.isRequired,
-    client: PropTypes.object.isRequired
-  }
-
   constructor(props) {
     super(props)
     this.handleFingerprintSuccess = this.handleFingerprintSuccess.bind(this)
--- a/src/ducks/pin/PinAuth.jsx
+++ b/src/ducks/pin/PinAuth.jsx
@@ -25,7 +25,7 @@ import fingerprint from 'assets/icons/icon-fingerprint.svg'
 const AttemptCount_ = ({ t, current, max }) => {
   return (
     <div className={styles['Pin__error']}>
-      {/* Have an unbreakspace so that the error when appearing does not // make
+      {/* Have an unbreakable space so that the error, when it appears, does not make
       the previous content jump */}
       {current > 0 ? t('Pin.attempt-count', { current, max }) : '\u00a0'}
     </div>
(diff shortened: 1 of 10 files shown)


src/ducks/pin/PinGuard.jsx
```diff
@@ -21,9 +21,11 @@ class PinGuard extends React.Component {
   initState() {
     const savedLast = lastInteractionStorage.load()
     const last = savedLast || Date.now()
+    const cachedPinSetting = pinSettingStorage.load()
```
drazik: Same for `pingSettingStorage`: where does it come from?

src/ducks/pin/PinAuth.jsx
```diff
@@ -112,9 +112,10 @@ class PinAuth extends React.Component {
     this.logout()
   }
 
-  logout() {
+  async logout() {
     const { client } = this.props
-    client.logout()
+    await client.logout()
+    window.location.reload()
```
drazik: Why do we need to reload the page after client logout?

ptbrowne: Since on desktop the page is not wrapped around authentication (authentication being served by the stack), nothing happens otherwise after logout.

src/ducks/pin/PinGuard.jsx
ptbrowne: Problem in the order of commits, solved now.

src/ducks/pin/PinGuard.jsx
```diff
@@ -12,12 +12,21 @@ import { queryConnect } from 'cozy-client'
...
     super(props)
-    this.state = { last: Date.now() }
+    this.initState()
     this.handleInteraction = this.handleInteraction.bind(this)
     this.handlePinSuccess = this.handlePinSuccess.bind(this)
     this.handleResume = this.handleResume.bind(this)
   }
 
+  initState() {
+    const savedLast = lastInteractionStorage.load()
```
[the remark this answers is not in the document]
ptbrowne: Problem in the order of commit, solved now: https://github.com/cozy/cozy-banks/pull/1685/commits/ffc65e943550db8d5f29c5fddf4bfbe756d8eb94

src/ducks/pin/PinAuth.jsx
```diff
@@ -25,7 +25,7 @@ import fingerprint from 'assets/icons/icon-fingerprint.svg'
 const AttemptCount_ = ({ t, current, max }) => {
   return (
     <div className={styles['Pin__error']}>
-      {/* Have an unbreakspace so that the error when appearing does not // make
+      {/* Have an unbreakable space space so that the error, when it appears, does not make
```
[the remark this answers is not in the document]
ptbrowne: 🦅

PR #1687, opened 2019-11-26
Moved MultiSelect to cozy-ui

ptbrowne:
In https://github.com/cozy/cozy-ui/pull/1270, MultiSelect is moved to
cozy-ui under the name NestedSelect (not to imply that multiple values
can be selected).

During the move, the UI to select a parent was changed. The "Choose
button" appearing on parent rows when `canSelectParent` is set to
`true` is no longer there.

Instead, the parent row is repeated at the top of the children rows.
It is separated from the children rows by a small blank line.

Since the core logic of MultiSelect is now tested in cozy-ui, I removed
the tests from Banks, and instead tested the logic passed to MultiSelect
to determine if a row should look selected (bold text).

Waiting for https://github.com/cozy/cozy-libs/pull/874 for build not be red.


src/ducks/categories/CategoryChoice.spec.jsx
```diff
@@ -24,66 +24,26 @@ describe('getCategoriesOptions', () => {
...
-      expect(selectedRow.length).toBe(1)
-      expect(selectedRow.text()).toBe('Everyday life')
-      selectedRow.simulate('click')
-
-      const selectedRow2 = findSelectedRow(root)
-      expect(selectedRow2.length).toBe(1)
-      expect(selectedRow2.text()).toBe('Dressing')
-    })
+describe('is selected logic', () => {
+  fit('should mark a category as selected if one of its children is selected', () => {
```
drazik: s/fit/it

src/ducks/categories/CategoryChoice.spec.jsx
```diff
@@ -24,66 +24,26 @@ describe('getCategoriesOptions', () => {
...
-      const selectedRow = findSelectedRow(root)
-      expect(selectedRow.length).toBe(1)
-      expect(selectedRow.text()).toBe('Everyday life')
-      selectedRow.simulate('click')
-
-      const selectedRow2 = findSelectedRow(root)
-      expect(selectedRow2.length).toBe(1)
-      expect(selectedRow2.text()).toBe('Others : Everyday life')
-    })
+  it('should not mark a category as selected if it is selected', () => {
```
drazik: The test description is weird. Isn't it « should mark a category as selected if it is selected » ?

PR #1690, opened 2019-11-28
fix: Fingerprint Pin

ptbrowne:
- Correctly pass t to Fingerprint paragraph (otherwise we had a crash)
- Also tweak the size of the fingerprint icon

### DIFF
--- a/src/ducks/pin/PinAuth.jsx
+++ b/src/ducks/pin/PinAuth.jsx
@@ -34,7 +34,7 @@ const AttemptCount_ = ({ t, current, max }) => {
 
 const AttemptCount = translate()(AttemptCount_)
 
-const FingerprintParagraph_ = ({ t, onSuccess, onError, onCancel }) => (
+const DumbFingerprintParagraph = ({ t, onSuccess, onError, onCancel }) => (
   <WithFingerprint
     autoLaunch
     onSuccess={onSuccess}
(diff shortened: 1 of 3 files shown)


src/ducks/pin/PinAuth.jsx
```diff
@@ -34,7 +34,7 @@ const AttemptCount_ = ({ t, current, max }) => {
 
 const AttemptCount = translate()(AttemptCount_)
 
-const FingerprintParagraph = ({ t, onSuccess, onError, onCancel }) => (
+const FingerprintParagraph_ = ({ t, onSuccess, onError, onCancel }) => (
```
drazik:
```suggestion
const DumbFingerprintParagraph = ({ t, onSuccess, onError, onCancel }) => (
```

src/ducks/pin/PinAuth.jsx
```diff
@@ -60,6 +60,8 @@ const FingerprintParagraph = ({ t, onSuccess, onError, onCancel }) => (
   </WithFingerprint>
 )
 
+const FingerprintParagraph = translate()(FingerprintParagraph_)
```
drazik:
```suggestion
const FingerprintParagraph = translate()(DumbFingerprintParagraph)
```

ptbrowne: Done :)

PR #1691, opened 2019-11-28
feat: Restyling rows in transaction modal

ptbrowne:
- Add arrows to application date row and category row
- Put category icon to the left to let space for the arrow
- Needed to add padding to the normal icons since the category icon
  is larger than normal icons

image


src/ducks/transactions/TransactionModal.jsx
```diff
@@ -285,14 +284,25 @@ const TransactionModalInfoContent = withTransaction(props => {
             <Spinner />
           </Img>
         ) : null}
+        {
+          <Img>
+            <RowArrow />
+          </Img>
+        }
```
drazik: It seems braces can be removed

PR #1693, opened 2019-12-02
feat(linkMyselfToAccounts): Use stack's POST /contact/myself

drazik:
This route returns the myself contact if it exists and creates it with
basic infos if it doesn't.
See https://docs.cozy.io/en/cozy-stack/contacts/#post-contactsmyself

So now we get myself contact from this route and we don't have to worry if the contact exists or not. The stack creates it with at least the instance email address if it doesn't exist.

We bail out if the request to the stack to fetch myself contact fails, though.


drazik:
> IMHO it would be good to have the fetchMyself helper inside cozy-client, it woud prevent every application to recode the id/_id trick. What do you think ?

Good idea.

PR #1696, opened 2019-12-03
Show and edit bank accounts owners

drazik:
This adds all features around the bank accounts owners.

On the home page, we show the owners if there are more than 1 differents unique owners accross all bank acounts. On desktop, we add a column on each row. On mobile, I had to update a little the DOM structure because vertical alignment between columns depends on the presence of owners (see screenshots below). (this is behind the `balance.show-owners` flag. It is also possible to use the `balance.force-show-owners` flag to force owners rendering, even if there is only one unique owners).

On the accounts list parameters page, we want to show the owners on each line. (this is behind the `settings.show-accounts-owners` flag).

On the account parameters page, we show a completely new component (behind the `settings.new-account-details-page` flag).

### DIFF
--- a/src/components/Table/styles.styl
+++ b/src/components/Table/styles.styl
@@ -116,4 +116,7 @@
             padding-bottom .5rem
             font-size .75rem
             color var(--coolGrey)
-            max-width fit-content
+            flex-basis auto
+            flex-shrink 0
+            flex-grow 1
+            max-width none
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -1,15 +1,7 @@
 /* global __TARGET__ */
 
 import React, { PureComponent, Fragment } from 'react'
-import {
-  flowRight as compose,
-  get,
-  sumBy,
-  set,
-  debounce,
-  flatten,
-  uniqBy
-} from 'lodash'
+import { flowRight as compose, get, sumBy, set, debounce } from 'lodash'
 
 import { queryConnect, withMutations, withClient } from 'cozy-client'
 import flag from 'cozy-flags'
@@ -36,7 +28,7 @@ import AccountsImporting from 'ducks/balance/components/AccountsImporting'
 
 import { getDefaultedSettingsFromCollection } from 'ducks/settings/helpers'
 import { isCollectionLoading, hasBeenLoaded } from 'ducks/client/utils'
-import { getAccountBalance } from 'ducks/account/helpers'
+import { getAccountBalance, getUniqueOwners } from 'ducks/account/helpers'
 import { isBankTrigger } from 'utils/triggers'
 import { getVirtualAccounts, getVirtualGroups } from 'selectors'
 
@@ -406,12 +398,9 @@ class Balance extends PureComponent {
             nbAccounts: accounts.length
           }
 
-    const allOwners = flatten(
-      accounts.map(account => get(account, 'owners.data', []))
-    )
-    const uniqOwners = uniqBy(allOwners, owner => owner._id)
+    const owners = getUniqueOwners(accounts)
     const showOwners =
-      (uniqOwners.length > 1 && flag('balance.show-owners')) ||
+      (owners.length > 1 && flag('balance.show-owners')) ||
       flag('balance.force-show-owners')
 
     return (
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -1,7 +1,15 @@
 /* global __TARGET__ */
 
 import React, { PureComponent, Fragment } from 'react'
-import { flowRight as compose, get, sumBy, set, debounce } from 'lodash'
+import {
+  flowRight as compose,
+  get,
+  sumBy,
+  set,
+  debounce,
+  flattenDeep,
+  uniqBy
+} from 'lodash'
 
 import { queryConnect, withMutations, withClient } from 'cozy-client'
 import flag from 'cozy-flags'
@@ -398,6 +406,14 @@ class Balance extends PureComponent {
             nbAccounts: accounts.length
           }
 
+    const allOwners = flattenDeep(
+      accounts.map(account => get(account, 'owners.data', []))
+    )
+    const uniqOwners = uniqBy(allOwners, owner => owner._id)
+    const showOwners =
+      (uniqOwners.length > 1 && flag('balance.show-owners')) ||
+      flag('balance.force-show-owners')
+
     return (
       <Fragment>
         <BarTheme theme="primary" />
@@ -418,6 +434,7 @@ class Balance extends PureComponent {
             panelsState={this.state.panels}
             onSwitchChange={this.handleSwitchChange}
             onPanelChange={this.debouncedHandlePanelChange}
+            showOwners={showOwners}
           />
         </Padded>
       </Fragment>
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -7,7 +7,7 @@ import {
   sumBy,
   set,
   debounce,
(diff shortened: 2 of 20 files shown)


src/ducks/settings/NewAccountSettings.styl
```diff
@@ -0,0 +1,46 @@
+@require 'settings/breakpoints'
+
+right_padding = 2.5rem
```
ptbrowne: Where is this value coming from ?

src/components/Table/styles.styl
```diff
@@ -116,4 +116,7 @@
             padding-bottom .5rem
             font-size .75rem
             color var(--coolGrey)
-            max-width fit-content
+            flex-basis auto
+            flex-shrink 0
+            flex-grow 1
+            max-width none
```
ptbrowne: For my culture, what was the problem here with max-width fit-content ?

src/ducks/settings/AccountsSettings.jsx
```diff
@@ -62,21 +74,33 @@ const renderAccount = account => (
...
+          <th className={styles.AcnsStg__number}>{t('Accounts.account')}</th>
+          <th className={styles.AcnsStg__type}>{t('Accounts.type')}</th>
+          <th
+            className={
+              showAccountsOwners
+                ? styles.AcnsStg__owner
+                : styles.AcnsStg__shared
+            }
+          >
+            {t(`Accounts.${showAccountsOwners ? 'owner' : 'shared'}`)}
```
ptbrowne: Nit: if the conditional was on the whole string, it would make it more grep friendly.

src/ducks/settings/AccountsSettings.styl
```diff
@@ -18,10 +18,18 @@
...
         maxed-flex-basis 20%
 
+        +small-screen
+            maxed-flex-basis 33%
+
     &__number
         maxed-flex-basis 15%
         color var(--coolGrey)
 
+        +small-screen
```
ptbrowne: I thought we had to use `+small-screen()` (with parenthesis), are you sur this is working ?

src/ducks/settings/AccountsSettings.styl
```diff
@@ -18,10 +18,18 @@
...
+            td&
+                background-color red
```
ptbrowne: red ? Shouldn' we use pomegranate or something ? I have trouble seeing where we want to use a red background.

src/ducks/balance/Balance.jsx
```diff
@@ -398,6 +406,14 @@ class Balance extends PureComponent {
             nbAccounts: accounts.length
           }
 
+    const allOwners = flattenDeep(
+      accounts.map(account => get(account, 'owners.data', []))
+    )
```
ptbrowne: Why does normal flatten does not work ? I am not very familiar with the owners.data structure. 

Nit: Maybe this could be extracted to a more descriptive helper ?

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -84,9 +86,16 @@ class AccountRow extends React.PureComponent {
...
-      triggerCol
+      triggerCol,
+      showOwners,
+      client
     } = this.props
 
+    const contacts = client.getCollectionFromState(CONTACT_DOCTYPE)
+    const contactsById = keyBy(contacts, contact => contact._id)
+    const ownerRelationships = get(account, 'relationships.owners.data', [])
+    const owners = ownerRelationships.map(data => contactsById[data._id])
```
ptbrowne: Nit: IMHO a selector would be better as it would memoize more easily.

src/ducks/balance/components/AccountRow.styl
```diff
@@ -96,3 +96,6 @@
     .AccountRow__logo
         opacity 0.5
 
+
+.AccountRow__ownersIcon
+    margin-right 0.25rem
```
ptbrowne: Maybe spacing_values could be used here ?

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -57,6 +57,33 @@ const Number = React.memo(function Number({ account }) {
...
+        <UpdatedAt account={account} t={t} />
+      )}
+    </div>
+  )
+}
+
+const UpdatedAtOrFail = compose(
+  translate(),
+  React.memo
+)(DumbUpdatedAtOrFail)
```
ptbrowne: Nice

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -134,7 +164,8 @@ class AccountRow extends React.PureComponent {
           [styles['AccountRow--hasWarning']]: hasWarning,
           [styles['AccountRow--hasAlert']]: hasAlert,
           [styles['AccountRow--disabled']]:
-            (!checked || disabled) && account.loading !== true
+            (!checked || disabled) && account.loading !== true,
+          [styles['AccountRow--withOwners']]: shouldShowOwners
```
ptbrowne: Nit: I am always troubled when I see this kind of code (class of the parent depending on the children), why is the content of the row not expanding automatically the row ?

src/ducks/balance/components/AccountRow.styl
```diff
@@ -96,6 +96,9 @@
     .AccountRow__logo
         opacity 0.5
 
+.AccountRow--withOwners
+    +small-screen()
+        height rem(82)
```
ptbrowne: What is this magic value ?

src/ducks/settings/NewAccountSettings.jsx
```diff
@@ -0,0 +1,304 @@
...
+  const handleSubmit = e => {
+    e.preventDefault()
+    onSubmit({ shortLabel, owners })
+  }
+
+  const fieldVariant = isMobile ? 'default' : 'inline'
+
+  return (
+    <form onSubmit={handleSubmit} {...rest}>
+      <Stack spacing={isMobile ? 's' : 'm'}>
```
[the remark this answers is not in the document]
y-lohse: I also had in mind that s/m/etc values should be responsive (as opposed to 1, 2, etc). Do you remember when we said the Stack should have fixed values?

src/ducks/balance/components/AccountRow.styl
drazik: This value does not exist in `spacing_values`. `xs` is `0.5rem`. But since we have `xxl`, maybe we can have `xxs`?

src/ducks/balance/components/AccountRow.styl
drazik: Fixing https://github.com/cozy/cozy-banks/pull/1696#discussion_r353166734 would also fix this magic value issue. I will try something

src/ducks/settings/NewAccountSettings.styl
drazik: I added a comment to explain this

src/components/Table/styles.styl
drazik: It was not working at all in Firefox. Also, MDN marks this value as not production-ready.

src/ducks/balance/Balance.jsx
drazik: Flatten is enough. I extracted it.

src/ducks/balance/components/AccountRow.jsx
drazik: I added a todo comment

src/ducks/settings/NewAccountSettings.jsx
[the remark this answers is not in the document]
ptbrowne: I think I talked about this IRL with Joel when I showed the Stack to him. I am not really sure. I would be glad to be wrong about this.

PR #1699, opened 2019-12-04
refactor: Remove reimbursement late flag

ptbrowne:
The setting for health reimbursements late limit was "transferred" to flags
   at app initialization. This was a shortcut/hack to easily get the late
   limit in helpers. Now that we can more easily use selectors, it's better to
   use them since they will refresh if the value is changed

### DIFF
--- a/src/ducks/balance/components/GroupPanel.spec.jsx
+++ b/src/ducks/balance/components/GroupPanel.spec.jsx
@@ -116,9 +116,16 @@ describe('GroupPanel', () => {
 })
 
 describe('Reimbursement virtual group styling', () => {
+  // We need a different date for each test otherwise selectors are not
+  // recomputed since queryCreateSelector bases its memoization on
+  // the lastUpdate of the query
+  const mockedDate = +new Date('2020-01-31T00:00')
+  let i = 0
   beforeEach(() => {
-    MockDate.set('2020-01-31T00:00')
+    MockDate.set(mockedDate + i++)
+    getClient.mockReturnValue(client)
   })
+
   const setup = ({
     lateHealthReimbursementNotificationSetting,
     transactions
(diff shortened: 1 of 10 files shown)


src/ducks/transactions/actions/ReimbursementStatusAction/ReimbursementStatusAction.jsx
```diff
@@ -113,7 +122,13 @@ export class DumbReimbursementStatusAction extends React.PureComponent {
 
 const ReimbursementStatusAction = compose(
   translate(),
-  withMutations()
+  withMutations(),
+  connect(state => {
+    const settings = getSettings(state)
+    return {
+      healthReimbursementLateLimit: getHealthReimbursementLateLimit(settings)
```
drazik: Why don't you use the selector `getHealthReimbursementLateLimitSelector` here?

src/ducks/balance/components/GroupPanel.spec.jsx
```diff
@@ -116,9 +116,16 @@ describe('GroupPanel', () => {
...
 
 describe('Reimbursement virtual group styling', () => {
+  // We need a different date for each test otherwise selectors are not
+  // recomputed since queryCreateSelector bases its memoization on
+  // the lastUpdate of the query
+  const mockedDate = +new Date('2020-01-31T00:00')
+  let i = 0
   beforeEach(() => {
-    MockDate.set('2020-01-31T00:00')
+    MockDate.set(mockedDate + i++)
```
drazik: It seems that we will have to add this everytime we have test with cozy-client data and mocked date? I'm afraid we will forgot it :(

ptbrowne: > I'm afraid we will forgot it :(

The test cannot work if we do not do this so I don't think we will forget it, I do not see yet a better solution, I will speak about this in Front meeting.

drazik: Okay

src/ducks/transactions/actions/ReimbursementStatusAction/ReimbursementStatusAction.jsx
ptbrowne: Good catch, fixed

PR #1707, opened 2019-12-09
Update accounts owners showing condition

drazik:
The previous implementation was not right due to a misunderstanding on my
side. We only to show owners that are not the myself contact.

The PR also contains other things related to owners and balance:

* Force `balance.show-owners` flag check so the flag appears in the debug tab of the settings
* Add cozy subdomain type attribute to app root node (used by contact picker to build the contacts app url)
* Remove file that are not used anymore since we don't have references to the old balance page design


src/index.ejs
```diff
@@ -66,6 +66,7 @@
   data-cozy-icon-path="{{.IconPath}}"
   data-cozy-tracking="{{.Tracking}}"
   data-cozy-app-slug="{{.AppSlug}}"
+  data-cozy-subdomain-type="{{.SubDomain}}"
```
ptbrowne: At some point I wonder if we should not only have the key `data-cozy` where all the data from the stack would go.

This would help us not having to add a key everytime we need a new piece of data from the stack. Additionally, the keys must be conventional and coherent between apps to be correctly picked up by cozy-client or other libs relying on those values, so having the app developer write those keys (and potentially mispelling them) is a risk that could be reduced if there was only one key. The interface would only be between cozy-client and the stack.

PR #1708, opened 2019-12-09
feat: Rework settings page

ptbrowne:
Use the same kind of card as the category alert edit modal.


src/locales/fr.json
```diff
@@ -404,13 +404,19 @@
...
       "settingTitle": "Solde bas",
+      "fieldLabels": {
+        "value": "Me prévenir quand mon solde passe sous"
+      },
       "email": {
         "innerTitle": "Notification de solde bas",
         "access_accounts": "Accéder à mes comptes"
       },
-      "description": "Vous recevrez une notification, si votre solde est inférieur à",
+      "description": "Vous recevrez une notification, si votre solde est inférieur à *%{value}*€",
```
drazik: There's no comma in the english version, and I think it's better to read without it.

src/locales/fr.json
```diff
@@ -425,10 +431,13 @@
     },
     "if_transaction_greater": {
       "settingTitle": "Mouvements",
-      "description": "Vous recevrez une notification, si un mouvement est supérieur à",
+      "description": "Vous recevrez une notification, si un mouvement est supérieur à *%{value}*€",
```
drazik: Same here for the comma

src/components/EditionModal/EditionModal.jsx
```diff
@@ -88,20 +94,19 @@ const InfoSlide = ({
...
             value={fieldSpec.getValue(doc)}
-            onClick={
-              fieldSpec.immediate ? null : () => onRequestChooseField(fieldName)
-            }
+            onClick={hasSelector ? () => onRequestChooseField(fieldName) : null}
             onChange={
-              fieldSpec.immediate
-                ? value => {
+              hasSelector
+                ? null
```
drazik: Don't we want to pass `undefined`, to act as if the prop wasn't passed at all? (for example if there is a default prop)

src/components/SettingCard/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+      {...mergeProps(props, {
+        className: cx(
+          styles.SettingCard,
+          dashed && styles['SettingCard--dashed'],
+          clickable && styles['SettingCard--clickable']
+        )
+      })}
+    />
+  )
+}
```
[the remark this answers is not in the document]
ptbrowne: No, you're right it was too complex.

src/components/EditionModal/EditionModal.jsx
```diff
@@ -22,6 +23,16 @@ export const CHOOSING_TYPES = {
...
 
+const BackArrow = ({ onClick }) => (
+  <span className="u-mr-half">
+    <Icon
+      style={{ cursor: 'pointer', color: 'var(--coolGrey)' }}
+      icon="previous"
+      onClick={onClick}
+    />
+  </span>
+)
```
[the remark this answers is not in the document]
ptbrowne: Done :)

PR #1711, opened 2019-12-09
Per account/group notifications

ptbrowne:
Add setting in UI to set account/group for transaction greater and
balance lower. This way you can restrict an alert to a particular
account / group.


src/ducks/client/utils.js
```diff
@@ -81,3 +81,8 @@ export const updateNotificationToken = (client, token) => {
 export const getNotificationToken = client => {
   return client.stackClient.oauthOptions.notification_device_token
 }
+
+export const fullyDehydrateDocument = doc => ({
```
drazik: Why is it named « fully »? It doesn't seem we have a `dehydrateDocument` (or I missed it).

src/ducks/settings/Configuration.jsx
```diff
@@ -82,13 +108,54 @@ const editModalProps = {
...
   })
 }
 
+const getTransactionGreaterDescriptionKey = props => {
+  if (props.doc && props.doc.accountOrGroup) {
+    return 'Notifications.if_transaction_greater.descriptionWithAccountGroup'
+  } else {
+    return 'Notifications.if_transaction_greater.description'
+  }
+}
```
drazik: Since it only uses `props.doc.accountOrGroup`, wouldn't it be better to directly pass `accountOrGroup` to this function? When I see the `props` name, I think the function is a react component. But here it's a regular getter function.

src/ducks/settings/Configuration.jsx
```diff
@@ -82,13 +108,54 @@ const editModalProps = {
...
+
+const getBalanceLowerDescriptionKey = props => {
+  if (props.doc && props.doc.accountOrGroup) {
+    return 'Notifications.if_balance_lower.descriptionWithAccountGroup'
+  } else {
+    return 'Notifications.if_balance_lower.description'
+  }
+}
```
drazik: Same suggestion here for `props.doc.accountOrGroup`

src/ducks/settings/Configuration.jsx
```diff
@@ -82,13 +108,54 @@ const editModalProps = {
...
+    : null,
+  value: props.doc.value
+})
+
+const getBalanceLowerDescriptionProps = props => ({
+  accountOrGroupLabel: props.doc.accountOrGroup
+    ? props.getAccountOrGroupLabelFromDehydratedDoc(props.doc.accountOrGroup)
+    : null,
+  value: props.doc.value
+})
```
drazik: Same suggestion as above for these two

src/ducks/settings/Configuration.jsx
ptbrowne: Good point, I thought it was more flexible to use directly props, but doc is sufficent.

src/ducks/client/utils.js
ptbrowne: Renamed to getDocumentIdentity, thanks I was not particularly happy with the name fullyDehydrateDocument since it can be confused with the dehydratation used for relationships.

src/ducks/settings/Configuration.jsx
ptbrowne: I actually need the full props to have access to getAccountOrGroupLabel from the props.

src/ducks/client/utils.js
ptbrowne: More explanation here : https://github.com/cozy/cozy-banks/pull/1711/commits/c104d08c8d29a3a8e0a04f8fe5fefc9942b9625d

src/ducks/settings/Configuration.jsx
ptbrowne: More explanation here : https://github.com/cozy/cozy-banks/pull/1711/commits/c104d08c8d29a3a8e0a04f8fe5fefc9942b9625d

PR #1722, opened 2019-12-12
Multiple rule support inside service for BalanceLower

ptbrowne:
Prepare to support the UI allowing to add several rules for a single
notification.


src/ducks/notifications/BalanceLower/index.js
```diff
@@ -132,6 +151,10 @@ class BalanceLower extends NotificationView {
     const onlyOne = accounts.length === 1
     const firstAccount = accounts[0]
 
+    const titleKey = `Notifications.if_balance_lower.notification.${
+      onlyOne ? 'one' : 'several'
+    }.title`
```
drazik: I see it has only been moved, but maybe we can also make it pass the grep test :D

src/ducks/notifications/BalanceLower/index.js
```diff
@@ -147,14 +148,19 @@ class BalanceLower extends NotificationView {
...
     const onlyOne = accounts.length === 1
     const firstAccount = accounts[0]
 
     const titleKey = `Notifications.if_balance_lower.notification.${
-      onlyOne ? 'one' : 'several'
+      onlyOne
+        ? 'one'
+        : matchingRules.length === 1
+        ? 'several'
+        : 'several-multi-rule'
```
drazik: That's a good occasion to make it pass the grep test!

ptbrowne: Done,good idea !

PR #1731, opened 2019-12-19
fix: Back button styles were wrong on mobile

ptbrowne:
BackArrow were not applied correctly.

We badly need automated screenshots.

image

image

image

image


ptbrowne:
Good idea, I opened https://github.com/cozy/cozy-banks/issues/1734

PR #1732, opened 2019-12-20
feat: Better add alert button

ptbrowne:
- Smaller
   - Detached from the rest with a margin
   - "Ajouter une alerte" instead of "Créer une alerte" in French


ptbrowne:
The snapshot failure is due to the implementation of Infos. I switched to a shallow test and updated the snapshot.

ptbrowne:
Since some tests were based on the current year, they failed after the new year. I used MockDate to fix this.

PR #1733, opened 2019-12-23
Handle uninstalled contacts app

drazik:
When the user wants to add a contact to a bank account via the « add a contact » button of the ContactPicker, but don't have the contacts app installed, we want to redirect him to the store. That's the job done by https://github.com/cozy/cozy-ui/pull/1326. But for it to work, we have to register a realtime plugin to our client. For now the realtime plugin is implemented here and we will be able to extract it elsewhere (cozy-client? cozy-realtime?)

Todo:

- [x] Update cozy-ui when https://github.com/cozy/cozy-ui/pull/1326 has been merged
- [x] Update cozy-realtime when https://github.com/cozy/cozy-libs/pull/902 has been merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,7 +184,7 @@
     "cozy-logger": "1.6.0",
     "cozy-notifications": "0.3.2",
     "cozy-pouch-link": "6.64.7",
-    "cozy-realtime": "3.2.1",
+    "cozy-realtime": "3.3.0",
     "cozy-stack-client": "8.7.0",
     "cozy-ui": "28.8.0",
     "d3": "5.11.0",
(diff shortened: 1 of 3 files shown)


Crash--:
I think we can extract it to cozy-client :)

drazik:
> I think we can extract it to cozy-client :)

Actually I think it will cause issues. If we put RealtimePlugin in cozy-client, we will want to export it from the entry point. But since there will be a `import CozyRealtime from 'cozy-realtime'` in `plugins/realtime.js`, then an app that just want to `import CozyClient from 'cozy-client'` will try to resolve the import from cozy-realtime, and fail if it can't be found.

A potential solution would be to put the RealtimePlugin in cozy-realtime.

ptbrowne:
> A potential solution would be to put the RealtimePlugin in cozy-realtime.

Seems like a good solution to me.

drazik:
See https://github.com/cozy/cozy-libs/pull/902

drazik:
This is OK now. Reviews welcome :pray:

PR #1735, opened 2020-01-03
Automate alert push notification tests

ptbrowne:
Push notifications E2E tests were semi automatic since we had
to start the mock push notification server and compare the requests
received to what was expected.

Now the tests are automatic:

- Mock push notification server is automatically started during tests
- Push notifications sent to the mock push notification server are
  compared to what's expected
- Tests are no longer run through a custom CLI but through Jest


test/e2e/alerts.js
```diff
@@ -362,38 +353,59 @@ const runScenario = async (client, scenarioId, options) => {
     if (!options.showOutput) {
       console.error(`Re-run with -v to see its output.`)
     }
-    return false
+    throw new Error('Error while running onOperationOrBillCreate')
```
drazik: The error message includes "onOperationOrBillCreate". Shouldn't it be "alerts"?

docs/dev.md
```diff
@@ -442,7 +442,7 @@ be done on the CI in the future.
 It is possible to test push notifications with the `--push` flag. You need first to
 configure the stack so that it uses a fake server that will receive push notifications.
 
-- Add this to `~/.cozy.yml`:
+- Add this to `~/cozy.yml`:
```
drazik: I think the stack config file can vary. For example in [this piece of doc](https://docs.cozy.io/en/howTos/dev/services/#copy-the-default-config-file), it's under `~/.cozy/cozy.yaml` (that's what I personally use). So I think we should not give a specific path here, but just say "Add this to your cozy-stack configuration file".

docs/dev.md
```diff
@@ -416,15 +420,33 @@ touch on the file being launched solves it.
...
+```
+{ NotRegisteredException: Client not registered or missing OAuth information
+...
+```
+
+Removing the token fixes the problem
+
+```
+rm /tmp/cozy-client-oauth-cozy-tools:8080-banks.alerts-e2e.json
+```
```
drazik: Maybe we can add a link to this documentation section when we catch a NotRegisteredException?

jest.config.js
```diff
@@ -1,3 +1,5 @@
+const JEST_PROJECT = process.env.JEST_PROJECT
```
drazik: Is JEST_PROJECT something standard in jest? If it's custom, can we rename it to `TEST_LEVEL`? I find it more easy to understand since it can be related to the test pyramid (unit -> integration -> e2e)

test/e2e/alerts.e2e.spec.js
```diff
@@ -198,55 +169,71 @@ const main = async () => {
...
-  for (const [scenarioId, answer] of Object.entries(answers)) {
-    console.log(
-      answer ? '✅' : '❌',
-      scenarioId,
-      `(${scenarios[scenarioId].description})`
-    )
-  }
+    for (const scenario of Object.values(scenarios)) {
+      describe(scenario.description, () => {
+        it('should work', async () => {
```
drazik: Nit: If there's only one `it` in the scenario, I think we can drop the `describe` and put the scenario description in the `it`. It will remove a nesting level (in test code and in test summary), and also useless "should work".

test/e2e/alerts.e2e.spec.js
```diff
@@ -198,55 +169,71 @@ const main = async () => {
...
-main()
-  .catch(e => {
-    console.error(e)
-    process.exit(1)
-  })
-  .then(() => {
-    process.exit(0)
+    for (const scenario of Object.values(scenarios)) {
+      describe(scenario.description, () => {
+        it('should work', async () => {
```
drazik: Same nit for the scenario description here.

test/e2e/alerts.js
ptbrowne: No here, this is the onOperationOrBillCreate.js service that is launched.

docs/dev.md
ptbrowne: Correct, I fixed and added a link to the doc.

docs/dev.md
ptbrowne: I rm the file now so that this piece of doc is not needed.

test/e2e/alerts.e2e.spec.js
ptbrowne: Fixed, thanks !

jest.config.js
ptbrowne: I was inspired by https://github.com/facebook/jest/issues/7542#issuecomment-523807670, I think the final implementation in Jest will use the project noun.

drazik: OK :+1:

PR #1739, opened 2020-01-03
Warning tests

ptbrowne:
Remove some warnings in tests. We are almost there ! At some point we will be able put

```
console.warn = logAndThrowException
```

Warnings are a useful tool when there are a small number of them but we had so much of them that the ratio signal / noise was too low, making us ignoring them where they can be of good help when refactoring / spotting bad patterns.

### DIFF
--- a/test/jest.setup.js
+++ b/test/jest.setup.js
@@ -1,6 +1,7 @@
 import { configure, mount, shallow } from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
 import logger from 'cozy-logger'
+import { makeDeprecatedLifecycleMatcher, ignoreOnConditions } from './jestUtils'
 
 // To avoid the errors while creating theme (since no CSS stylesheet
 // defining CSS variables is injected during tests)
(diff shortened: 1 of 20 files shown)


test/jest.setup.js
```diff
@@ -25,3 +26,29 @@ global.cozy = {
...
+    reason: 'Preact compatibility',
+    matcher: makeDeprecatedLifecycleMatcher('I18n')
+  },
+  ModalContent: {
+    reason: 'Preact compatibility',
+    matcher: makeDeprecatedLifecycleMatcher('ModalContent')
+  },
+  ModalAriaLabel: {
+    reason: 'Dont know',
+    matcher: msg => msg.includes('aria-label') && msg.includes('If your modal')
```
drazik: I think this would hide real problems: cozy-ui's modal requires to have an aria-label for accessibility reasons. We should add missing aria-labels on our modals.

test/jestUtils.js
```diff
@@ -0,0 +1,19 @@
+const isDeprecatedLifecycleWarning = (msg, componentName) => {
+  return (
+    msg.includes('has been renamed, and is not recommended for use') &&
+    msg.includes(componentName)
```
drazik: Do we know if we have these warnings because we didn't migrate to UNSAFE_ lifecycle methods, or if one of our dependencies didn't do it yet?

ptbrowne: When using react-redux 5's `connect`, the wrapped component will issue this warning. The jest.setup.js details the "reason"s for each ignored warning.

test/jest.setup.js
ptbrowne: OK fair enough, the problem is that the Modal issues this warning even if we have `aria-label`. I am on it. https://github.com/cozy/cozy-ui/issues/1338

test/jestUtils.js
drazik: OK thanks

test/jest.setup.js
drazik: :+1:

PR #1742, opened 2020-01-08
feat: Pin tweaks

ptbrowne:
- Specific look for buttons on iOS
- Small tweaks when handling interactions (+ tests)


src/ducks/pin/styles.styl
```diff
@@ -29,6 +29,7 @@
 
 .Pin__button:active
     background-color rgba(255, 255, 255, 0.25)
+    transition background-color 0s
```
drazik: Nit: did you try to use `transition unset`? If it works, it would be more DRY

ptbrowne: Done, thanks !

PR #1744, opened 2020-01-10
Health reimbursement period

ptbrowne:
- Reimbursements are considered in the last 6 month period in account summary in the home (instead of current year)
- Current period is shown in the empty message

### DIFF
--- a/src/ducks/account/helpers.js
+++ b/src/ducks/account/helpers.js
@@ -13,7 +13,7 @@ import {
   hasPendingReimbursement
 } from 'ducks/transactions/helpers'
 import { isHealthExpense } from 'ducks/categories/helpers'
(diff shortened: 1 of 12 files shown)


src/ducks/account/helpers.js
```diff
@@ -134,12 +134,19 @@ export const getAccountUpdatedAt = account => {
   }
 }
 
+const isLast6Month = () => {
+  const SIX_MONTHS_AGO = new Date(Date.now() - 1000 * 60 * 60 * 24 * 30 * 6)
```
drazik: Nit: It may be clearer using date-fns:

```js
const SIX_MONTHS_AGO = subMonths(new Date(), 6)
```

PR #1746, opened 2020-01-13
test: Add QAWolf test

ptbrowne:
QAWolf is a tool to generate scenarios from a browsing session. I think it could be
a good way for us to generate E2E tests.

- [x] Add QAWolf dependency
- [x] Add first test scenario
- [x] Add screenshots sent to Argos
- [x] Make selectors more generic (not depending on cozy-ui generated classes) or use `data-test-id` attribute

[QAWolf quick start](https://docs.qawolf.com/docs/quick_start)
data-test-id: [Use data-qa attributes](https://docs.qawolf.com/docs/edit_test_code#use-data-attributes) + [QAW attribute](https://docs.qawolf.com/docs/api#qaw_attribute)


.qawolf/banks.test.js
```diff
@@ -0,0 +1,45 @@
...
+    fs.mkdirSync(directory, { recursive: true })
+  }
+  return page.screenshot({ path: path.join(directory, `${screenshotName}.png`) })
+}
+
+test('banks transactions', async () => {
+  await page.goto('http://banks.cozy.tools:8080/')
+  await page.click('#password')
+  await page.type('#password', 'cozy')
+  await page.click('text=SE CONNECTER')
```
Crash--, line 32: What if we had a data-testid on the button? Does QAWolf use it or use the string?

package.json
```diff
@@ -192,7 +196,9 @@
     "node-polyglot": "2.3.1",
     "number-to-locale-string": "1.2.0",
     "piwik-react-router": "0.12.1",
+    "playwright-core": "^1.0.1",
     "pouchdb-adapter-cordova-sqlite": "2.0.5",
+    "qawolf": "^1.0.1",
```
Crash--, line 201: why having qawolf / playwright in both dep and dev-dep? 

Should argos-cli be a dev-dep also?

src/ducks/balance/components/AccountRow.jsx
```diff
@@ -183,6 +183,7 @@ class AccountRow extends React.PureComponent {
           [styles['AccountRow--withOwners']]: shouldShowOwners
         })}
         onClick={onClick}
+        data-testid={`balance.account-row.${account._id}`}
```
Crash--, line 186: This kind of selector will be hard to maintain, specially when we'll introduce few add  / remove actions. `account._id` will change.

ptbrowne: _id in fixtures are fixed. We could also fix ids of docs that we add/remove. Or get the testid from the row that was just added in the UI (after adding a doc). I am not so sure about the difficulty here. Do you have a suggestion here ?

.qawolf/banks.test.js
ptbrowne: It will use the data-testid.

Crash--: Thanks!

src/ducks/balance/components/AccountRow.jsx
Crash--: Pour le moment non et c'est pas bloquant car ton test est pour le moment "statique" ou provient de ta fixture. Mais je sais juste que j'évite de mettre des ID dans le nom des sélecteurs, car dans ma tête après on a un scénario pour créer un account (sans se baser sur des fixtures) et pouvoir se balader dedans. Et donc sélectionner quand on aura un `id` dans le sélecteur ça va être compliqué car pas prédictible. Alors que si tu mets par exemple le nom de l'account, tu le connais vraiment par avance.

ptbrowne: OK interesting, basing the testid on the name could be a solution indeed :) Thanks.

src/components/Title/__snapshots__/Title.spec.jsx.snap
```diff
@@ -1,23 +1,23 @@
...
 exports[`Title should display children 1`] = `
-<ForwardRef
+<ForwardRef(MainTitle)
   className="Title"
   ellipsis={true}
   tag="h1"
 >
   content
-</ForwardRef>
+</ForwardRef(MainTitle)>
```
y-lohse, line 10: Nit, but isn't it strange that these snapshots have changed?

package.json
ptbrowne: I think its qawolf that installed playwright, anyway I put everything in devDeps, good catch.

src/components/Title/__snapshots__/Title.spec.jsx.snap
ptbrowne: Yes, I suspect that at some point a newer version of cozy-ui was used to make the screenshot. THis should not be part of the PR. I removed this change.

ptbrowne:
@y-lohse  The screenshots upload is not done through the CI, see the commit description here [https://github.com/cozy/cozy-banks/pull/1746/commits/12baf28cdee5318d9b8e6651cd7b1ec0356049eb ](https://github.com/cozy/cozy-banks/pull/1746/commits/12baf28cdee5318d9b8e6651cd7b1ec0356049eb). I will improve the CI part in a second part.

.gitignore
```diff
@@ -48,3 +48,6 @@ desktop.ini
 # Default
 # /!\ KEEP THIS SECTION THE LAST ONE
 !.gitkeep
+
+# Screenshots from qawolf
+.qawolf/screenshots
```
Crash--, line 53: Is it really wanted to have a `.`? It'll hide the folder on some OS / config.

ptbrowne:
I leave this branch open as it maybe interesting in the future ? cc @Crash--

PR #1748, opened 2020-01-13
Pin tweaks

ptbrowne:
- Larger buttons
- Keyboard now takes 2/3 of the screen

### DIFF
--- a/src/ducks/pin/PinTimeout.debug.jsx
+++ b/src/ducks/pin/PinTimeout.debug.jsx
@@ -1,11 +1,16 @@
 import React from 'react'
 
+const redAffixStyle = {
+  color: 'white',
+  background: 'crimson',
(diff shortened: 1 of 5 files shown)


src/ducks/pin/PinTimeout.debug.jsx
```diff
@@ -1,11 +1,16 @@
 import React from 'react'
 
+const redAffixStyle = {
+  color: 'white',
+  background: 'crimson',
```
drazik: Is this a color from cozy-ui? Shouldn't we use a `var(--crimson)`?

ptbrowne: No, this is a CSS color. I don't think it's necessary to use a variable from cozy-ui here as it is only for debug.

PR #1761, opened 2020-01-16
Merge pull request #1761 from cozy/no-green-biometric

ptbrowne:
- We prefer not to have any delay when authenticating via
biometric mean.
- Repair scroll lock/restore (scrollTop / scrollLeft should be saved from document.scrollingElement)
- Do not show pin settings on desktop


ptbrowne:
I reworded the commit, good point, thanks !

PR #1773, opened 2020-02-18
feat(Balance): Show professional expenses and others reimbursements

drazik:
We want to build two new virtual accounts under the reimbursements virtual group:

* Professional reimbursements
* Others reimbursements (not health reimbursements nor professional reimbursements)

These two new accounts are hidden under the `balance.professional-reimb-account` and `balance.others-reimb-account` respectively. This is because they are shown on the balance page but their detailed page is not implemented yet.

I refactored the reimbursements virtual account generation to have something generic. We now have the `buildReimbursementsVirtualAccount` that takes a filter function and some options and returns a function that will generate the requested virtual account from an array of transactions.

PR #1774, opened 2020-02-19
New reimbursements accounts details

drazik:
Following https://github.com/cozy/cozy-banks/pull/1773, this implements the details page for each of new reimbursements accounts.

The Reimbursements component is now able to get the transactions for the document on which we are currently filtering:

* Reimbursements group
* One of the reimbursements accounts

Each account type has its own message when one section (pending reimbursements / reimbursed expenses) is empty. The group uses the same messages as the "others reimbursements" type.

Other than that, the behaviour of the page is the same as before, it's just able to render more than just health expenses now.

I also updated some tests to use @testing-library/react because they were previously quite complicated and testing lots of implementation details.


package.json
```diff
@@ -174,7 +175,7 @@
     "cozy-authentication": "1.19.2",
     "cozy-bar": "7.10.0",
     "cozy-ci": "0.4.1",
-    "cozy-client": "9.7.0",
+    "cozy-client": "9.8.0",
```
Crash--: why not 10.X ? :D

drazik: I just needed the `createMockClient` function for this PR. I didn't want to take the time to deal with breaking change or the risk of breaking something in this PR. I'm going to update it to the latest version in another one.

PR #1778, opened 2020-02-25
feat(AccountsList): Keep original accounts order for reimbursements

drazik:
For "classic" groups, we want to sort the accounts by ascending balance.
But for the reimbursements virtual group, we want to keep to original
order in which we created them (Health / professional / others)


ptbrowne:
Tests are very clear, thanks !

PR #1786, opened 2020-03-03
fix: Update mobile import keys scripts

ptbrowne:
- Variables are declared at the same place
- Gangsters directory is no longer there, directories are organized by
apps

Check-dir now tries to check for the cozy-banks directory inside
the password store, otherwise we could have a false positive, since we
could check existence of the root password directory instead of the
cozy one.

### DIFF
--- a/scripts/import_mobile_keys
+++ b/scripts/import_mobile_keys
@@ -2,12 +2,10 @@
 
 set -e
 
-COZY_DIRECTORY=${COZY_DIRECTORY:-""}
-PASSWORD_STORE_PATH=${PASSWORD_STORE_PATH:-"$HOME/.password-store/${COZY_DIRECTORY}"}
-
 function check_dir {
-  if [[ ! -d $PASSWORD_STORE_PATH ]]; then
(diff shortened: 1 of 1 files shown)


scripts/import_mobile_keys
```diff
@@ -28,10 +26,14 @@ function cp_from_keepass {
   cp "$path" "$2"
 }
 
-check_dir
+
+COZY_DIRECTORY=${COZY_DIRECTORY:-""}
+PASSWORD_STORE_PATH=${PASSWORD_STORE_PATH:-"$HOME/.password-store/${COZY_DIRECTORY}"}
+BANKS_DIR=${PASSWORD_STORE_PATH}cozy-banks
```
Crash--:
```suggestion
BANKS_DIR_IN_PASSWORD_STORE=${PASSWORD_STORE_PATH}cozy-banks
```
?

At first I thought the cozy-banks' code directory should be contained within password store path

ptbrowne: 💯 Fixed

PR #1790, opened 2020-03-13
E2E testing improvements

ptbrowne:
Since we do not run them often, they tend to fail at some point
due to small errors (for example the update of cozy-ui that leads
to the import of components with HammerJS, which does not work
in a nodeJS environment). So lots of small fixes.

Simplification : removal of cozy-client-js from services


config/webpack.target.services.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
     filename: '[name].js'
   },
   optimization: {
+    minimize: false,
```
Crash--: Why this?

ptbrowne: Because it's easier to understand stack trace, and it is speedier to build, this should not be commmited. I am fixing right now.

PR #1793, opened 2020-03-30
Use cozy-scripts

ptbrowne:
Banks used ad-hoc scripts. Since we have cozy-scripts, it is better to
use it as we will beneficiate from its improvements.

I haven't touched to "services" config, it should be done later.


package.json
```diff
@@ -187,12 +187,15 @@
...
     "cozy-realtime": "3.3.1",
+    "cozy-scripts": "3",
     "cozy-stack-client": "8.7.0",
     "cozy-ui": "29.1.0",
     "d3": "5.11.0",
     "date-fns": "1.30.1",
     "detect-node": "2.0.4",
     "element-scroll-polyfill": "1.0.1",
+    "eslint-plugin-prettier": "^3.1.2",
+    "eslint-plugin-react-hooks": "^3.0.0",
```
Crash--, line 198: It should come with https://github.com/cozy/cozy-libs/blob/master/packages/eslint-config-cozy-app/package.json#L27 non ?

config/webpack.target.browser.js
```diff
@@ -2,41 +2,20 @@
...
 
-module.exports = {
-  entry: [
-    require.resolve('@babel/polyfill'),
-    path.resolve(__dirname, '../src/main')
-  ],
+module.exports = merge(appConfig, {
+  resolve: {
+    extensions: ['.browser.js', '.browser.jsx']
+  },
```
Crash--, line 11: it can be a great addition to cc default config I think. 

I think it's too bad that banks doesn't use https://github.com/cozy/create-cozy-app/blob/master/packages/cozy-scripts/config/webpack.target.browser.js & https://github.com/cozy/create-cozy-app/blob/master/packages/cozy-scripts/config/webpack.target.mobile.js

ptbrowne: > I think it's too bad that banks doesn't use cozy/create-cozy-app:packages/cozy-scripts/config/webpack.target.browser.js@master & cozy/create-cozy-app:packages/cozy-scripts/config/webpack.target.mobile.js@master 


Yes we cannot solve all problems in one go. Let's merge this first.

package.json
ptbrowne: I figured I would remove dependencies managed via cozy-scripts in a later PR, but we may as well do it now : https://github.com/cozy/cozy-banks/pull/1793/commits/4b1bb1a15f6282a5d392b8f9abc76065605bd21f

PR #1796, opened 2020-04-07
feat: Add WEB_OAUTH flag to build the web version but with OAuth Auth

Crash--:
The idea is to be able to embed our web build in an other native app
and load it locally.

run: env WEB_OAUTH=true yarn build:browser

### DIFF
--- a/config/webpack.target.browser.js
+++ b/config/webpack.target.browser.js
@@ -3,7 +3,31 @@
 const path = require('path')
 const webpack = require('webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
+const { WEB_OAUTH } = require('./webpack.vars')
 
+let plugins = [
+  new webpack.DefinePlugin({
+    __TARGET__: JSON.stringify('browser'),
+    __POUCH__: JSON.stringify(process.env.FORCE_POUCH ? true : false),
+    __OAUTH_AUTH__: JSON.stringify(WEB_OAUTH ? true : false)
+  }),
+  new HtmlWebpackPlugin({
+    alwaysWriteToDisk: true,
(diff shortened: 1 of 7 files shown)


config/webpack.target.browser.js
```diff
@@ -3,7 +3,31 @@
...
+    alwaysWriteToDisk: true,
+    template: path.resolve(__dirname, '../src/index.ejs'),
+    title: 'Cozy Banks',
+    minify: {
+      collapseWhitespace: true
+    }
+  })
+]
+
+if (WEB_OAUTH) {
```
ptbrowne, line 24: I am not sure that WEB_OAUTH is the right term here since we have a reference to the cozy-bar below, which is not linked to "oauth". Could we find a better term here ? Should we have another target instead of webpack.target.browser ?

Crash--: I didn't want to create too much noise around the webpack config since you're migrating it to cozy-scripts. I don't know if I really want that PR to be merged right now. 

But yeah you're right, I think the goal is to: 
- keep the __OAUTH__AUTH__ flag to know if we want an OAuth auth or not and remove the check for __TARGET__ === mobile
- add an `embedded` target 
- add an `embedded` config

The result will be: 
- target mobile will use the `webpack.config.embedded` to inject cozy-bar, cozy-client-js but keep other plugins like providing pouchDB stuff
- target embedded will only use `webpack.config.embedded` I think 
- Remove the check of target === mobile to use OAUTH_AUTH when needed. 

WDYT about that?

ptbrowne: I like it !

Crash--:
I'm closing this PR since there is no need anymore to use this kind of hack. We use an override to do what we have to do, and we only serve the web version for now.

PR #1799, opened 2020-04-15
Update cozy-client

ptbrowne:
We were a bit late :sweat:

Had to remove withMutations otherwise we had warnings in our
tests and we do not allow warnings in Banks tests.


src/actions/accounts.spec.js
```diff
@@ -1,11 +1,14 @@
 import { removeStats } from './accounts'
 import CozyClient from 'cozy-client'
 
-jest.mock('cozy-client')
-
-const client = new CozyClient()
+const client = new CozyClient({})
```
Crash--, line 4: not createMockClient() ?

ptbrowne: No

ptbrowne: This is not the point of the PR, PR that is already making things better by removing withMutations.

Crash--: J'avais pas vu la ligne suivante...

PR #1805, opened 2020-05-04
Recurrence UI step 5 : Tweaks and fixes

ptbrowne:
Lots of little things here. Commits are self explaining and/or
have descriptions.


src/components/Header/__snapshots__/Header.spec.jsx.snap
```diff
@@ -1,41 +1,96 @@
...
+    >
+      content
+    </div>
+  </div>,
+  <div
+    className="HeaderColor_normal"
+  >
+    content
+  </div>,
+]
```
Crash--, line 19: I don't understand why we have an array with twice the content here?

src/ducks/recurrence/RecurrencePage.jsx
```diff
@@ -54,6 +55,9 @@ const useDocument = (doctype, id) => {
   return client.getDocumentFromState(doctype, id)
 }
 
+const identity = x => x
+const portal = children => ReactDOM.createPortal(children, document.body)
```
Crash--, line 60:
```suggestion
//see Cozy-UI's issue: https://github.com/cozy/cozy-ui/issues/1462
const identity = x => x
const portal = children => ReactDOM.createPortal(children, document.body)
```

webpack.config.js
```diff
@@ -58,6 +59,23 @@ const modeConfig = production
 
 const config = merge(modeConfig, withTarget)
 
+const removeCSSMQPackerPlugin = config => {
```
Crash--, line 62: I know that you know and you'll remove those lines after an cozy-scripts upgrade, but hey adding a link to the issue and adding a todo comment can be great for external ppl.
```suggestion
//TODO remove those lines after https://github.com/cozy/create-cozy-app/pull/1326 is fixed
const removeCSSMQPackerPlugin = config => {
```

src/ducks/account/AccountSwitch.styl
```diff
@@ -1,7 +1,5 @@
 // TODO understand why this is necessary
 // for the account switch to work on mobile
-@require 'objects/*'
```
Crash--, line 3: can we remove also the comment above?

src/components/Header/__snapshots__/Header.spec.jsx.snap
ptbrowne: Good catch. A bad bit of inattention on my part. The reason is simple actually : as I was `find`ing the `divs` in the `setup`, two divs were returned (one child of the other), thus causing two elements in the array.

PR #1809, opened 2020-05-05
feat: Ability to delete a recurrence

ptbrowne:
- Added a suppression button in the recurrence status menu
- Transactions are marked as non recurrent
- Recurrence object is deleted

### DIFF
--- a/src/ducks/recurrence/utils.js
+++ b/src/ducks/recurrence/utils.js
@@ -27,7 +27,7 @@ export const getLabel = bundle => {
 }
 
 export const getCategories = bundle => {
-  return bundle.categoryId.split(' / ')
+  return bundle.categoryIds
 }
 
 export const getFrequencyWord = freq => {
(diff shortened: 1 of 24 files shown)


src/ducks/recurrence/utils.js
```diff
@@ -54,8 +54,17 @@ export const getFrequencyText = (t, recurrence) => {
...
+  if (t === 'string') {
+    const amount = bundle.amount.split(' / ')
+    return parseInt(amount[0], 10)
+  } else if (t === 'array') {
+    return bundle.amount[0]
+  } else if (t === 'null') {
+    return 0
+  } else if (t === 'number') {
+    return t
+  }
```
Crash--, line 67: Hmmm 🤔 

Why not only having an array of number?

ptbrowne: Good point : https://github.com/cozy/cozy-banks/pull/1809/commits/3b30f7eee489ac5f92d0d98496a0a60aaadeeb97

PR #1812, opened 2020-05-11
Recurrence UI 9

ptbrowne:
Tweaks after our first user feedback (Fab)

Each commit should be self explanatory


src/ducks/recurrence/DebugRecurrencePage.jsx
```diff
@@ -124,17 +126,21 @@ const getColor = bundle => {
...
 }
 
-const List = ({ children }) => {
-  return <ul className="u-m-0 u-ph-1 u-pv-half">{children}</ul>
-}
-
 const CategoryNames = ({ bundle }) => {
   return (
-    <List>
+    <div>
```
Crash--, line 131: Do we need this div? Or a fragment is enough?

ptbrowne: It it not really need but not really a problem also. I don't know :) I keep it, it is in the debug page so it is not really a problem for the user.

PR #1815, opened 2020-05-15
Recurrence UI 10

ptbrowne:
Lots of small improvements. Most come from UI.
Each commit should be self explanatory.

### DIFF
--- a/src/components/ActionMenuHelper/styles.styl
+++ b/src/components/ActionMenuHelper/styles.styl
@@ -1,0 +1,3 @@
+.ActionMenuHelper__menu
+    position relative
+    top  .5rem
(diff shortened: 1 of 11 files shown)


src/components/ActionMenuHelper/styles.styl
```diff
@@ -0,0 +1,3 @@
+.ActionMenuHelper__menu
+    position relative
+    top  .5rem
```
Crash--, line 3: Is it the behavior we want for all our ActionMenu ?

ptbrowne: Yep I think so. @joel-costa, the margin between the button and the menu here is expected to be in all actions menu ?

image

@Crash-- there is already a discussion on cozy-ui on ActionMenuHelper here : https://github.com/cozy/cozy-ui/issues/1463. I plan to move the component ASAP.

Crash--: Thanks for pointing the issue!

PR #1820, opened 2020-05-29
fix: No need to reload page when changing blur setting

ptbrowne:
Connect the flag store to Figure defaultProp "blurred" so that when
the flag amount_blur is changed, the default prop is changed. Before,
we needed to reload the page/app after changing the setting.

### DIFF
--- a/src/utils/flag.js
+++ b/src/utils/flag.js
@@ -55,4 +55,10 @@ Figure.defaultProps = Object.assign({}, Figure.defaultProps, {
   blurred: flag('amount_blur') ? true : false
 })
 
+flag.store.on('change', function(flagName) {
+  if (flagName == 'amount_blur') {
+    Figure.defaultProps.blurred = flag('amount_blur')
+  }
+})
+
 window.flag = flag


src/utils/flag.js
```diff
@@ -55,4 +55,10 @@ Figure.defaultProps = Object.assign({}, Figure.defaultProps, {
   blurred: flag('amount_blur') ? true : false
 })
 
+flag.store.on('change', function(flagName) {
+  if (flagName == 'amount_blur') {
+    Figure.defaultProps.blurred = flag('amount_blur')
```
Crash--, line 60: Does it reload also the mounted Figure component?

ptbrowne: No, at this point in the app, you are in the settings, there are no mounted Figure components.

Crash--: OK thanks 👍

ptbrowne: Another solution would be in banks to use an "overloaded" FlagFigureComponent component that would observe flags through `connect`. This component would also need to be used in FigureBlock. That'd mean FigureBlock should have a FigureComponent prop set by default to FlagFigureComponent.

ptbrowne: I've moved the discussion there as I think the fix is OK and that the discussion could take some time to resolve : https://github.com/cozy/cozy-ui/issues/1478

PR #1821, opened 2020-05-29
fix: Moment should be present in stat service

ptbrowne:
We should not mock moment as empty in services, 
otherwise the stat service fails


ptbrowne:
Yes we use moment-timezones.

PR #1823, opened 2020-06-02
fix: Do not crash when spanish locale is chosen

ptbrowne:
When choosing spanish, Banks would crash and show a blank page.

1. Spanish is added in builds
2. We do not crash if an unknown locale is chosen


ptbrowne:
Good point, thanks, I added a test in 38ad430d

PR #1825, opened 2020-06-04
feat: Add search icon in balance page and back button in search page

ptbrowne:
Add a way to go to search page from balance page.

The goal is to have a 1st version of the search available for internal
use.

### DIFF
--- a/src/ducks/balance/components/BalanceHeader.styl
+++ b/src/ducks/balance/components/BalanceHeader.styl
@@ -12,3 +12,18 @@
         flex-basis 159px
         padding-top: .5rem
         padding-bottom 0
+
+.SearchIcon
+    position absolute
+    right 2rem
+    color white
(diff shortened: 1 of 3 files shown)


src/ducks/balance/components/BalanceHeader.styl
```diff
@@ -12,3 +12,18 @@
         flex-basis 159px
         padding-top: .5rem
         padding-bottom 0
+
+.SearchIcon
+    position absolute
+    right 2rem
+    color white
+    z-index 1
```
y-lohse, line 20: Would it make sense to use the z-index scale from cozy-ui here?

ptbrowne: No, I don't think so. The z-index is very local here, just to say that the search icon is above the rest of the current container.

PR #1828, opened 2020-06-05
Improvements

ptbrowne:
2 unrelated small improvements

- fix: Do not create virtual accounts if there are no transactions

   If we have virtual accounts we do not display the special Empty page that
   is shown if the user has no data

- feat: Sort bundles by category then by recency

In the recurrence page, we want the recurrence bundles
grouped by category and then sorted by date.

### DIFF
--- a/src/targets/mobile/config.xml
+++ b/src/targets/mobile/config.xml
@@ -87,7 +87,7 @@
     </platform>
     <preference name="StatusBarBackgroundColor" value="#95999D" />
     <engine name="android" spec="7.1.4" />
-    <engine name="ios" spec="4.5.5" />
+    <engine name="ios" spec="^5.1.1" />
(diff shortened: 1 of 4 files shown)


src/targets/mobile/config.xml
```diff
@@ -87,7 +87,7 @@
     </platform>
     <preference name="StatusBarBackgroundColor" value="#95999D" />
     <engine name="android" spec="7.1.4" />
-    <engine name="ios" spec="4.5.5" />
+    <engine name="ios" spec="^5.1.1" />
```
Crash--, line 90: c'est pas trop du "chore" ça ^^'

ptbrowne: Ah yes effectivement. Merci :/

PR #1838, opened 2020-07-09
feat: Update libs

ptbrowne:
Update cozy-harvest-lib to have the direct connection to
BI. Then update the dependencies of cozy-harvest-lib.

### DIFF
--- a/test/__mocks__/cozy-keys-lib/index.js
+++ b/test/__mocks__/cozy-keys-lib/index.js
@@ -1,0 +1,6 @@
+const keysLib = jest.requireActual('cozy-keys-lib')
+
+module.exports = {
+  withVaultClient: x => x,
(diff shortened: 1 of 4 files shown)


test/__mocks__/cozy-keys-lib/index.js
```diff
@@ -0,0 +1,6 @@
+const keysLib = jest.requireActual('cozy-keys-lib')
+
+module.exports = {
+  withVaultClient: x => x,
```
Crash--, line 4: Is it a new method?

ptbrowne: Nope but before cozy-harvest-lib did not import cozy-keys-lib so we did not have the problem.

PR #1840, opened 2020-07-13
Refactor for overrides

ptbrowne:
Split and make it easier to override logic for

- reconnection of an account due to a konnector error
- customisation of account errors visible on transaction page
- customisation of account caption of balance page


src/ducks/transactions/TransactionPageErrors/errors.js
```diff
@@ -0,0 +1,53 @@
+import sortBy from 'lodash/sortBy'
+import get from 'lodash/get'
+import keyBy from 'lodash/keyBy'
+import mapValues from 'lodash/mapValues'
+import { isErrored, isBankTrigger } from 'utils/triggers'
+
+const getCreatedByApp = acc => get(acc, 'cozyMetadata.createdByApp')
```
y-lohse, line 7: This and `isActionableError ` could be in a cozy-client models imo, but I don't know if that's possible while keeping them overridable?

ptbrowne: I am not sure I understand your question ? Any package can be overrided in an override repository.

y-lohse: As far as I remember, overriding a dependency was a bit trickier, but maybe I misremember. But yeah in that case I think cozy-client models would be a good place for these functions.

ptbrowne: Done in https://github.com/cozy/cozy-client/pull/738

ptbrowne:
I'm merging and will deal with the replacements of methods when https://github.com/cozy/cozy-client/pull/738 is merged.

PR #1844, opened 2020-07-20
Improvements

ptbrowne:
Several unrelated improvements

- Remove the presence of cozy-ui's Toggle to use Material UI's Switch

We had already begun the transition work, the Switch had been used on
the account panels. Now, the group settings and categories page's
IncomeToggle use Switches.

- Defensive programming for a case that should not happen for the
  onScroll handler. This callback is throttled, so it can
  happen that it is called while the component has been unmounted.

- Add missing stats object in recurrence objects inside fixtures. This
  would cause the transactions page to crash. This is turn would cause
  the transactions to be unmounted while the scroll was happening leading
  to the bug fixed above (handleScroll does nothing when component is
  unmounted).

- Use validColor instead of emerald for theming

### DIFF
--- a/src/ducks/categories/CategoriesHeader.jsx
+++ b/src/ducks/categories/CategoriesHeader.jsx
@@ -1,8 +1,8 @@
-import React, { PureComponent, Fragment } from 'react'
+import React, { PureComponent, Fragment, useCallback } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 import { translate, withBreakpoints } from 'cozy-ui/transpiled/react'
-import Toggle from 'cozy-ui/transpiled/react/Toggle'
+import Switch from 'components/Switch'
 import RawBreadcrumb from 'cozy-ui/transpiled/react/Breadcrumbs'
 import { AccountSwitch } from 'ducks/account'
 import BackButton from 'components/BackButton'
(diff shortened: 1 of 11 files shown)


src/ducks/categories/CategoriesHeader.jsx
```diff
@@ -36,9 +36,22 @@ const stTableCategory = catStyles['bnk-table-category']
...
+      return onToggle(ev.target.checked)
+    },
+    [onToggle]
+  )
   return (
     <div className={cx(styles.CategoriesHeader__Toggle, styles[theme])}>
-      <Toggle id="withIncome" checked={withIncome} onToggle={onToggle} />
+      <Switch
+        id="withIncome"
+        disableRipple
```
y-lohse, line 50: I'm surprised we need to specify that, isn't there a global option to disable the ripple effects?

ptbrowne: I haven't seen one yet.

Crash--: Capture d’écran 2020-07-20 à 17 24 46

I guess we can do the same with the Switch component (https://material-ui.com/getting-started/faq/)

ptbrowne: Nice, thanks !

PR #1847, opened 2020-07-23
feat: Ask for personal info before a transfer

ptbrowne:
In some cases, we will want to ask the user some information
before allowing transfers.

- The myself io.cozy.contacts is checked to see that the birth city
  and nationality are present
- If not, the PersonalForm asks for the data and updates the myself contact
- The personal form can be displayed in two flavors depending on the page.
  It can be in a "Page" in the transfer page tab : when present, the
  transfer stepper is hidden, and the form takes its place. It can also
  be in a modal because it can be also accessed from the settings.

🚩 The "banks.transfers.need-personal-information" flag controls

- If the personal information section is available in the settings
- If personal information is asked before a transfer

image


src/locales/en.json
```diff
@@ -683,6 +683,18 @@
...
   },
 
+  "PersonalInfo": {
+    "birthcity": "Birth city",
+    "nationality": "Nationality",
+    "save-cta": "Save information",
+    "modal-title": "Edit personal information",
+    "info-saved-succesfully": "Your personal info has been updated",
+    "info": {
+      "title": "Why is this information necessary ?",
```
y-lohse, line 698:
```suggestion
      "title": "Why is this information necessary?",
```

Spaces before question marks is a french thing :)

src/locales/fr.json
```diff
@@ -694,6 +694,18 @@
     "generic": "Mes factures"
   },
 
+  "PersonalInfo": {
+    "birthcity": "Commune de naissance",
+    "nationality": "Nationalité",
+    "save-cta": "Sauver mon profil",
```
y-lohse, line 705:
```suggestion
    "save-cta": "Sauvegarder mon profil",
```

J'aurais plus mis ca non?

src/ducks/transfers/TransferPage.jsx
```diff
@@ -105,6 +108,10 @@ const NoBank = () => {
   )
 }
 
+const isMyselfSufficientlyFilled = myself => {
```
y-lohse, line 111: This could be good to have in cozy-client models as we may need it elsewhere than in banks.

src/ducks/personal-info/index.spec.jsx
```diff
@@ -0,0 +1,73 @@
...
+    fireEvent.click(btn)
+    expect(client.save).toHaveBeenCalledWith({
+      _id: '1337',
+      _type: 'io.cozy.contacts',
+      birthcity: 'Pau',
+      id: '1337',
+      nationality: 'FR'
+    })
+
+    await sleep(0) // Since the save is await
```
y-lohse, line 64: Nit: we could also use `waitFor`.

src/ducks/transfers/TransferPage.jsx
```diff
@@ -137,7 +152,12 @@ class TransferPage extends React.Component {
       amount: '',
       password: '',
       label: '',
-      date: new Date().toISOString().slice(0, 10)
+      date: new Date().toISOString().slice(0, 10),
+      showingPersonalInfo:
+        this.props.myself &&
+        this.props.myself.data &&
+        this.props.myself.data[0] &&
```
y-lohse, line 159: Nit: `_.get(this.props, 'myself.data[0]')` makes this sort of code less error prone imo.

src/ducks/transfers/TransferPage.jsx
```diff
@@ -370,9 +422,14 @@ class TransferPage extends React.Component {
       accounts.data
     )
 
+    const { isMobile } = this.props.breakpoints
+    const PersonalInfoComponent = isMobile
```
y-lohse, line 426: Slightly surprised by this, Modals on mobile are fullscreen so I'm not sure what the difference between a page and a not closable, full screen modal is.

src/ducks/transfers/TransferPage.spec.jsx
```diff
@@ -59,3 +83,57 @@ describe('transfer page', () => {
...
+        <TransferPage />
+      </AppLike>
+    )
+
+    return { root }
+  }
+
+  afterEach(() => {
+    flag('banks.transfers.need-personal-information', false)
+  })
```
y-lohse, line 121: Why do we need this, since the module is mocked?

src/ducks/transfers/TransferPage.spec.jsx
```diff
@@ -59,3 +83,57 @@ describe('transfer page', () => {
...
+
+  it('should show personal info if myself contact not sufficiently filled', async () => {
+    window.innerWidth = 300 // isMobile -> true
```
y-lohse, line 124: Out of curiosity, why the explicit test on mobile?

src/ducks/transfers/TransferPage.jsx
ptbrowne: The fullscreen modal would hide the bottom navigation tabs.

src/ducks/transfers/TransferPage.spec.jsx
ptbrowne: No need you're right !

src/ducks/transfers/TransferPage.jsx
ptbrowne: Here, the sufficiently filled is dependent on the transfer function. I don't think it should be in the models yet. I'll try to keep in mind if it resurfaces elsewhere.

src/ducks/transfers/TransferPage.spec.jsx
ptbrowne: Because otherwise on desktop, the transfer stepper is shown behind the modal. I've added a test for desktop that checks that the stepper is shown behind the modal.

src/ducks/personal-info/index.spec.jsx
ptbrowne: I tried to use waitFor, but to use it, I had to update the jest environment to use jsdom sixteen. I do not want to do this in the app. I would prefer to do this via cozy-scripts. Otherwise our jest.setup become different between our apps. I would like to consolidate them more. So I kept the sleep.

ptbrowne: I opened an issue here https://github.com/cozy/create-cozy-app/issues/1336

PR #1849, opened 2020-07-24
feat: Personal info 2

ptbrowne:
Two small things that I had overlooked. Commits are hopefully
self explanatory.


src/ducks/transfers/TransferPage.spec.jsx
```diff
@@ -133,7 +133,14 @@ describe('personal info', () => {
       const { root } = setup({
         myselfData: {
           birthcity: 'Compiègne',
-          nationality: 'FR'
+          nationality: 'FR',
+          cozyMetadata: {
+            updatedByApps: [
+              {
+                slug: 'slug'
```
Crash--, line 140: Does it mean that `manifest.slug` === "slug" when testing?

ptbrowne: Yep.

PR #1857, opened 2020-08-10
Blank emails

ptbrowne:
Since the introduction of cozy-scripts in the banks repository, we use
the TerserPlugin from cozy-scripts. Previously, we would configure this
plugin not to mangle function names in the webpack.target.services file.
Now, we need to directly tweak the options of cozy-scripts to achieve
the same goal.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const get = require('lodash/get')
 const merge = require('webpack-merge')
 const { mergeAppConfigs } = require('cozy-scripts/utils/merge')
 const {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -207,6 +207,7 @@
     "redux-raven-middleware": "1.2.0",
     "redux-thunk": "2.3.0",
     "reselect": "4.0.0",
+    "terser-webpack-plugin": "^4.0.0",
```
Crash--, line 210: Do we need to add this requirement since you don't use it directly, no? It should be updated on cozy-script side? Also, this is "just" a devDep ?

webpack.config.js
```diff
@@ -71,9 +72,27 @@ const removeCSSMQPackerPlugin = config => {
...
+  const terserOptions = config.optimization.minimizer[0].options.terserOptions
+  terserOptions.mangle = {
+    keep_fnames: true
+  }
+}
+
 // TODO remove those lines after https://github.com/cozy/create-cozy-app/pull/1326
 // is fixed
 removeCSSMQPackerPlugin(config)
+configureTerserPlugin(config)
```
Crash--, line 95: Are we keeping the name even for non `services` bundles?

ptbrowne: You're right, I had planned to add the if and forgot it, it is there now :)

PR #1858, opened 2020-08-10
Various styling tweaks

ptbrowne:
Small graphic tweaks.
The most important is a fix of ellipsis in the bar title.

Commits are self explanatory.

### DIFF
--- a/src/components/BarBalance/styles.styl
+++ b/src/components/BarBalance/styles.styl
@@ -1,3 +1,4 @@
 .BarBalance
     font-size 1.125rem
     color var(--primaryTextColor)
+    margin-right 0.5rem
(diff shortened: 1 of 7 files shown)


src/components/BarBalance/styles.styl
```diff
@@ -1,3 +1,4 @@
 .BarBalance
     font-size 1.125rem
     color var(--primaryTextColor)
+    margin-right 0.5rem
```
Crash--, line 4: The idea was not to use `spacing_values` from UI to deal with the gutters ?

ptbrowne: No, I had simply not thought of using spacing_values :) Fixed, thanks !

PR #1863, opened 2020-08-24
chore: Upgrade harvest to version 2

doubleface:
This upgrade contains a bug fix for running banking connectors in maif-epa-custom context


ptbrowne (approved):
Nothing to say about the upgrade. I would have liked a description, telling us why the upgrade is important though. LGTM

Issue #1864, opened 2020-08-25
Refactor : use isQueryLoading and hasQueryBeenLoaded from cozy-client


JF-Cozy:
`isQueryLoading` and `hasQueryBeenLoaded` from `cozy-client` should replace `isCollectionLoading` and `hasBeenLoaded` from here https://github.com/cozy/cozy-banks/blob/master/src/ducks/client/utils.js#L8

should be `import { isQueryLoading, hasQueryBeenLoaded } from 'cozy-client'` instead of `import { isCollectionLoading, hasBeenLoaded } from 'ducks/client'`

PR #1866, opened 2020-08-27
refactor: Use isQueryLoading and hasQueryBeenLoaded from cozy-client

JF-Cozy:
Refactor to use cozy-client functions

fix https://github.com/cozy/cozy-banks/issues/1864


ptbrowne (approved):
Thanks !

PR #1867, opened 2020-08-28
Harvest in banks

ptbrowne:
AccountSettings now use Harvest to edit an account.

- This functionality is behind the flag "banks.harvest-accounts-settings"
- Since banks is on react-router@3, we cannot use directly the routes from
  Harvest and we have to simulate a Switch on our end.

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -208,7 +208,7 @@
       "verbs": ["GET"]
     },
     "triggers": {
-      "description": "Used to configure bank account triggers through Harvest",
+      "description": "Used to configure bank accounts",
       "type": "io.cozy.triggers"
     },
     "settings": {
@@ -241,7 +241,7 @@
       "type": "io.cozy.jobs"
     },
     "accounts": {
-      "description": "Used to create a temporary account for transfers and to configure accounts through Harvest",
+      "description": "Used to create a temporary account for transfers and to configure accounts",
       "type": "io.cozy.accounts"
     },
     "contacts": {
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -208,9 +208,8 @@
       "verbs": ["GET"]
     },
     "triggers": {
-      "description": "Used to add CTA to collect",
-      "type": "io.cozy.triggers",
-      "verbs": ["GET", "POST"]
+      "description": "Used to configure bank account triggers through Harvest",
+      "type": "io.cozy.triggers"
     },
     "settings": {
       "description": "Required by the cozy-bar to display Claudy and know which applications are coming soon",
@@ -242,9 +241,8 @@
       "type": "io.cozy.jobs"
     },
     "accounts": {
-      "description": "Used to create a temporary account for transfers",
-      "type": "io.cozy.accounts",
-      "verbs": ["POST", "DELETE", "GET"]
+      "description": "Used to create a temporary account for transfers and to configure accounts through Harvest",
+      "type": "io.cozy.accounts"
     },
     "contacts": {
       "description": "Used to link accounts to contacts",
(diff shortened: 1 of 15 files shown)


src/ducks/settings/AccountsSettings.jsx
```diff
@@ -96,58 +92,61 @@ const AccountsTable = ({ accounts }) => {
...
-        )}
-      </div>
-    )
+const AccountsSettings = props => {
+  const { t } = useI18n()
+  const { accountsCollection } = props
+
+  if (
+    isCollectionLoading(accountsCollection) &&
+    !hasBeenLoaded(accountsCollection)
```
Crash--, line 110: https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/utils.js#L30

src/ducks/settings/AccountsSettings.jsx
```diff
@@ -96,58 +92,61 @@ const AccountsTable = ({ accounts }) => {
...
+  ])
+  const accountBySharingDirection = groupBy(sortedAccounts, account => {
+    return account.shared === undefined
+  })
+
+  const myAccounts = accountBySharingDirection[true]
+  const sharedAccounts = accountBySharingDirection[false]
+
+  return (
+    <div>
```
Crash--, line 126: Do we really need this div? A fragment should be enough?

src/ducks/settings/HarvestBankAccountSettings.jsx
```diff
@@ -0,0 +1,100 @@
...
+    <Query query={Q('io.cozy.accounts').getById(accountId)}>
+      {({ data: account, fetchStatus }) => {
+        if (fetchStatus === 'loading') {
+          return <HarvestSpinner />
+        } else {
+          const konnectorSlug = account.account_type
+          return (
+            <Query query={Q('io.cozy.konnectors').getById(konnectorSlug)}>
+              {({ data: { attributes: konnector }, fetchStatus }) => {
+                if (fetchStatus === 'loading') {
```
Crash--: shouldn't we check if there is no error too?

manifest.webapp
```diff
@@ -242,9 +241,8 @@
       "type": "io.cozy.jobs"
     },
     "accounts": {
-      "description": "Used to create a temporary account for transfers",
-      "type": "io.cozy.accounts",
-      "verbs": ["POST", "DELETE", "GET"]
+      "description": "Used to create a temporary account for transfers and to configure accounts through Harvest",
+      "type": "io.cozy.accounts"
```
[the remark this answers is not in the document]
Crash--: Also, we shoud use "verbs": ["ALL"]

Crash-- (approved):

manifest.webapp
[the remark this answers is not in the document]
ptbrowne: Why should we use verbs : all ? When not used, it is implicit, IMHO we should not add stuff that does not serve.

src/ducks/settings/AccountsSettings.jsx
ptbrowne: It is taken care of by julien in another PR.

src/ducks/settings/HarvestBankAccountSettings.jsx
ptbrowne: Done

manifest.webapp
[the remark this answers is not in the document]
ptbrowne: > No need to translate it in french?

No as banks is pre-installed and we do not have any text ready for this. cc @fffflo : it could be handy someday to have permissions texts in French if we plan to have a platform where banks would be installed by the user.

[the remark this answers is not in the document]
Crash--: > Why should we use verbs : all ? When not used, it is implicit, IMHO we should not add stuff that does not serve.

I prefer to be explicit. Specially on permissions.

Crash-- (approved):

PR #1869, opened 2020-09-02
fix: Handle io.cozy.bank.accounts without connection relationship

ptbrowne:
Historically, the bank connectors would not link the
io.cozy.bank.accounts they retrieved with the io.cozy.accounts that ran
the connector. We did not support this usecase. This is done by grouping
those unconnected accounts by konnector slug. This is not optimal as
ideally, we would like those to be grouped by io.cozy.accounts and not
by io.cozy.konnectors. But since we cannot know the io.cozy.accounts,
it is the best we can do.

### DIFF
--- a/src/ducks/settings/AccountsSettings.jsx
+++ b/src/ducks/settings/AccountsSettings.jsx
@@ -32,6 +32,7 @@ import { Unpadded } from 'components/Spacing/Padded'
 
 import HarvestBankAccountSettings from './HarvestBankAccountSettings'
 import DisconnectedAccountModal from 'cozy-harvest-lib/dist/components/DisconnectedAccountModal'
+import { utils } from 'cozy-client/dist/models'
 
 const AccountListItem = ({ account, onClick, secondaryText }) => {
   return (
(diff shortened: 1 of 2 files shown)


src/ducks/settings/AccountsSettings.jsx
```diff
@@ -32,6 +32,7 @@ import { Unpadded } from 'components/Spacing/Padded'
 
 import HarvestBankAccountSettings from './HarvestBankAccountSettings'
 import DisconnectedAccountModal from 'cozy-harvest-lib/dist/components/DisconnectedAccountModal'
+import { utils } from 'cozy-client/dist/models'
```
y-lohse, line 35: It seems `utils` is exported from `models` which is exported from `cozy-client` — imo it's better to import from the root of the package as it's immune from changes in the file structure.

y-lohse (approved):

src/ducks/settings/AccountsSettings.jsx
ptbrowne: OK, will do in subsequent commits.

PR #1870, opened 2020-09-02
Add an unsynchronized mention for disconnected accounts

ptbrowne:
To prepare the user for the slightly different UI that opens when clicking
on the account row, a mention is added that the bank accounts are no
longer synchronized.

Depends on https://github.com/cozy/cozy-ui/pull/1539 and https://github.com/cozy/cozy-libs/pull/1082 not to have red CI (warnings during tests).

Can be reviewed.


src/ducks/settings/AccountsSettings.jsx
```diff
@@ -98,7 +100,16 @@ export const AccountsList_ = ({ accounts }) => {
           <AccountListItem
             key={accounts[0]._id}
             account={accounts[0]}
-            secondaryText={connection ? connection.auth.identifier : null}
+            secondaryText={
+              connection ? (
+                connection.auth.identifier
+              ) : (
+                <Caption className="u-warn">
+                  <Icon icon="sync" size="8" className="u-mr-half" />
```
Crash--, line 108: I thought we had defined size (small / ...) but in fact... no... just on circle

Crash-- (approved):

PR #1873, opened 2020-09-04
Update packages

ptbrowne:
Update cozy-client to solve a deleted document bug.
See https://github.com/cozy/cozy-client/pull/776

Update cozy-harvest-lib and cozy-ui to have the NavigationList
in the ConfigurationTab of the AccountModal in harvest.

Update cozy-scripts and remove the adhoc check-locales script.


ptbrowne:
Since this is only an upgrade of packages (and a small chore), I'm merging as soon as this is green.

PR #1874, opened 2020-09-07
feat: Rely on the bar from the stack

ptbrowne:
For a time, banks needed a new version of the bar and we had relied
on the bar from the node_modules. This is different from the other
apps and we need now to revert this. Having the bar served from
the stack has several advantages:

- It diminishes the build size of the app
- It allows apps to cache the bar
- Every app uses the same bar


Crash-- (approved):
thanks !

PR #1876, opened 2020-09-08
Reconnect trigger from TransactionsPage through Harvest

ptbrowne:
Now we can use Harvest directly to fix a connection from the transaction page.


src/ducks/balance/AccountRowCaption.jsx
```diff
@@ -6,13 +6,15 @@ import Icon from 'cozy-ui/transpiled/react/Icon'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
 import flag from 'cozy-flags'
-import { isErrored } from 'utils/triggers'
+import { trigger as triggerLibs } from 'cozy-client/dist/models'
```
Crash--, line 9: nit: it's sounds weird to import from dist/

Crash-- (approved):

src/ducks/balance/AccountRowCaption.jsx
ptbrowne: I made the change, thanks.

PR #1887, opened 2020-09-23
Fix recurrence bundle page loading

ptbrowne:
Without a "fake" selector, the index could not be inferred by PouchDB,
breaking the loading of the recurrence bundle data.

Since there was no check to see if the query loading was in "failed"
status, the "no data found" message was shown to the user.

### DIFF
--- a/src/ducks/recurrence/queries.js
+++ b/src/ducks/recurrence/queries.js
@@ -17,11 +17,15 @@ export const queryRecurrencesTransactions = recurrences =>
 
 export const bundleTransactionsQueryConn = ({ bundle }) => {
   return {
-    query: () =>
-      queryRecurrenceTransactions(bundle)
+    query: () => {
+      const initialQDef = queryRecurrenceTransactions(bundle)
+      const qDef = initialQDef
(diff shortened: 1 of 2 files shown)


doubleface (approved):

src/ducks/recurrence/queries.js
```diff
@@ -17,11 +17,15 @@ export const queryRecurrencesTransactions = recurrences =>
 
 export const bundleTransactionsQueryConn = ({ bundle }) => {
   return {
-    query: () =>
-      queryRecurrenceTransactions(bundle)
+    query: () => {
+      const initialQDef = queryRecurrenceTransactions(bundle)
+      const qDef = initialQDef
         .sortBy([{ date: 'desc' }])
+        .where({ ...initialQDef.selector, date: { $gt: null } })
```
paultranvan, line 24: The attributes will be automatically indexed  in this order:  `date` should be first to actually use the index. Otherwise, couchdb will do a full-scan to run it.

ptbrowne: Ok thanks, I've modified the indexing.

PR #1889, opened 2020-10-01
Cozy flags fix

ptbrowne:
On Android (built with Travis), there was two copies of cozy-flags used:
one in Harvest and one in Banks. This caused problems since one 
cozy-flag had the correct flags and the other did not have the flag.

To prevent the problem:

- I force the usage of the right cozy-flags in Banks
- I upgrade cozy-flags in banks, so that it is the same version as in
  harvest, which means yarn can hoist the package to the root node_modules
- I will do a PR in harvest to have cozy-flags as a peer dependency


Crash-- (approved):
Nice catch 👍

JF-Cozy (approved):

PR #1898, opened 2020-10-08
refactor: Extract getTracker to be able to override it

ptbrowne:
To be able to override the getTracker, we need to extract this function from the browser.jsx otherwise functions inside browser.jsx that use getTracker would retain a reference to the original getTracker function.


ptbrowne:
I am merging since this is a small refactor and test pass.

PR #1912, opened 2020-10-23
Improve tracking in modals

ptbrowne:
- For events to correctly be attributed to each page, when closing the modal, we need to send a hit to the parent page.
- The transaction modal cannot know from where it is opened so it needs to send a page name that depends on the previous page. To achieve this, a functional variant of trackPage is added.
- Also fixed a bug where the events in the alert edition modal were sent with the complete page name, when we want only the events to have the last part.

### DIFF
--- a/src/components/EditionModal/EditionModal.jsx
+++ b/src/components/EditionModal/EditionModal.jsx
@@ -22,7 +22,11 @@ import resultWithArgs from 'utils/resultWithArgs'
 import Confirmation from 'components/Confirmation'
 
 import { BackIcon } from 'components/BackButton'
-import { useTrackPage, useTracker } from 'ducks/tracking/browser'
+import {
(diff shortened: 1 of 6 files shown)


src/components/EditionModal/EditionModal.jsx
```diff
@@ -212,6 +216,18 @@ const EditionModal = props => {
 
   useTrackPage(trackPageName)
 
+  const handleDismiss = () => {
+    onDismiss()
+
+    // :/ We need to set a timeout otherwise we fire the track page too soon
```
JF-Cozy, line 222: c'est une faute de frappe le `:/` ?

ptbrowne: C'est un smiley

PR #1919, opened 2020-11-02
Remove the fake triggers in the manifest

nono:
With a recent change in the stack^1, it is no longer mandatory to have
a trigger for each service. So, we can remove the triggers for services
that are called from other services in the manifest.

1: https://github.com/cozy/cozy-stack/pull/2672

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -73,8 +73,7 @@
     "categorization": {
       "type": "node",
       "file": "categorization.js",
-      "comment": "This service must be called programmatically from konnectors, this is why the trigger is in the past. See https://github.com/cozy/cozy-banks/blob/master/docs/services.md for more information.",
-      "trigger": "@at 2000-01-01T00:00:00.000Z"
+      "comment": "This service must be called programmatically from konnectors, this is why the trigger is in the past. See https://github.com/cozy/cozy-banks/blob/master/docs/services.md for more information."
     },
     "stats": {
       "type": "node",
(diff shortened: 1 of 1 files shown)


manifest.webapp
```diff
@@ -73,8 +73,7 @@
     "categorization": {
       "type": "node",
       "file": "categorization.js",
-      "comment": "This service must be called programmatically from konnectors, this is why the trigger is in the past. See https://github.com/cozy/cozy-banks/blob/master/docs/services.md for more information.",
-      "trigger": "@at 2000-01-01T00:00:00.000Z"
+      "comment": "This service must be called programmatically from konnectors, this is why the trigger is in the past. See https://github.com/cozy/cozy-banks/blob/master/docs/services.md for more information."
```
ptbrowne, line 76:
```suggestion
      "comment": "This service must be called programmatically from konnectors. See https://github.com/cozy/cozy-banks/blob/master/docs/services.md for more information."
```

manifest.webapp
```diff
@@ -97,8 +96,7 @@
     "budgetAlerts": {
       "type": "node",
       "file": "budgetAlerts.js",
-      "comment": "Service is run inside onOperationOrBillCreate. This is why the trigger is in the past. The service described here is for diagnosis/debug.",
-      "trigger": "@at 2000-01-01T00:00:00.000Z"
+      "comment": "Service is run inside onOperationOrBillCreate. This is why the trigger is in the past. The service described here is for diagnosis/debug."
```
ptbrowne, line 99:
```suggestion
      "comment": "Service is run inside onOperationOrBillCreate. The service described here is for diagnosis/debug."
```

ptbrowne (approved):
Thanks !

PR #1930, opened 2020-11-16
Small mobile fixes

ptbrowne:
- Hide nav bar when keyboard is shown (use cordova-plugin-ionic-keyboard)
- Add sort by field to query otherwise recurrence page does not work

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -294,6 +294,12 @@ While developing, in XCode, in the "Signing" section :
 
 When deploying, Fastlane manages the certificate and may change it.
 
+The certificate with its password is stored in the shared password store.
+
+- `cozy-banks/xcode-signing-certificate-password` : the password
+- `cozy-banks/xcode-signing-certificate-password.p12` : the certificate
+
+
 ##### Push iOS build
 
 To push an iOS build on Testflight, use the following command :
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -8,6 +8,8 @@
 - [Fixtures](#fixtures)
 - [Develop on mobile](#develop-on-mobile)
   - [Get a working Android environment](#get-a-working-android-environment)
+    - [Linux](#linux)
+    - [macOS](#macos)
   - [Build and run the mobile app](#build-and-run-the-mobile-app)
   - [Hot reload on mobile app](#hot-reload-on-mobile-app)
 - [Release](#release)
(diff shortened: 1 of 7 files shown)


docs/dev.md
```diff
@@ -96,11 +98,26 @@ If you want to inject other data or remove some data, you can use [`ACH`](https:
...
+#### macOS
+
+On macOS you can use adoptOpenJDK to have a working Java environment.
+
+See https://github.com/AdoptOpenJDK/homebrew-openjdk
+
+```
+brew cask install adoptopenjdk8
+# Check the adopt open JDK page to see how to install the shell function "jdk"
+jdk 8 # Cordova supports JDK version 1.8 (same as version 8 it seems)
```
Crash--, line 117: Don't we want to break line before # Cordova supports ?

src/ducks/recurrence/queries.js
```diff
@@ -20,7 +20,10 @@ export const bundleTransactionsQueryConn = ({ bundle }) => {
     query: () => {
       const initialQDef = queryRecurrenceTransactions(bundle)
       const qDef = initialQDef
-        .sortBy([{ date: 'desc' }])
+        .sortBy([
+          { 'relationships.recurrence.data._id': 'desc' },
+          { date: 'desc' }
+        ])
```
Crash--, line 26: From https://docs.cozy.io/en/tutorials/data/queries/#sort-with-cozy-client 

> ⚠️ If the fields involved in the sortBy are not indexed, this will force CouchDB to make the sort in memory: this can be acceptable if the query returns few documents, but it is not efficient for large queries.

You should create an index ;)

src/targets/mobile/config.xml
```diff
@@ -112,4 +112,5 @@
     <plugin name="cordova-blur-app-privacy-screen" spec="https://github.com/lifeofcoding/cordova-blur-app-privacy-screen.git" />
     <plugin name="cordova-universal-links-plugin" spec="https://github.com/Mailbutler/cordova-universal-links-plugin" />
     <plugin name="cordova-plugin-network-information" spec="^2.0.1" />
+    <plugin name="cordova-plugin-ionic-keyboard" spec="^2.2.0" />
```
Crash--, line 115: I don't know exactly if this plugin provides a shrink method or not but you should test the login on iOS and Android to see if nothing breaks (I remember that cozy-authentication has a few conditions on Keyboard.X)

Crash-- (approved):

docs/dev.md
ptbrowne: No the comment is intended for the line "jdk 8".

src/ducks/recurrence/queries.js
ptbrowne: The index is automatically created by cozy-client. https://github.com/cozy/cozy-client/blob/ea768adecaf0dd1c52e9876bf68e86d099579a9b/packages/cozy-stack-client/src/DocumentCollection.js#L339

src/targets/mobile/config.xml
ptbrowne: The plugin does not have the shrinkView method of cozy-authentication. Do you have somewhere the reasons why we had to remove the cordova-plugin-keyboard ? I've searched the PR and found https://github.com/cozy/cozy-banks/pull/1311. This PR uninstalls the plugin but I still see the Keyboard being checked in cozy-authentication. I wonder if we still have something to do on the subject ?

src/ducks/recurrence/queries.js
Crash--: Yep also here https://github.com/cozy/cozy-client/blob/master/packages/cozy-pouch-link/src/CozyPouchLink.js#L325

thanks

PR #1937, opened 2020-11-23
feat: Block transfer if flag is set on Bender

ptbrowne:
A flag can be set on Bender to block transfers.

- [x] UI


Crash-- (approved):
I know this is not a new mention in Banks, but first time I see it. I Like the idea of "Gate" 👍

PR #1942, opened 2020-11-27
Tracking improvements

ptbrowne:
Add and modify tracking events
Other small improvements that are described in the commits

### DIFF
--- a/src/ducks/categories/CategoriesPage.jsx
+++ b/src/ducks/categories/CategoriesPage.jsx
@@ -90,7 +90,7 @@ class CategoriesPage extends Component {
     const { categoryName, subcategoryName } = router.params
     trackPage(
       `analyse:${categoryName ? categoryName : 'home'}${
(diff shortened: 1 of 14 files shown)


src/ducks/categories/CategoriesPage.jsx
```diff
@@ -90,7 +90,7 @@ class CategoriesPage extends Component {
     const { categoryName, subcategoryName } = router.params
     trackPage(
       `analyse:${categoryName ? categoryName : 'home'}${
-        subcategoryName ? `:${subcategoryName}` : ''
+        subcategoryName ? `:details` : ''
```
Crash--, line 93: We don't use the subcategoryName anymore? Is it wanted?

Crash-- (approved):

src/ducks/categories/CategoriesPage.jsx
ptbrowne: Yep this is wanted.

PR #1950, opened 2020-12-09
Notification improvements / fixes

ptbrowne:
- Notifications were not sent when the health bill linked notification
was not set up. This was due to an incorrect check to see if 
notifications were enabled

- Support notification that bring directly to transactions inside an
  account


src/ducks/balance/SetFilterAndRedirect.jsx
```diff
@@ -0,0 +1,71 @@
...
+const mkSetFilterAndRedirect = () =>
+  function SetFilterAndRedirect() {
+    const router = useRouter()
+    const params = useParams()
+    const dispatch = useDispatch()
+    const accounts = useQuery(accountsConn.query, accountsConn)
+    const groups = useQuery(groupsConn.query, groupsConn)
+    useEffect(() => {
+      if (
+        !accounts.fetchStatus === 'loading' ||
```
Crash--, line 28: I have an hard time to understand this condition 🤔

Crash-- (approved):
Just a reserve on a condition but the rest looks good 👍

src/ducks/balance/SetFilterAndRedirect.jsx
ptbrowne: It was a typo. Anyway, after talking to Flo, we concluded that we did not want to keep the condition, since we have push notifications that could be sent without being configured in
settings (for example konnector alerts).

PR #1953, opened 2020-12-11
Improvements

ptbrowne:
- Update cozy-ui
- Put button and divider on ListItems where needed
- Use Dialog for settings edition


src/components/DialogSections/index.jsx
```diff
@@ -0,0 +1,83 @@
+/**
+ * Should be moved inside cozy-ui, see
+ * https://github.com/cozy/cozy-ui/issues/1290
```
Crash--, line 3: In the issue, you say "This will not be done via ModalSections, we prefer now to use Dialog APIs". 

So what should be moved in UI?

src/components/ModalSections/index.jsx
```diff
@@ -4,23 +4,80 @@
...
 
 import React from 'react'
-import { ModalContent } from 'cozy-ui/transpiled/react'
-import Row from 'components/Row'
-import styles from './styles.styl'
+import PropTypes from 'prop-types'
+import { DialogContent } from 'cozy-ui/transpiled/react/Dialog'
+import ListItem from '@material-ui/core/ListItem'
+import ListItemIcon from '@material-ui/core/ListItemIcon'
+import ListItemText from '@material-ui/core/ListItemText'
```
Crash--: We don't need to have Ui's theme on these List components? Should we import theme from UI instead of m-ui?

ptbrowne: Importing them from material ui is not a problem, the theme is applied to them.

Anyway, it's true that it'd be better to be coherent and import them from cozy-ui. I am doing the changre.

src/components/DialogSections/index.jsx
ptbrowne: I removed the comment as I think DialogSections should not be moved to cozy-ui. It's OK if it stays in Banks.

Crash-- (approved):

PR #1955, opened 2020-12-14
feat: Native comm part in debug

ptbrowne:
Added this part to test event going to the native part


Crash--:
Interesting! Have you ever had a case where bridge didn't work? Or is it because you tried to send a message to the bridge before it was ready?

ptbrowne:
> Interesting! Have you ever had a case where bridge didn't work? Or is it because you tried to send a message to the bridge before it was ready?

No, I put the try/catch mostly because I am not sure of the implementation on the other side, not sure if the global objects are there by default for example.

PR #1956, opened 2020-12-18
Smallish improvements

ptbrowne:
One typo fixed
Added two events for tracking


ptbrowne:
Merged directly as it was small.

PR #1961, opened 2021-01-06
fix: Change highlight value theme to buttonTheme property

flohhhh:
- buttonTheme has highlight as default value for Account
  and AddAccountButton components
- Export NoAccount and AddAccountButton components


ptbrowne (approved):
👍  Thanks !

PR #1964, opened 2021-01-07
fix: Display neatly some empty pages

flohhhh:
- Display neatly the follow pages: Reccurences, Categorization and PlannedTransaction
- Simplify simple condition

### DIFF
--- a/src/ducks/categories/CategoriesHeader.styl
+++ b/src/ducks/categories/CategoriesHeader.styl
@@ -45,6 +45,12 @@
     bottom 3.5rem
     left calc(50% - 93px)
 
+.NoAccount_empty
+    position: relative
+    transform: translateZ(0)
+    height: 140px
+    display: flex
+
 .CategoriesHeader__Toggle
     > span
         margin-right 1em
--- a/src/ducks/categories/CategoriesHeader.styl
+++ b/src/ducks/categories/CategoriesHeader.styl
@@ -45,6 +45,7 @@
     bottom 3.5rem
     left calc(50% - 93px)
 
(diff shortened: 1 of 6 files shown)


src/ducks/categories/CategoriesHeader.styl
```diff
@@ -45,6 +45,12 @@
     bottom 3.5rem
     left calc(50% - 93px)
 
+.NoAccount_empty
+    position: relative
+    transform: translateZ(0)
+    height: 140px
+    display: flex
```
ptbrowne, line 52: Why is this styling necessary ?

flohhhh: To display it correctly in the page

flohhhh: if we do not put the style, it display the component in the wrong place

flohhhh: See https://docs.cozy.io/cozy-ui/react/#/Content?id=empty

ptbrowne: OK I think the PR in cozy-ui should be merged so that we have not to do adhoc styling here. Maybe this PR can be mentioned here so that when coming back to the code, we know that the code can be removed ?

flohhhh: Yes it's better indeed. So I will remove the style when PR [#1642](https://github.com/cozy/cozy-ui/pull/1642) will be merged

ptbrowne (approved):

PR #1967, opened 2021-01-08
fix: We can pass any icon name for Empty component

flohhhh:
-  Empty Component has cozy as default value


ptbrowne (approved):
This is for easy overriding I guess

PR #1971, opened 2021-01-14
fix: Remove of  command info about hot reload on app mobile

flohhhh:
Remove yarn watch:mobile:hot (does not exist anymore)
instead use watch: mobile

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -153,7 +153,7 @@ $ yarn ios:run:emulator
 # Replace the host with your own host (find it with ifconfig)
 $ env DEV_HOST=192.168.1.36 yarn build:mobile:hot
 $ yarn ios:run # at this point the app is blank since it cannot access the files from your host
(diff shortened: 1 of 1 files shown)


docs/dev.md
```diff
@@ -153,7 +153,6 @@ $ yarn ios:run:emulator
 # Replace the host with your own host (find it with ifconfig)
 $ env DEV_HOST=192.168.1.36 yarn build:mobile:hot
 $ yarn ios:run # at this point the app is blank since it cannot access the files from your host
-$ env DEV_HOST=192.168.1.36 yarn watch:mobile:hot # launch a webpack-dev-server
```
ptbrowne, line 156: Instead of removing the line, I think it would be better to replace it by 

```
 $ env DEV_HOST=192.168.1.36 yarn watch:mobile # launch a webpack-dev-server
```

don't you think ?

flohhhh: you're right, it's done

ptbrowne (approved):

PR #1972, opened 2021-01-14
fix: Change header for transfer page when there is no bank

flohhhh:
In NoBank component (Transfer page}:

- addition of the nominal header and changing of text by "Transfer"
- The label (with : "Ajoutez une banque pour pouvoir effectuer
 un virement") is added above "Add a bank" button
- Add of spacing between label and button


ptbrowne (approved):
LGTM

PR #1974, opened 2021-01-14
Improvements

ptbrowne:
Each commit is self explanatory.

The most important is the correct deletion of orphan transactions when an account is deleted.

I took the opportunity to rework the GroupPanel into a functional component to reduce component nesting (each HOC added a level of nesting and this makes it harder to debug through React dev tools).


src/locales/en.json
```diff
@@ -154,7 +154,8 @@
     "included": "Included",
     "account-number": "Account number",
     "total-balance": "Total",
-    "deletion_error": "An error occurred while group deletion."
+    "deletion_error": "An error occurred while group deletion.",
```
flohhhh, line 157: There is a rule between underscore and dash naming ? (deletion_error ; no-account)

ptbrowne: No rules, but I think we use the dash-between-words more than underscore_between_words. It could be good to change this.

flohhhh (approved):

src/locales/en.json
ptbrowne: https://github.com/cozy/cozy-banks/pull/1975

PR #1975, opened 2021-01-14
refactor: Use snake-case convention in locales

ptbrowne:
This impacts every page, I have done semi automatically, this should be OK, we
can keep this to merge after a release so that we have time to properly review
all the pages


ptbrowne:
Bon encore quelque fichiers à modifier apparemment (cf CI) :)

flohhhh (approved):

PR #1977, opened 2021-01-15
fix: Fix edit personal info

flohhhh:
- fix: set correctly nationality to initialize page
- feat: adding birthcountry field and adding of test about it
- fix: adding of test about mandatory fields and translation missing
- chore: Adding of optional chaining in babel plugins


ptbrowne (approved):
:1: Thanks !

PR #1981, opened 2021-01-19
fix: Only need to store konnector triggers for konnectorAlerts service

ptbrowne:
Triggers are not only for konnectors and can be used for sharing / notes 
etc... If we do not scope the query, we end up with all the trigger states
inside the io.cozy.bank.settings that tracks the konnector trigger states,
where we only need the states for konnector triggers. In a given cozy, we
stored 173 trigger states, and now we only store 23.

### DIFF
--- a/src/targets/services/konnectorAlerts.js
+++ b/src/targets/services/konnectorAlerts.js
@@ -157,7 +157,11 @@ export const buildNotification = (client, options) => {
  * @return {Promise}
  */
 export const sendTriggerNotifications = async client => {
-  const { data: triggers } = await client.query(Q('io.cozy.triggers'))
+  const { data: triggers } = await client.query(
+    Q('io.cozy.triggers').where({
+      worker: 'konnector'
+    })
+  )
   const konnectorTriggers = triggers.filter(trigger =>
     isKonnectorWorker(trigger.attributes)
   )
--- a/src/targets/services/konnectorAlerts.spec.js
+++ b/src/targets/services/konnectorAlerts.spec.js
@@ -142,7 +142,7 @@ const {
 describe('job notifications service', () => {
   const setup = ({ triggersResponse, settingsResponse, jobsResponse } = {}) => {
     const client = new CozyClient({})
-    client.query = async queryDef => {
+    client.query = jest.fn(async queryDef => {
       const { doctype, id } = queryDef
       if (doctype === 'io.cozy.triggers') {
         return triggersResponse || mockTriggersResponse
@@ -153,7 +153,7 @@ describe('job notifications service', () => {
       } else {
         throw new Error(`No mock for queryDef ${queryDef}`)
       }
-    }
+    })
     client.save = jest.fn()
     client.stackClient.fetchJSON = jest
       .fn()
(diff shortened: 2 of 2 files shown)


flohhhh (approved):

src/targets/services/konnectorAlerts.js
```diff
@@ -157,7 +157,11 @@ export const buildNotification = (client, options) => {
...
 export const sendTriggerNotifications = async client => {
-  const { data: triggers } = await client.query(Q('io.cozy.triggers'))
+  const { data: triggers } = await client.query(
+    Q('io.cozy.triggers').where({
+      worker: 'konnector'
+    })
+  )
   const konnectorTriggers = triggers.filter(trigger =>
     isKonnectorWorker(trigger.attributes)
   )
```
Crash--, line 167: There is no need anymore of filtering on `isKonnectorWorker` since this method checks worker === konnector (https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/trigger.js#L38-L39) 

Also, don't you think you should cleanup the database by removing existing trigger that has nothing to do with konnector? 

Other point (and the last one, promise), should we update the permissions to use a selector worker = konnector here https://github.com/cozy/cozy-banks/blob/master/manifest.webapp#L224 in order to only get access to that part and not the all doctype?

Crash--: Never trust me, an other point... Now we can create a trigger with `worker: konnector` and `type: webhook`. Banks will be one of the first concerned. Does the app will handle correctly the fact we can have 2 triggers (webhook & cron for instance) for the same slug?

ptbrowne: > There is no need anymore of filtering on isKonnectorWorker

You're right, thanks.

> Also, don't you think you should cleanup the database by removing existing trigger that has nothing to do with konnector?

No, the service will clean the triggerStates by itself.

> Other point (and the last one, promise), should we update the permissions to use a selector worker = konnector here https://github.com/cozy/cozy-banks/blob/master/manifest.webapp#L224 in order to only get access to that part and not the all doctype?

I am not sure this works for triggers ? Here it is shown to work for jobs : https://docs.cozy.io/en/cozy-stack/permissions/#selector

> Does the app will handle correctly the fact we can have 2 triggers (webhook & cron for instance) for the same slug?

Since we filter on worker=konnector, I think we are good to go ? Clearly we need to check this but this is related to every app.

Crash--: > Since we filter on worker=konnector, I think we are good to go ?

Nop 

> Clearly we need to check this but this is related to every app.

Yep clearly, but the 💡 comes when reviewing this PR, sorry. 

>I am not sure this works for triggers ? Here it is shown to work for jobs : https://docs.cozy.io/en/cozy-stack/permissions/#selector

It works for each doctype. You can use the selector you want as long as it refers to something in the doctype

ptbrowne: > Since we filter on worker=konnector, I think we are good to go ?

Now : https://github.com/cozy/cozy-banks/pull/1987

PR #1993, opened 2021-01-27
Some fixes abount amount blur and Graph

flohhhh:
- fix: Amount blur was broken
- fix: Graph show most recent months

### DIFF
--- a/src/utils/flag.js
+++ b/src/utils/flag.js
@@ -47,12 +47,12 @@ if (isDemoCozy()) {
 }
 
 Figure.defaultProps = Object.assign({}, Figure.defaultProps, {
-  blurred: flag('amount_blur') ? true : false
+  blurred: flag('amount-blur') ? true : false
 })
 
 flag.store.on('change', function(flagName) {
-  if (flagName == 'amount_blur') {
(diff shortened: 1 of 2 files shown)


src/utils/flag.js
```diff
@@ -47,12 +47,12 @@ if (isDemoCozy()) {
...
 Figure.defaultProps = Object.assign({}, Figure.defaultProps, {
-  blurred: flag('amount_blur') ? true : false
+  blurred: flag('amount-blur') ? true : false
 })
 
 flag.store.on('change', function(flagName) {
-  if (flagName == 'amount_blur') {
-    Figure.defaultProps.blurred = flag('amount_blur')
+  if (flagName === 'amount-blur') {
+    Figure.defaultProps.blurred = flag('amount-blur')
```
ptbrowne, line 55: What was broken ? Where did we use "amount-blur" with a dash ?

ptbrowne (approved):
LGTM, only a small question on what was broken.

src/utils/flag.js
flohhhh: both, the amounts were not blurred

PR #1994, opened 2021-01-28
Fix on unauthorized click

flohhhh:
- fix: Change dateAndGroup to id (from array) key for Transactions component
Allows to have unique key. Because, it can happen to have the same date And Group
- fix: Disable onClick for RowMobile Component if not _id
  - Disable onClick if transaction._id is undefined
  - Active css when transaction._id exist
- chore: Update cozy-harvest-lib from 3.2.0 to 3.7.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,7 +169,7 @@
     "cozy-device-helper": "1.9.2",
     "cozy-doctypes": "1.69.0",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "3.2.0",
+    "cozy-harvest-lib": "3.7.1",
     "cozy-interapp": "0.5.0",
     "cozy-keys-lib": "3.7.0",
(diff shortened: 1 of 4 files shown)


src/components/List/index.jsx
```diff
@@ -39,10 +39,28 @@ export const Header = ({ children }) => (
...
+  />
 )
 
+_Row.defaultProps = {
+  active: true
+}
+
+_Row.propTypes = {
+  active: PropTypes.bool
+}
```
ptbrowne, line 62: Instead of improving our adhoc/custom list, I think we should use MUI/cozy-uis's List / ListItem. The `button` props of ListItems controls whether the ListItem is clickable.

flohhhh: Ok thanks for the information.

It's changed

ptbrowne (approved):

PR #1995, opened 2021-01-29
UX/UI improvements

ptbrowne:
- Use MUI button for buttons in settings
- Use IconButton when possible (back arrow, icons in cozy-bar)
- Use @testing-library/react for tests

### DIFF
--- a/src/components/BarSearchInput/index.jsx
+++ b/src/components/BarSearchInput/index.jsx
@@ -1,6 +1,7 @@
 import React, { useRef, useCallback } from 'react'
 import cx from 'classnames'
 import Icon from 'cozy-ui/transpiled/react/Icon'
+import IconButton from '@material-ui/core/IconButton'
 import styles from './styles.styl'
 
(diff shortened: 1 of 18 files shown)


src/components/BarSearchInput/index.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { useRef, useCallback } from 'react'
 import cx from 'classnames'
 import Icon from 'cozy-ui/transpiled/react/Icon'
+import IconButton from '@material-ui/core/IconButton'
```
flohhhh, line 4: Icon Button should be imported by Cozy UI rather than Material UI, no ? And in Icon Button from Cozy UI, it import Material UI.

ptbrowne: Mmh yes, the problem here is that IconButton in cozy-ui is not a direct export from material UI. I propose to solve the import problem when IconButton is reexported as is from cozy-ui.

flohhhh (approved):

PR #2000, opened 2021-02-08
fix: Set default reducer for mobile object

flohhhh:
Removing of error in console


ptbrowne:
What was the problem ? I remember that there is a warning concerning the persistedState when it's bootstrapped into the redux store, but I can't see clearly how does the change here removes the warning.

flohhhh:
There was a Warning (in red) in console.  The keys from persistedState object must match to object provided in createStore.

ptbrowne:
I do not understand why mobile is set in the persistedState ?

ptbrowne (approved):

PR #2004, opened 2021-02-12
Fix settings and display loader import

flohhhh:


src/ducks/balance/Balance.jsx
```diff
@@ -345,8 +345,16 @@ class Balance extends PureComponent {
     const accounts = this.getAllAccounts()
     const triggers = triggersCollection.data
 
+    const hasNoAccounts =
+      accounts.filter(
+        a =>
+          a._id !== 'health_reimbursements' &&
+          a._id !== 'professional_reimbursements' &&
+          a._id !== 'others_reimbursements'
+      ) === 0
```
ptbrowne, line 354: We do not have a `isVirtualAccount` helper function somewhere ?

src/ducks/balance/AccountsList.jsx
```diff
@@ -43,9 +43,14 @@ export const AccountsList = props => {
     [filterByDoc, router]
   )
 
+  const sortedAndFilteredAccounts = useMemo(
+    () => getSortedAccounts(group, accounts).filter(a => a.status !== 'done'),
```
ptbrowne, line 47: The `status` is an attribute of the `io.cozy.bank.accounts` ? I am a bit surprised by this.

src/ducks/balance/Balance.jsx
flohhhh: Not found, I created it.

ptbrowne: I've checked quickly by grepping on "virtual" and found that normally virtual accounts should have a "virtual" property set to true.

src/ducks/balance/AccountsList.jsx
flohhhh: There is a status when importing accounts. Some accounts are in done, I guess they are accounts that have been added and then deleted.

package.json
```diff
@@ -169,7 +169,7 @@
     "cozy-device-helper": "1.9.2",
     "cozy-doctypes": "1.69.0",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "3.13.1",
+    "cozy-harvest-lib": "3.9.0",
```
ptbrowne, line 172: Pourquoi tu downgrade harvest ?

ptbrowne (approved):
LGTM

PR #2018, opened 2021-02-18
fix: Show correct icons in Accounts and settings page

flohhhh:
The slug for show the konnector icon was provided by the createdByApp.

In the event that the nestor konnector is used for the first time with
a Boursorama account and after we connect with the Boursorama konnector
with the same account.

So the nestor icon was always displayed instead of icon from the
last connected konnector.

In this example it will be boursorama icon.

Now Accounts and settings/accounts page show correctly icons

### DIFF
--- a/src/ducks/account/helpers.js
+++ b/src/ducks/account/helpers.js
@@ -8,6 +8,7 @@ import compose from 'lodash/flowRight'
 import overEvery from 'lodash/overEvery'
 import sumBy from 'lodash/sumBy'
 import get from 'lodash/get'
+import sortBy from 'lodash/sortBy'
 
 import { models } from 'cozy-client'
 import {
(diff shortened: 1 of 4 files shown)


src/ducks/account/helpers.js
```diff
@@ -106,8 +106,18 @@ export const getAccountType = account => {
...
 
-export const getAccountInstitutionSlug = account =>
-  get(account, 'cozyMetadata.createdByApp')
+export const getAccountInstitutionSlug = account => {
+  const updatedByApps = get(account, 'cozyMetadata.updatedByApps')
+
+  if (Array.isArray(updatedByApps) && updatedByApps.length > 0) {
+    const sortedByDate = updatedByApps.sort(
+      (app1, app2) => new Date(app2.date) - new Date(app1.date)
+    )[0]
```
ptbrowne, line 115: I think we should use lodash's sortBy here.

1. Array::sort modifies the array in place, account is modified at the end of this method
2. lodash's sortBy has an easier to use API (not the principal reason)

src/ducks/account/helpers.spec.js
```diff
@@ -426,3 +427,39 @@ describe('buildOthersReimbursementsVirtualAccount', () => {
...
+    {
+      date: '2021-02-18T16:35:54.125Z',
+      slug: 'boursorama83'
+    },
+    {
+      date: '2021-02-18T16:30:46.247Z',
+      slug: 'ce'
+    }
+  ]
+  const account = updatedByApps => ({
```
ptbrowne, line 446: Nit: makeAccount maybe.

ptbrowne: Nit: I personnally like the setup method to have several tests based on the same logic.

const setup = ({ updatedByApps }) => {
  const account = { ... }
  const slug = getAccountInstitutionSlug(account)
  return { slug }
}

it('', () => {
  const { slug } = setup({ updatedByApps: undefined })
  expect(slug).toBe()
})

ptbrowne (approved):

PR #2024, opened 2021-03-01
fix: Wait for flag initialization in onOperationOrBillCreate

ptbrowne:
To properly use flags coming from the server, we need to register the flag plugin and wait for its complete initialization. Here, flags were only coming from the settings document in ad-hoc way (predating the existence of flags stored on the server / context / instance).


flohhhh (approved):

ptbrowne:
I checked locally for test and lint, merging.

PR #2026, opened 2021-03-03
Deduplicate accounts in autogroups service

ptbrowne:
In production, we have seen cases where there were duplicated accounts
in groups. Here, we add the removal of duplicate accounts in the
autogroups service.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -27,11 +27,6 @@ $ export COZY_CREDENTIALS=$(cozy-stack instances token-app cozy.tools:8080 banks
 $ node build/budgetAlerts.js
 ```
 
-Some services like the "budgetAlerts" one expose a CLI that can perform service related tasks.
-
-```
-yarn run services:budgetAlerts --help
-```
 
 ## Services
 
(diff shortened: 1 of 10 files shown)


flohhhh (approved):

docs/services.md
```diff
@@ -83,6 +78,8 @@ Whenever an `io.cozy.bank.accounts` is created, we check if it could belong in a
 on its type (checkings, savings, credit cards). These `io.cozy.bank.groups` documents are created with
 the `auto: true` attributes.
 
+ℹ️  This service an be run via CLI via `yarn service:autogroups`
```
Crash--, line 81: s/an/can

docs/services.md
```diff
@@ -108,6 +105,8 @@ to existing recurrence groups.
 
 See Paper "Paiements recurrents" for more information on the service.
 
+ℹ️  This service an be run via CLI via `yarn service recurrence`
```
Crash--, line 108: can

Crash-- (commented):
If I understand correctly, we fix the issue by deduplicating the accounts, but we don't know yet why these accounts have been duplicated in the first place, right? 

And now, since we deduplicate before adding the account, we will never know the root case, right?

ptbrowne:
Yes and yes.

PR #2027, opened 2021-03-05
Better import in progress

flohhhh:
We want to view when an account import is in
progress on the balance page and in the settings
page. For that we activate the realtime on the jobs
when we are on one of these 2 pages otherwise we
stop it. We also want to display a notification to
the user whether it went correctly or not.

### DIFF
--- a/src/components/JobsContext.jsx
+++ b/src/components/JobsContext.jsx
@@ -1,0 +1,127 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, {
+  createContext,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState
+} from 'react'
+import { JOBS_DOCTYPE } from '../doctypes'
+import CozyRealtime from 'cozy-realtime'
+import logger from 'cozy-logger'
+import isFunction from 'lodash/isFunction'
+
+const log = logger.namespace('import.context')
+
+export const JobsContext = createContext({})
+
+export const useJobsContext = () => {
+  return useContext(JobsContext)
+}
+
+// @TODO add tests for JobsContext
+/** Allows to subscribe to jobs and thus to know jobs in progress
+ *
+ * @param client
+ * @param options
+ * @returns jobsInProgress
+ */
+const JobsProvider = ({ children, client, options }) => {
+  const [jobsInProgress, setJobsInProgress] = useState([])
+  const realtimeStartedRef = useRef(false)
+  const jobsInProgressRef = useRef([])
+
+  const realtime = useMemo(() => {
+    return new CozyRealtime({ client })
+  }, [client])
+
+  const handleRealtime = data => {
+    const { worker, state, message: msg } = data
+
+    // hack: Using jobsInProgressRef because jobsInProgress (from state)
+    // keep its default value
+    // The ideal would be to use only jobsInProgress and do not use the ref
+    const { current: currJobsInProgress } = jobsInProgressRef
+    const index = currJobsInProgress.findIndex(a => a.account === msg.account)
+    const exist = index !== -1
+
+    const { onSuccess, onError } = options
+    let arr = [...currJobsInProgress]
+    if (worker === 'konnector') {
+      if (state === 'running' && !exist) {
+        arr.push(msg)
+      } else if (state === 'done' && exist) {
+        arr.splice(index, 1)
+
+        if (isFunction(onSuccess)) {
+          onSuccess()
+        }
+      } else if (state === 'errored' && exist) {
+        arr.splice(index, 1)
+        if (isFunction(onError)) {
+          onError()
+        }
+      }
+      jobsInProgressRef.current = arr
+      setJobsInProgress(jobsInProgressRef.current)
+    }
+  }
+
+  const startJobsRealtime = () => {
+    if (!realtimeStartedRef.current) {
+      log('info', 'Start Jobs Realtime')
+      realtimeStartedRef.current = true
+
+      realtime.subscribe('created', JOBS_DOCTYPE, handleRealtime)
+      realtime.subscribe('updated', JOBS_DOCTYPE, handleRealtime)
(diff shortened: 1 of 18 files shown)


ptbrowne:
Please take the time to write a PR description :)

src/ducks/balance/ImportGroupPanel.spec.jsx
```diff
@@ -0,0 +1,39 @@
...
+  // it('should displays multiple konnector in progress', () => {
+  //   const jobsInProgress = [
+  //     { konnector: 'caissedepargne1', account: '1111' },
+  //     { konnector: 'caissedepargne1', account: '2222' }
+  //   ]
+  //   const { root } = setup({ jobsInProgress })
+  //
+  //   expect(root.getByText('Importing accounts')).toBeTruthy()
+  //   expect(root.getAllByText('Import in progress').length).toEqual(3)
+  // })
```
ptbrowne, line 38: If you want to keep this, you can use `// @TODO Add multiple konnector tests` and it will create automatically an issue when merging the code.

src/components/JobsContext.jsx
```diff
@@ -0,0 +1,111 @@
...
+  const realtimeStartedRef = useRef(false)
+  const jobsInProgressRef = useRef([])
+
+  const realtime = useMemo(() => {
+    return new CozyRealtime({ client })
+  }, [client])
+
+  const handleRealtime = data => {
+    const { worker, state, message: msg } = data
+    const { current: currJobsInProgress } = jobsInProgressRef
```
ptbrowne, line 36: Pourquoi on a besoin de jobsInProgressRef et en même temps de jobsInProgress ? C'est pas clair pour moi, y-a-t il besoin d'un commentaire ou alors pourrait-on juste utiliser jobsInProgress ?

src/components/JobsContext.jsx
```diff
@@ -0,0 +1,111 @@
...
+    if (worker === 'konnector') {
+      if (state === 'running' && !exist) {
+        arr.push(msg)
+      } else if (state === 'done' && exist) {
+        arr.splice(index, 1)
+        Alerter.success(t('JobsContext.alerter-success'))
+      } else if (state === 'errored' && exist) {
+        arr.splice(index, 1)
+        Alerter.error(t('JobsContext.alerter-errored'))
+      }
```
ptbrowne, line 50: It would be nice if the Alerter logic was provided as callbacks to the JobsProvider, I am sure the JobsProvider can be used in other contexts like in Harvest and this behavior should be configurable by the app.

src/components/JobsContext.jsx
```diff
@@ -0,0 +1,111 @@
...
+      realtime.unsubscribe('created', JOBS_DOCTYPE, handleRealtime)
+      realtime.unsubscribe('updated', JOBS_DOCTYPE, handleRealtime)
+      realtime.unsubscribe('deleted', JOBS_DOCTYPE, handleRealtime)
+    }
+  }
+
+  useEffect(() => {
+    startJobsRealtime()
+    return () => stopJobsRealtime()
+  }, [])
```
ptbrowne, line 81: There is useRealtime hook that could simplify the logic, it does the start/stop/subscribe/unsubscribe for you:

https://github.com/cozy/cozy-ui/blob/db54f27493b6ba61cf41a1b48cf363b9957dd049/react/hooks/useRealtime.js

cozy-ui might not be the good place for it though.

src/components/JobsContext.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <JobsContext.Provider
+      value={{
+        jobsInProgress
```
ptbrowne, line 86: Nit: It could be nice to prepare a bit the context value to make its consumption easier for the components below. For example, indexing the accounts or the triggers, or the connectors.

```
  const value = useMemo(() => ({
    inProgress: jobsInProgress,
    byAccountId: keyBy(jobsInProgress, x => x.account),
    byKonnectorSlug: keyBy(jobsInProgress, x => x.konnector),,
    byTriggerId: keyBy(jobsInProgress, x => x.trigger), // useMemo could be used 
  }, [jobsInProgress])
  return <JobsContext.Provider
    value={value}
  ...
  />
```

The value passed below is here memoized not to incure unnecessary renders (not sure if this is really necessary but it cannot hurt I guess) and indexed so that a component can access info easily.

ptbrowne (commented):
This looks really nice 👏 , the real blocker for me is the lack of test for the JobsContextProvider though.

I have some remarks and nits.

src/components/JobsContext.jsx
flohhhh: Yes I think so too, it would be better in realtime lib

src/components/JobsContext.jsx
flohhhh: Yes I think so too, it would be better in realtime lib

src/components/JobsContext.jsx
ptbrowne: Ah yes, this is because of the handleRealtime closure that is not updated when jobsInProgress changes.

flohhhh: Yes exactly

src/ducks/balance/ImportGroupPanel.jsx
```diff
@@ -0,0 +1,146 @@
...
+
+  useEffect(() => {
+    const slug = account.konnector
+    client
+      .query(Q(KONNECTOR_DOCTYPE).getById(`${KONNECTOR_DOCTYPE}/${slug}`))
+      .then(resp => {
+        const name = resp.data.attributes.name
+        setName(name)
+      })
+  }, [account.konnector, client])
```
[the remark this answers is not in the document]
ptbrowne: From @flohhhh 

> pour le useQuery je l'ai laissé en commentaire et un TODO car lorsque j'avais 2 konnecteurs en cours il y avait le même nom de banque sur chaque item.

ptbrowne (approved):
Thanks !

PR #2036, opened 2021-03-11
fix: Refactor of JobsContext

flohhhh:
- Refactor JobsContext to have the institution label
when there are jobs in progress and also be able to click on konnector
in progress in Settings/Accounts
- Add test about JobsContext
- Show correctly name from konnector at Settings/Accounts page and
ImportGroupPanel


src/components/JobsContext.jsx
```diff
@@ -36,6 +38,22 @@ const JobsProvider = ({ children, client, options }) => {
...
+      )
+      const name = get(resp, 'data.attributes.name')
+      const newJobInProgress = { ...jobInProgress }
+      newJobInProgress.institutionLabel = name
+      return newJobInProgress
+    })
+
+    const newJobsInProgress = await Promise.all(promises)
+    setJobsInProgress(newJobsInProgress)
+  }
```
ptbrowne, line 55: I think this particular business logic could be extracted to another provider, using JobsProvider underneath. This way we could extract JobsProvider for consumption in other non-banking contexts and keep the banking business logic separated from the generic jobs state logic.

ptbrowne: Instead of doing several requests to fetch the individual jobs, it would be better

* either to use Q().getByIds (with an s) to fetch several documents in one request
* or use the relationship mechanism to have the jobs doctype be linked via a relationship to konnectors, and use include(['konnectors']) in the query definition for them to be automatically fetched by cozy-client

src/ducks/context/BanksContext.jsx
```diff
@@ -0,0 +1,69 @@
...
+ * @param client
+ * @returns {JSX.Element}
+ * @constructor
+ */
+const BanksProvider = ({ children, client }) => {
+  const { jobsInProgress = [] } = useJobsContext()
+  const [banksJobsInProgress, setBanksJobsInProgress] = useState([])
+
+  useEffect(() => {
+    const arr = [...jobsInProgress]
```
ptbrowne, line 27: Nit: Why is the copy necessary here ?

src/ducks/context/BanksContext.jsx
```diff
@@ -0,0 +1,69 @@
...
+    if (arr.length > 0) {
+      const formatJobsInProgress = async () => {
+        const ids = arr.map(jobsInProgress => {
+          const slug = jobsInProgress.konnector
+          return `${KONNECTOR_DOCTYPE}/${slug}`
+        })
+        const resp = await client.query(Q(KONNECTOR_DOCTYPE).getByIds(ids))
```
ptbrowne, line 34: Nit: Seems like we will refetch all the running konnectors every time there is a change of running jobs. Not sure this is necessary, an improvement could be to check in the client state if we already have the ids.

```
const idsToFetch = ids.filter(id => !client.getDocumentFromState(KONNECTOR_DOCTYPE, id))
```

src/ducks/context/BanksContext.jsx
```diff
@@ -0,0 +1,69 @@
...
+        const newJobInProgress = resp.data.map(konnector => {
+          const slug = konnector.slug
+          const jobInProgress = jobsInProgress.find(j => j.konnector === slug)
+          return {
+            konnector: slug,
+            account: jobInProgress.account,
+            institutionLabel: konnector.name
+          }
```
ptbrowne, line 42: Nit: I am not sure this is necessary to "commit" to a format so early. This format could be done in the view, and the provider could here only return the konnector/account/institutionLabel.

src/ducks/context/BanksContext.jsx
```diff
@@ -0,0 +1,69 @@
...
+        })
+        return newJobInProgress
+      }
+
+      formatJobsInProgress().then(formatJobs => {
```
ptbrowne, line 47: Nit: queryJobsInProgress would be a bit clearer that we make queries/fetching.

src/ducks/context/BanksContext.jsx
```diff
@@ -0,0 +1,69 @@
...
+        setBanksJobsInProgress(formatJobs)
+      })
+    }
+  }, [jobsInProgress])
+
+  return (
+    <BanksContext.Provider
+      value={{
+        jobsInProgress: banksJobsInProgress,
+        hasJobsInProgress: banksJobsInProgress.length > 0
```
ptbrowne, line 57: I think the `hasJobsInProgress` is not really adding anything over `banksJobsInProgress.length > 0`. I think it would be best to remove the property as it is a sort of duplicte truth, if we forgot to update `hasJobsInProgress` when changing jobsInProgress, we have a problem.

ptbrowne (approved):
Thanks, LGTM, I have some remarks and nits, feel free to take them into account.

src/ducks/context/BanksContext.jsx
flohhhh: I don't think it necessary to delete it even if the added value is slim. 

I do not see in which case we could have a problem, it is also the responsibility of the developer to make changes when necessary.

src/ducks/context/BanksContext.jsx
flohhhh: Thx done

src/ducks/context/BanksContext.jsx
flohhhh: Not necessary, an oversight. Thx

src/ducks/context/BanksContext.jsx
flohhhh: Indeed, it can be optimized.

src/ducks/context/BanksContext.jsx
ptbrowne: AMHA the responsibility of the developer is to design data structures that cannot be in a bad state. Here I have the impression that we explicitly ask for a problem. It would be a different thing if hasJobInProgress was a getter that could not be false.

flohhhh: I do not really understand where here it could be in bad state ? What is the difference between checking in one place if there are items than doing so in several places? Do you have an example where this could be a problem?

ptbrowne: 1. Here we have a structure that contains twice the same information, there is the length of the array and the array itself. I am not noticing a bug in this code, I am just noticing that duplicating the sources of truth leads to bugs.

2. As an API, the additional field makes it easier to have several users of the code that do things differently. Some will use array.length, and some will use the already computed length. I would find it easier to maintain if it was not possible to use the API in two different ways.

flohhhh: ok I see what I mean, thank you for this remark

PR #2042, opened 2021-03-23
Categories improvements

ptbrowne:
Fixes small problems related to categories.

* fix: Category change is reflected in the search page

Before, we would store complete transactions in the state of the search
page. After a recategorization, the transactions were not updated in
the state of the component. Here, we store only the transaction ids
in the results and we retrieve the transactions from the "live"
transactions col that is updated when an item inside is updated.

* Sort sub categories inside category choice dialog

Subcategories were not sorted and this could lead to inconsistencies. Now subcategories are sorted alphabetically, and the "Others" category is last.

* Remove unnecessary flag
* Remove icon sprite

### DIFF
--- a/src/ducks/search/SearchPage.jsx
+++ b/src/ducks/search/SearchPage.jsx
@@ -13,7 +13,6 @@ import { Media, Bd, Img } from 'cozy-ui/transpiled/react/Media'
 import { useI18n, Empty } from 'cozy-ui/transpiled/react'
 import NarrowContent from 'cozy-ui/transpiled/react/NarrowContent'
 
-import flag from 'cozy-flags'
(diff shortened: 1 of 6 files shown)


src/ducks/search/SearchPage.jsx
```diff
@@ -136,9 +137,9 @@ const SearchPage = () => {
         console.log('---') // eslint-disable-line no-console
```
flohhhh, line 137: An oversight (of long-standing) ?

ptbrowne: Thanks, removed

flohhhh (approved):

PR #2045, opened 2021-03-25
feat: Add devtools

ptbrowne:
Add start of a devtool panel that could be reused across apps. It is displayed via a floating-action-button only shown if the flag('debug') is set.

- Add queries viewer that list all queries from cozy-client with their fetch status and data
- Ability to receive devtools panels from the app, this way we can move the debug page from settings to this panel
image

### DIFF
--- a/src/ducks/devtools/Queries.jsx
+++ b/src/ducks/devtools/Queries.jsx
@@ -2,6 +2,7 @@ import React, { useState, useMemo, useCallback } from 'react'
 import sortBy from 'lodash/sortBy'
 import overEvery from 'lodash/overEvery'
 import get from 'lodash/get'
+
 import Grid from '@material-ui/core/Grid'
 import Box from '@material-ui/core/Box'
 import Table from '@material-ui/core/Table'
@@ -25,7 +26,7 @@ import { NavSecondaryAction, ListGridItem } from './common'
 const styles = {
   panelRight: { height: '100%', overflow: 'scroll', flexGrow: 1 },
   mono: { fontFamily: 'monospace' },
-  input: { width: 400 },
+  input: { width: 400 }
 }
 
 const TableCell = withStyles({
@@ -185,14 +186,13 @@ const QueryListItem = ({ name, selected, onClick }) => {
 }
 const QueryPanels = () => {
   const queries = useSelector(state => state.cozy.queries)
-  const [selectedQuery, setSelectedQuery] = useState(
-    () => Object.keys(queries)[0]
-  )
   const sortedQueries = useMemo(() => {
     return sortBy(Object.values(queries), queryState => queryState.lastUpdate)
       .map(queryState => queryState.id)
       .reverse()
   }, [queries])
+
+  const [selectedQuery, setSelectedQuery] = useState(() => sortedQueries[0])
 
   return (
     <>
--- a/src/ducks/devtools/Queries.jsx
+++ b/src/ducks/devtools/Queries.jsx
@@ -39,7 +39,7 @@ const FetchStatus = ({ fetchStatus }) => {
   return (
     <span
       className={
-        fetchStatus == 'loaded'
+        fetchStatus === 'loaded'
           ? 'u-valid'
           : fetchStatus === 'pending'
           ? 'u-warn'
--- a/src/ducks/devtools/Queries.jsx
+++ b/src/ducks/devtools/Queries.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react'
+import React, { useState, useMemo, useCallback } from 'react'
 import sortBy from 'lodash/sortBy'
 import Grid from '@material-ui/core/Grid'
 import Box from '@material-ui/core/Box'
@@ -9,7 +9,7 @@ import ListItem from 'cozy-ui/transpiled/react/ListItem'
 import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
 import { useSelector } from 'react-redux'
 import format from 'date-fns/format'
-import Inspector from 'react-inspector'
+import { TableInspector, ObjectInspector } from 'react-inspector'
 
 import { NavSecondaryAction, ListGridItem } from './common'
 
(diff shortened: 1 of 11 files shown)


src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,140 @@
...
+        </tr>
+        <tr>
+          <td>fetchStatus</td>
+          <td>
+            <FetchStatus fetchStatus={queryState.fetchStatus} />
+          </td>
+        </tr>
+        <tr>
+          <td>lastFetch</td>
+          <td>{format(lastFetch, 'HH:mm:ss')}</td>
```
Crash--, line 130: Hmm, should we use f() from useI18n() or should we use the default format? Does the timezone is useful here?

src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,140 @@
...
+  const { data } = queryState
+  const { lastFetch, lastUpdate } = useMemo(() => {
+    return {
+      lastFetch: new Date(queryState.lastFetch),
+      lastUpdate: new Date(queryState.lastUpdate)
+    }
+  }, [queryState])
+  return (
+    <>
+      <table style={styles.mono}>
```
Crash--, line 117: `<Table>` from UI ? :D

src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,140 @@
...
+      <ListItemText primary={name} secondary={lastUpdate} />
+      <NavSecondaryAction />
+    </ListItem>
+  )
+}
+const QueryPanels = () => {
+  const queries = useSelector(state => state.cozy.queries)
+  const [selectedQuery, setSelectedQuery] = useState(
+    () => Object.keys(queries)[0]
+  )
```
Crash--, line 165: Are they sorted by default? Or should we set the default selected after the sort?

src/ducks/devtools/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+        {panels.map(panelOptions =>
+          currentPanel === panelOptions.id ? <panelOptions.Component /> : null
+        )}
+      </Grid>
+    </Paper>
+  )
+}
+
+const DevTools = ({ panels }) => {
+  const [open, setOpen] = useLocalState('cozydevtools__open', false)
```
Crash--, line 90: 👍

src/ducks/devtools/useLocalState.js
```diff
@@ -0,0 +1,24 @@
...
+      return item !== null ? JSON.parse(item) : initialState
+    } catch (e) {
+      return initialState
+    }
+  })
+
+  const setLocalState = useCallback(
+    newState => {
+      setState(newState)
+      localStorage.setItem(key, JSON.stringify(newState))
```
Crash--: should be in a useEffect() no?

src/ducks/devtools/Queries.jsx
```diff
@@ -48,15 +71,33 @@ const QueryData = ({ data, doctype }) => {
...
+
+  const viewData = results || storeData
+
   return (
     <>
       Table:{' '}
       <input type="checkbox" onClick={handleShowTable} checked={showTable} />
       <br />
+      Search: <input type="text" onChange={handleSearch} value={search} />
```
Crash--, line 95: adding a placeholder telling us how to search can be useful since we split on ";" and ":"

Crash-- (approved):
Excellent! 👏

src/ducks/devtools/useLocalState.js
Crash--: done in an other commit

src/ducks/devtools/Queries.jsx
ptbrowne: Since we do not internationalize anything, f() will not bring anything to the table. The timezone is not important here since we display really recent events in the same timezone as the user.

src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+const QueryData = ({ data, doctype }) => {
+  const client = useClient()
+
+  const documents = useSelector(state => state.cozy.documents[doctype])
+
+  const storeData = useMemo(() => {
+    return data.map(id => client.hydrateDocument(documents[id]))
+  }, [client, data, documents])
```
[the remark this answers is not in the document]
ptbrowne: Not really since documents is the whole `documents` slice (key = doctype, value =documents), whereas in your example `documents` is already an array of documents.
We could do `client.hydrateDocuments(data.map(id => documents[id]))` though but it is not really bringing anything to the table.

src/ducks/devtools/Queries.jsx
ptbrowne: :+1: Done, thanks

src/ducks/devtools/Queries.jsx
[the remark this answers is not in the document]
Crash--: thankd

src/ducks/devtools/Queries.jsx
ptbrowne: We can set the default selected after the sort but it does not change much since queries are updated quickly and soon, the selected query is no longer the most recent.

src/ducks/devtools/Queries.jsx
ptbrowne: Done (from material-ui).

src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,140 @@
+import React, { useState, useMemo } from 'react'
+import sortBy from 'lodash/sortBy'
+import Grid from '@material-ui/core/Grid'
+import Box from '@material-ui/core/Box'
```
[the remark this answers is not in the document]
ptbrowne: I am not sure about that, I find that having one level of indirection between cozy-ui does not really serves us much in practice, it makes it hard to know if we are using a customized version of a MUI component (like we do for example for MuiListItemText) or we if are just re-exporting material ui component.

I may not have understood what you meant by "upgrade / migration path", I cannot see what is useful here for a migration.

src/ducks/devtools/Queries.jsx
```diff
@@ -0,0 +1,196 @@
...
+const styles = {
+  panelRight: { height: '100%', overflow: 'scroll', flexGrow: 1 },
+  mono: { fontFamily: 'monospace' }
+}
+
+const FetchStatus = ({ fetchStatus }) => {
+  return (
+    <span
+      className={
+        fetchStatus == 'loaded'
```
flohhhh, line 27: nit: ===

flohhhh (commented):
Good job 👏🏼

src/ducks/devtools/Queries.jsx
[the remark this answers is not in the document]
Crash--: It's not 100% clear yet, but here are my thoughts: 

- We want to target RN in a near future 
- We have our RN's boilerplate that is looking for in priority for `.native.js(x)` extension before having a fallback on the `.js(x)` file 
- So within an RN context, having `import Table from cozy-ui/transpiled/Table` would be converted first to `from cozy-ui/transpiled/Table.native.js` 
- We can easily add a file to cozy-ui as compared to m-ui 

So having an import from ui will make it easier to add the native part of the component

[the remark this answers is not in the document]
ptbrowne: OK I see, I am not sure cozy-ui should be supporting both React and React native though.

src/ducks/devtools/Queries.jsx
ptbrowne: Done, thanks.

src/ducks/devtools/Queries.jsx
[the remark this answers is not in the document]
ptbrowne: Moving the discussion to https://github.com/cozy/cozy-ui/issues/1787

PR #2058, opened 2021-04-01
Fix banks context

flohhhh:
We check if there are items in jobsInProgress
but if there is none so we have to update with empty
value jobsInProgress

Otherwise in the user interface the last job in
progress remains displayed

### DIFF
--- a/src/ducks/context/BanksContext.spec.jsx
+++ b/src/ducks/context/BanksContext.spec.jsx
@@ -1,0 +1,106 @@
+import React from 'react'
+
+import JobsProvider from '../context/JobsContext'
+import BanksProvider, { BanksContext } from '../context/BanksContext'
+import { render } from '@testing-library/react'
+import CozyClient from 'cozy-client'
+import CozyRealtime from 'cozy-realtime'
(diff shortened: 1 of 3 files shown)


src/ducks/context/BanksContext.spec.jsx
```diff
@@ -0,0 +1,106 @@
...
+                  KONNECTORS[0].account
+                )
+              )
+              handleRealtime(
+                createKonnectorMsg(
+                  RUNNING,
+                  KONNECTORS[1].konnector,
+                  KONNECTORS[1].account
+                )
+              )
```
ptbrowne, line 62: Nit: I think it'd be a bit clearer if konnectors could be passed to the setup method instead of konnectors.

```
// here 
          if (eventName === 'created') {
            for (let konn of konnectors) {
              handleRealtime(
                createKonnectorMsg(
                  RUNNING,
                  konn.konnector,
                  konn.account
                )
            }
              
         }

// in the tests
const { root } = setup({ konnectors: [konnectors[0], konnectors[1]] })
```

ptbrowne (approved):

ptbrowne (approved):

PR #2062, opened 2021-04-07
Fix show banks konnector

flohhhh:
Before we displayed all konnectors in progress, except that it is not necessary.

Now we display only the banking konnector in progress.


ptbrowne (approved):
👍

PR #2069, opened 2021-04-14
Payments Page

flohhhh:
Add payments hide page

This page allows:
1. call a bank connector to authenticate with BI and retrieve a payment object which contains several information including the token.

2. To create a payment object at BI and to record the transaction in bank-operations.

3. To open a webview in order to make a payment.


src/ducks/payments/api.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+  return h
+}
+
+export const biApi = (method, url, payload) => {
+  const config = { method, headers: headers() }
+  if (payload) {
+    config.body = JSON.stringify(payload)
+  }
+  return fetch(new Request(url, config)).then(handleErrors)
```
ptbrowne, line 38: handleErrors does a bit more than handlingError, it also takes care of parsing as a JSON, I think we should find a more generic name, `handleBIResponse` ?

I am not sure that it should be extracted into another method, it looks like it would be simpler to just await the fetch here, and do the `handleErrors` logic directly here.

src/ducks/payments/api.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const getAccessTokenPayload = () => {
+  const biPayment = getBiPayment()
```
ptbrowne, line 42: It's a bit hard to follow I think, where does the payment come from here ?

src/ducks/payments/service.js
```diff
@@ -0,0 +1,75 @@
...
+    client,
+    jobResponse.data.attributes,
+    'result',
+    30 * 1000
+  )
+  return event.data.result
+}
+
+const savePaymentCreation = async (client, paymentCreation) => {
+  const label = get(paymentCreation, 'instructions[0].label')
```
ptbrowne, line 25: Nit: Seems like intrustructions[0] is used everywhere, it could be put into a variable.

ptbrowne (commented):
I think there is room for improvements here, the payment flow logic should be consolidated as to not be scattered across components and tests could be added. Also, I am not sure why is the payment creation saved in localStorage ?

src/ducks/payments/AccessToken.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { useEffect, useState } from 'react'
+import palette from 'cozy-ui/transpiled/react/palette'
+import { LinearProgress } from 'cozy-ui/transpiled/react/Progress'
+import { styles } from './Payments'
+import { getAccessTokenPayload } from './api'
+import Banner from 'cozy-ui/transpiled/react/Banner'
+import Label from 'cozy-ui/transpiled/react/Label'
+import { useClient } from 'cozy-client'
+import { createBiPayment } from './service'
+import { usePaymentContext } from './PaymentContext'
```
Crash--, line 10: nit: Just to note that we're slowly working on improving our import statements. 

first: 
- react
- third party lib 
- cozy-lib (ui/client/ ...)
- local import 

Also I think you should import from relative path since we can't override them easily. 
Instead of 
`import { usePaymentContext } from './PaymentContext'``
write
`import { usePaymentContext } from 'ducks/payments/PaymentContext'`

src/ducks/payments/PaymentValidation.jsx
```diff
@@ -0,0 +1,36 @@
...
+    loadUrlWebView()
+  }, [biPayment, payment.id])
+
+  return (
+    <div>
+      Validate the payment presenting the webview:
+      <div>
+        <ButtonLink
+          href={webViewUrl}
+          target="_blank"
```
Crash--, line 26: Does this functionality will be only available in a web environnement? Or also in Banks mobile? 

Since we open in a new tab, how do we go back to banks?

src/ducks/payments/service.js
```diff
@@ -0,0 +1,75 @@
...
+import { biApi } from './api'
+import { makePayload } from './helpers'
+import { waitForRealtimeEvent } from 'cozy-harvest-lib/dist/services/jobUtils'
+import { TRANSACTION_DOCTYPE } from 'doctypes'
+import get from 'lodash/get'
+
+const log = logger.namespace('payment-service')
+
+export const createBiPayment = async client => {
+  const jobResponse = await client.stackClient.jobs.create('konnector', {
```
Crash--, line 11: It should generate a warning. getStackClient() is better 👍

src/ducks/payments/service.js
```diff
@@ -0,0 +1,75 @@
...
+    })
+  } catch (e) {
+    log('error', "Can't save transaction with payment object")
+    log('error', e)
+  }
+}
+
+export const createPaymentCreation = async ({ client, payment, biPayment }) => {
+  const { url, token, clientRedirectUri } = biPayment
+  const payload = makePayload(payment, clientRedirectUri)
```
Crash--, line 60: OK got it for the previous question. But other question. Where does this `clientRedirectUri` is defined?

src/ducks/payments/AccessToken.jsx
ptbrowne: The eslint plugin for imports could be useful to activate in CCA.

src/ducks/payments/service.js
flohhhh: There is in our global secret database, accessible via the banking connector.

flohhhh:
No action is planned at the moment,  I close.

PR #2073, opened 2021-04-20
fix: Use isQueryLoading

flohhhh:
If fetch status is pending or loading so show spinner, it avoids to try access at "attributes" which is undefined and provokes this error.

For further, I also changed several fetchStatus === 'loading' by isQueryLoading because since this [Cozy Client PR](https://github.com/cozy/cozy-client/pull/925) fetchStatus through pending and after loading.
So it is therefore preferable to use this function which checks if it is pending or loading

Capture d’écran 2021-04-20 à 12 06 31

### DIFF
--- a/src/ducks/settings/HarvestBankAccountSettings.jsx
+++ b/src/ducks/settings/HarvestBankAccountSettings.jsx
@@ -70,16 +70,16 @@ const HarvestLoader = ({ connectionId, children }) => {
               fetchPolicy={fetchPolicy}
             >
               {({ data, fetchStatus, lastUpdate }) => {
-                if (!data) {
-                  return null
-                }
-
                 if (fetchStatus === 'loading' && !lastUpdate) {
                   return <HarvestSpinner />
                 }
 
                 if (fetchStatus === 'error') {
                   return <HarvestError />
+                }
+
+                if (!data) {
+                  return null
                 }
 
                 const { attributes: konnector } = data
--- a/src/ducks/settings/HarvestBankAccountSettings.jsx
+++ b/src/ducks/settings/HarvestBankAccountSettings.jsx
@@ -88,14 +88,19 @@ const HarvestLoader = ({ connectionId, children }) => {
                 // Related issue : https://github.com/cozy/cozy-client/issues/767
                 return (
                   <Query query={Q(TRIGGER_DOCTYPE)} as="triggers">
-                    {({ data: allTriggers, fetchStatus, lastUpdate }) => {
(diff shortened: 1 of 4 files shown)


JF-Cozy (commented):
Ce ne serait pas plutôt à mettre après le spinner ?

C'est normal qu'on veuille retourner null si on a bien fini le fetch, mais qu'il n'y a pas de donnée, plutôt qu'un render de type "désolé aucune donnée bla bla bla" ?

(j'ai pas le contexte en tête désolé)

flohhhh:
> Ce ne serait pas plutôt à mettre après le spinner ?
Oui tu as raison c'est corrigé

flohhhh:
> C'est normal qu'on veuille retourner null si on a bien fini le fetch, mais qu'il n'y a pas de donnée, plutôt qu'un render de type "désolé aucune donnée bla bla bla" ?

C'est plus dans le cas ou data est undefined car dans ces cas là on est pas capable d'afficher quelque chose.
Notamment dans le cas ou c'est en loading ou pending
Capture d’écran 2021-04-20 à 13 12 39

src/ducks/settings/HarvestBankAccountSettings.jsx
```diff
@@ -78,6 +78,10 @@ const HarvestLoader = ({ connectionId, children }) => {
                   return <HarvestError />
                 }
 
+                if (!data) {
+                  return null
+                }
+
```
Crash--, line 84: So if I understand well: 
- We can have an account related to a connector that is not installed anymore. 
- In that case we return null. 

Several remarks: 
- I think Query return an empty array, if there is no result. So data.length === 0 should be a better check 
- What are we displaying if we return null here? What do we want to do in that case? 

Is it possible to add a test to be sure we have the correct behavior?

Crash--:
@flohhhh This is certainly due to the modification I've done in Cozy-Client. 

I don't think the fix you've done is the right one. 

https://github.com/cozy/cozy-banks/blob/e3f98e4db986e2636bb92db51187b812a63af2ee/src/ducks/settings/HarvestBankAccountSettings.jsx#L73

you should also check if fetchstatus === pending 

In fact, you should use `isQueryLoading` as in other queries in Banks ;)

flohhhh:
> In fact, you should use isQueryLoading as is done in other queries ;)

Indeed, it's better

Crash-- (approved):

JF-Cozy (approved):

PR #2074, opened 2021-04-22
chore: Upgrade harvest for new bank

LucsT:
Upgrading harvest to 5.10.1
Add the banque de savoie banks (in banque populaire)


Crash-- (approved):

Crash--:
FYI, this is not "chore". This should be "feat". "chore" is just about CI and other binaries not used in production

PR #2077, opened 2021-05-06
Draft: Stateful notifications

ptbrowne:
To remove the possibility of duplicate notifications, we send a state
to the stack at the same time that we send the notification. When a
notification is declared as stateful in the manifest, the stack will
not send a notification if it has the same state as the previous one.

This PR is in draft as we have to work more on the state that we want
to store. Particularly for balance lower alerts, it might be better
to store all balances for all accounts instead of only the balance
for accounts for which there are changed transactions.

Since we have to work more on the spec, this PR is deprioritized
in favor of the quickfix "not sending the notification for a change
of category".

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -115,8 +115,8 @@
     "balance-lower": {
       "description": "Alert the user when his account balance is lower than a certain value",
       "collapsible": true,
-      "stateful": false,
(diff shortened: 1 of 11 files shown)


ptbrowne:
I still think it could be interesting

PR #2090, opened 2021-05-12
fix: Consider the right settings for notifications

ptbrowne:
This fixes a bug where configurated settings for notifications were 
not taken into account since the service would use the first
io.cozy.bank.settings found; on some cozy this first document could
be the trigger-states document used for konnector alerts notifications.
The service would pick it up and think that nothing was configured.
Hence, default notification settings were used.

Previously, we took the first setting document since we did not have many
settings documents. Now that there trigger-states, pin and more to come, we
cannot expect the configuration settings document to be the right one.

I fixed the bug then refactored so that the service uses the same method
as the UI to find the settings document. This is still a bit fragile, since
to "recognize" the settings document, we look at its shape. We should
at some point migrate to a document with a fixed id like for pin and trigger
states setting document.


JF-Cozy (approved):

src/targets/services/helpers.js
```diff
@@ -15,7 +15,8 @@ const log = logger.namespace('services')
  *
  * @return {Object} an object containing the `documents` and the `newLastSeq`
  */
-export const fetchChangesOrAll = async (Model, lastSeq) => {
+export const fetchChangesOrAll = async (client, doctype, lastSeq) => {
```
Crash--, line 18: jsdoc to update

ptbrowne: Thanks, done

flohhhh (approved):

PR #2091, opened 2021-05-12
Customization

flohhhh:
This PR does not add any direct functionality but allows the possibility of overriding some components as desired in our case.

- Some grouping of components in one, to make the code more readable and easily override
- Several default props overrided

### DIFF
--- a/src/ducks/settings/TabsHeader/index.jsx
+++ b/src/ducks/settings/TabsHeader/index.jsx
@@ -1,0 +1,49 @@
+import React, { useCallback } from 'react'
+
+import { useI18n } from 'cozy-ui/transpiled/react'
+import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
+
+import { PageTitle } from 'components/Title'
+import Padded from 'components/Padded'
+import Header from 'components/Header'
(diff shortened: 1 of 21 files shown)


src/ducks/settings/TabsHeader/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+      </Padded>
+      <Header fixed theme={isMobile ? 'inverted' : 'normal'}>
+        <Tabs value={tabNames.indexOf(defaultTab)}>
+          {tabNames.map((tabName, i) => (
+            <Tab
+              disableRipple
+              classes={{ root: i === 0 && !isMobile ? 'u-ml-2' : 0 }}
+              key={tabName}
+              name={tabName}
+              onClick={goTo(`/settings/${tabName}`)}
```
ptbrowne, line 38: Nit: the tab is re-rendering at each render since the goTo function changes, we could memoize the goTo functions to not hve the problem.

ptbrowne (approved):
👍

src/ducks/settings/TabsHeader/index.jsx
flohhhh: Fixed thx

PR #2092, opened 2021-05-18
Migration modal

flohhhh:
Migration of the last modals in Dialog

### DIFF
--- a/src/ducks/demo/AugmentedModal.styl
+++ b/src/ducks/demo/AugmentedModal.styl
@@ -1,32 +1,35 @@
-.AugmentedModalIntent
-    padding 0
+.ContainerIntentSide
+    display flex
+    height 100%
+    padding 0 !important
 
-.AugmentedModalDescription--NoHeader
-    margin-top 0 !important
-
-.AugmentedModalDescription
-    padding 0!important
-    & > div
-        height 100%
-        & > *
-            box-sizing border-box
-
-            &:last-child
-                flex 0 1 35%
-                padding 1.5rem
+.IntentIframeContainer
+    flex 5
 
 .FakeInfos
     position relative
+    flex 2
+    padding 20px
+    background-color var(--defaultBackgroundColor)
+    overflow auto
 
 .FakeInfosLinkVendor
 .FakeInfosLinkDrive
     position absolute
-    left 0
-    width 125px
+    left 20px
     height 25px
 
 .FakeInfosLinkVendor
-    top 203px
+    width 140px
+    top 225px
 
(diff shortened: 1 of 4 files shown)


src/ducks/demo/AugmentedModal.styl
```diff
@@ -1,32 +1,35 @@
...
+    width 140px
+    top 225px
 
 .FakeInfosLinkDrive
-    top 359px
+    width 104px
+    top 380px
+
+.FakeVeePeeHeader
+    background-color #eee
```
ptbrowne, line 31: Default background color could (--defaultBackgroundColor) be used here.

src/ducks/demo/AugmentedModal.styl
```diff
@@ -1,32 +1,35 @@
...
+    display flex
+    justify-content center
+    width calc(100% - 64px) !important
+    padding 10px 32px !important
```
ptbrowne, line 35: Naive question: why do we have to put hardcoded values and !important.

ptbrowne (approved):

src/ducks/demo/AugmentedModal.styl
flohhhh: we need to override some values ​​concerning the DialogContent and DialogTitle components

ptbrowne: Do we need to have the exact same visual result as before ?

src/ducks/demo/AugmentedModal.styl
flohhhh: Thx I changed the line 13 which contained #F5F6F7 (defaultBackgroundColor), on the other hand i didn't find semantic value about #eee (L31).

src/ducks/demo/AugmentedModal.styl
flohhhh: Yes

ptbrowne: OK then :)

src/ducks/demo/AugmentedModal.styl
ptbrowne: OK, it's weird that we use it then, by any chance, do you know why ? Ideally, when doing a migration it's good to also check with the design/product to see if we can normalize/put in coherence the visuals so that we do not have to do adhoc styling.

Anyway, I understand that this is only for demos, maybe the adhocness is justified. Good for me.

flohhhh: There was a discussion with Ben and he wanted to keep this display. :)

PR #2093, opened 2021-05-18
Render perfs

ptbrowne:
After installing why-did-you-render I saw many problems of re-renders
that could have been avoided. I investigated a bit more and solved
some problems. It improves the performance of the home and transactions
pages; there should be significantly less lagging while scrolling the
transaction list since transaction rows are not re-rendered.

It should also put less pressure on the GC.


Crash-- (approved):
Nice 🎉

PR #2095, opened 2021-05-20
feat: Upgrade harvest

flohhhh:
Allows you to have certain requested features, in particular :
- feat: Remove references to "cozy" in harvest locales
Makes it easier to customize and whitelabel the UI
- Show konnector error in both data and configuration tab in harvest
- Open external links in a new tab not to lose the Cozy context

BREAKING CHANGES
When using the TriggerManager outside the main Harvest Routes, you need to import { IntentTriggerManager } from 'cozy-harvest-lib', instead of `import { TriggerManager } from 'cozy-harvest-lib'.

For more information on features: 
[CHANGELOG](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/CHANGELOG.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "cozy-device-helper": "1.12.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "5.10.1",
+    "cozy-harvest-lib": "6.1.0",
     "cozy-interapp": "0.6.2",
     "cozy-keys-lib": "3.8.0",
     "cozy-konnector-libs": "4.30.3-beta.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -173,7 +173,7 @@
     "cozy-device-helper": "1.12.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "5.10.1",
+    "cozy-harvest-lib": "6.1.0",
```
ptbrowne, line 176: As it is a breaking change, can you describe how/why it does not affect banks ?

ptbrowne (commented):
Looks OK but since it is a major update, I would like to have a little more info.

Crash--:
It seems we had an issue with the 6.0.0 publication 🤔 We shouldn't have this BC since it's a pretty old commit.

https://github.com/cozy/cozy-libs/compare/cozy-harvest-lib@5.10.6...cozy-harvest-lib@6.0.0

Really don't know why we had all this diff in the changelog

PR #2096, opened 2021-05-20
fix: Blank screen on store (incorrect url)

flohhhh:
fix: Blank screen on store (incorrect url)

useRedirectionURL returns a tuple value
not just an url.
The url used was therefore not usable as such


Crash--:
here https://github.com/cozy/cozy-banks/blob/master/src/ducks/transactions/TriggerErrorCard.spec.jsx#L7 and here https://github.com/cozy/cozy-banks/blob/master/src/ducks/balance/AccountsImporting.spec.jsx#L11  it seems we do not return an array. We should either update them if needed or delete mock if they are not necessary anymore

Crash-- (approved):

JF-Cozy (approved):

PR #2098, opened 2021-05-24
fix: Add missing query string

flohhhh:
fix: Add missing query string to show only pending update

Before we display all the store's bank connectors, now only those that are awaiting update

### DIFF
--- a/src/components/KonnectorUpdateInfo/index.jsx
+++ b/src/components/KonnectorUpdateInfo/index.jsx
@@ -58,7 +58,7 @@ const KonnectorUpdateInfo = ({ outdatedKonnectors }) => {
             className="u-mh-0"
             label={t('KonnectorUpdateInfo.cta')}
             icon="openwith"
-            href={url}
+            href={`${url}&pendingUpdate=true`}
           />
         }
         title={t('KonnectorUpdateInfo.title')}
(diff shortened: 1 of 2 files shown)


src/components/KonnectorUpdateInfo/index.jsx
```diff
@@ -58,7 +58,7 @@ const KonnectorUpdateInfo = ({ outdatedKonnectors }) => {
             className="u-mh-0"
             label={t('KonnectorUpdateInfo.cta')}
             icon="openwith"
-            href={url}
+            href={`${url}&pendingUpdate=true`}
```
Crash--, line 61: Just searched in the store, I didn't find anything related to this querystring. Can you tell me more about it?

Crash-- (approved):

src/components/KonnectorUpdateInfo/index.jsx
Crash--: J'ai approvué un peu vite : 
Tu as le pendingUpdate en argument ici https://github.com/cozy/cozy-banks/blob/4f06282d18f26b2ea96a258dcd0b9c3a02c29439/src/components/KonnectorUpdateInfo/index.jsx#L27 

Du coup c'est plus le hook ici https://github.com/cozy/cozy-banks/blob/master/src/hooks/useRedirectionURL.js#L8 à corriger

PR #2099, opened 2021-05-24
fix: Pass pendingUpdate in useRedirectionURL

flohhhh:
fix:  Pass pendingUpdate in useRedirectionURL to show only pending update

Before we display all the store's bank connectors, now only those that are awaiting updateto show only pending update


src/components/KonnectorUpdateInfo/index.spec.jsx
```diff
@@ -22,15 +22,15 @@ describe('KonnectorUpdateInfo', () => {
...
     ])
     const { root } = setup({
       outdatedKonnectors: { data: [{ categories: 'banking' }] }
     })
 
     const link = root.getByText('Update my banks').closest('a')
     expect(link.getAttribute('href')).toEqual(
-      'http://store.cozy.tools:8080/#/discover/?type=konnector&category=banking',
+      'http://store.cozy.tools:8080/#/discover/?type=konnector&category=banking&pendingUpdate=true',
       () => {}
```
Crash--, line 34: This test doesn't test much, no? We expect that the URL returned by useRedirectionURL is well displayed in the a href. 

But the setup `{categories: 'banking'}`is useless, no? This setup can be interesting when you test the `useRedirectionURL` hook since you should test that the args passed to it are present in the response.

Crash-- (approved):

PR #2105, opened 2021-05-28
fix: Hack to avoid slower transition in tabs

flohhhh:
in the case of override where we add a navigation
with tabs that have an animation. The user has an
impression of slowness / lag. To avoid this, we
use a delay which allows the content to be
 rendered a few ms later.

By default in banks There's no delay

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -52,6 +52,7 @@ import { filterByAccounts } from 'ducks/filters'
 import { trackPage } from 'ducks/tracking/browser'
 import { isVirtualAccount } from 'ducks/balance/helpers'
 import ImportGroupPanel from 'ducks/balance/ImportGroupPanel'
+import Delayed from 'components/Delayed'
 
 const syncPouchImmediately = async client => {
   const pouchLink = client.links.find(link => link.pouches)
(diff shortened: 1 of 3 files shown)


JF-Cozy (commented):
J'ai pas du tout le projet en tête mais on rajoute un délai dans une fonction qui s'appelle `syncPouchImmediately` ça pose pas de souci ailleurs, alors qu'on s'attendrait à ce que ce soit immédiat ?

Et set par défaut delayContent: 0 partout, j'imagine que c'est en ms et que 0 veut dire "pas de délai" ? Du coup ou est-ce qu'on rajoute du délai ?

ptbrowne (commented):
I actually am not sure that Delayed is helpful :/ 

I think it hides problems rather than solving them. But then if we really think this is a good solution, why not.

As for wrapping individually each page component, I think we want this delay to potentially be applied to all the pages, and I think it might have its place directly in the AppRoute file where we can see all the pages.
There is already a scrollToTopOnMount wrapper on all the pages, maybe we can have a `page()` HOC  that does both the scrollToTopOnMount and the Delayed wrapping ? I think it'd be easier to maintain if everything was in the same place, since there would be less files changed and every page would be auditable at a central point. What do you think ?

flohhhh:
Yes we agree on this solution, it does not solve the problem but bypasses it.

We cannot put a HOC there where you indicate it because we want to hide some content and not the whole page. This will hide the whole page and this is not what we want. You still want to access navigation when you override it.

PR #2106, opened 2021-05-31
Fix recurrence

flohhhh:
The recurrence service no longer worked. (Otherwise we have an error of the type: Cannot read property 'label' of undefined)

Before we filtered if there were no upstream operations then on updated the recurrences
```
const recurrences = (await fetchHydratedBundles(client)).filter(
      x => x.ops.length > 0
    )

findAndUpdateRecurrences(...)
```
Now we do it after, so it's necessary to check if there is at least one value in ops.
Because access to a property is impossible if the array is empty.


JF-Cozy (approved):

JF-Cozy:
J'ai approuvé un peu vite, peut-être rajouter un test vite fait ?

ptbrowne (approved):

Issue #2101, opened 2021-05-25
Lighter transaction requests


ptbrowne:
Right now, all transactions are fetched at the start of the Banks app, and they can be heavy (~12mo for 20000 transactions, corresponding to 5 transactions per day for 10 years).

Instead of fetching everything we have to fetch only what is needed.

- [x] Home page

For the chart, we can only fetch the last year and fetch only relevant attributes from transactions (date, amount, account, currency).

- [x] Transaction page

We can fetch transactions month per month taking into account the group/account filter that is set

- [ ] Search page

Fetch the last year of transactions with labels and search into that. If no results are found, provide a way to load up more years.

- [ ] Reimbursement page

Last year of transactions with categoryId. Provide a button to load up more years.

- [ ] Category page

Full transactions can be loaded month per month.

PR #2107, opened 2021-06-01
Improve data fetching app wide

ptbrowne:
This is a sequel to the meta issue https://github.com/cozy/cozy-banks/issues/2101.

In this large PR, we gain incremental queries for the transactions page and the categories page.

## Transaction page

For the transaction page, queries are based on the account/group filter to only fetch transactions pertaining to those group. When the user jump from month to month, new queries are created that start at the selected month. Infinite fetching is implemented via the fetchMore function for those queries. Since we do not want data from other months to disturb one particular month, queries on the transactions pages are created with the query options { add: false }. For example, this prevents transactions that have been fetched through the january query to end ud in the march query. This would prevent the infinite loading mechanism to work correctly (since we rely on a graphical element to trigger the fetchMore).

Example of queries that are made

- `transactions-<accountId>`

Transactions for a particular account when the user has not yet navigated in time. Transactions are fetched sorted by date, so we end up with the most recent transactions for a particular account. When the user scrolls down and we have more transactions (query.hasMore), we can request more transactions through `query.fetchMore`

- `transactions-<accountId>-<endDate>`

When the user navigates from month to month via the date selector, we create a new query that contains the ending month ({ date: { $gt: endDate }}). This enables us to support the infinite loading through the previous mechanism of fetchMore.

- `transactions-<groupId>`

When the user selects a group. Very similar to the `transactions-<accountId>` query but we fetch transactions for several accounts with a CouchDB `$or` query on the `account` attribute. Unfortunately this kind of query is an order of magnitude slower than `account` queries (between 150 and 500ms), but no solution has been found yet.

- `transactions-<groupId>-<endDate>`

Similar to the `transactions-<accountId>-endDate>`, this is when the user has filtered by a group and has navigated to a month via the date selector.

## Date selector

For the date selector to work, we need to have the full extent of dates that is available for the particular account/group that we have chosen. Here, we make 2 queries, one for the earliest transaction and one for the latest transaction and then we generate all the months between those two trnsactions to display the date selector options.

Those queries are indexed by date and account and are quick (30ms).
While those queries are in flight, the date selector is displayed but not functional.

## Categories page

For this page, we do not need to support infinite loading since we either stay in 1 particular month/year. Thus we end up with queries like `transactions-<accountId/groupId>-<startDate>-<endDate>` = queries only for the current month transactions, filtered by account and group.

## Keeping previous data while loading next one

To prevent a jarring loading screen while loading the month, the previous query's data
is shown while the next query data is loading. While data is loading, a progress bar
is shown below the transaction header.


src/components/AccountIcon/index.jsx
```diff
@@ -23,6 +24,9 @@ const _AccountIcon = ({ account, className, size }) => {
   if (!institutionSlug) {
     return null
   }
+  if (flag('banks.perf.deactivate-account-icons')) {
```
flohhhh, line 27: Can you explain why we need these flags?

ptbrowne: Yep, this are development flags to check performance to check if perfomance improves if we show/hide account icons. I suspect account icons to have an impact since they trigger many requests and are not properly cached.

I can remove them since they are not necessary for this particular PR.

ptbrowne: I've removed those flags.

src/hooks/useIntersectionObserver.jsx
```diff
@@ -0,0 +1,19 @@
...
+  const [ref, setRef] = useState(null)
+  const handleRef = useCallback(node => {
+    setRef(node)
+  }, [])
+
+  useEffect(() => {
+    if (!ref) {
+      return
+    }
+    const o = new IntersectionObserver(handleIntersection, observerOptions)
```
Crash--, line 13: Does Banks already includes the polyfill? 

IntersectionObserver are only available on iOS 12.1+ (https://developer.mozilla.org/fr/docs/Web/API/IntersectionObserver). Banks supports iOS 11, EPA 12.0

ptbrowne: Thanks, I forgot about the iOS 11 limit. Done in https://github.com/cozy/cozy-banks/pull/2107/commits/b8c57aab9f369201839def45946091be68be794c

Crash-- (approved):
Great work!

src/components/AccountIcon/index.jsx
flohhhh: Thx for the explanation

PR #2119, opened 2021-06-15
fix: Do not crash when groups have not yet been loaded

ptbrowne:
If groups have not yet been loaded, groupsCollection.data is null and
cannot be spread.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "cozy-device-helper": "1.12.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "6.2.0",
+    "cozy-harvest-lib": "6.3.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -175,7 +175,7 @@
     "cozy-device-helper": "1.12.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "6.2.0",
+    "cozy-harvest-lib": "6.3.0",
```
flohhhh, line 178: what more in this version?

ptbrowne: It uses deep imports and fixes the crash in Chrome while developing due to the pdf js worker.

flohhhh: Thx

flohhhh (approved):

PR #2136, opened 2021-06-28
fix: Double request sent for transactions

ptbrowne:
Flo noticed a double request to fetch operations on the balance page.

The selector had a date with milliseconds, so the selector was changing 
between two requests, and since the selector is used to deduplicate queries
in cozy-client, queries were not de-duplicated.

I could not find why query() was called twice in the first place, it
seems that the Balance component is mounted/unmounted at some point
since the useQuery() hook changes identity at some point.


Crash--:
> seems that the Balance component is mounted/unmounted at some point

This can be problematic since we've a `trackPage` event on every mount, no? Also, this `Balance` component is pretty big, having a mount / unmount / mount must be expensive. Maybe this is something we should track and invest time to understand and fix.

Crash-- (approved):

ptbrowne:
trackPage does not send twice the same page.

Regarding expensiveness of the mount/unmount, I am not sure it is so expensive since both requests were fired with 3 milliseconds of difference. I will try to keep this in mind, but I am not sure how to debug that (if you have any idea ?) and am afraid that the investment will not be worth the time.

PR #2141, opened 2021-06-30
Better mobile docs

ptbrowne:
Update the docs to reflect the workflow we have now to publish
Android and iOS apps.

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -22,7 +22,7 @@
     - [Fastlane](#fastlane)
     - [iOS](#ios)
       - [Signing](#signing)
-      - [Push iOS build](#push-ios-build)
+      - [iOS build](#ios-build)
     - [Android](#android)
 - [Notifications](#notifications)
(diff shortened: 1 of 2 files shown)


docs/dev.md
```diff
@@ -298,39 +298,47 @@ The certificate with its password is stored in the shared password store.
...
-yarn run android:publish
-```
 
-This will upload create a new version in the store, in the `beta` track that you can review visually on the Play Store and then promote to the normal track.
+#### Android
 
+Anytime a beta is tagged, Travis produces an APK in the Android stage.
+In the logs is displayed an URL for the APK that has been built.
+This APK can then be tested and then pushed and published in the Play
+Developer Console.
```
Crash--, line 341: Does that mean that your travis script is updating the android-version code in the config.xml?

Crash-- (approved):

docs/dev.md
ptbrowne: No, you have to edit the config.xml file manually.

PR #2148, opened 2021-07-01
Fix style for transactions row

JF-Cozy:
Ajout d'un trait de séparation entre les transactions + décalage des actions

before :
after :


flohhhh (approved):

Merkur39 (approved):

src/ducks/transactions/TransactionRow/TransactionRowMobile.jsx
```diff
@@ -51,7 +52,8 @@ const TransactionRowMobile = ({
   showRecurrence,
   isSelected,
   isSelectionModeActiveFn,
-  toggleSelection
+  toggleSelection,
+  hasSeparator
```
ptbrowne, line 56: I think it would have been a bit better to use the same wording as the component: s/hasSeparator/hasDivider to try not to add too much vocabulary.

JF-Cozy: Effectivement, bien vu 👍

PR #2154, opened 2021-07-06
fix: Wrong order of transactions for credit card accounts

ptbrowne:
Transactions should be re-sorted after coming from the stack since we
display the realisationDate for credit cards. If transactions are sorted
"only" by date, the order of displayed transactions seems wrong since the
date that the user sees is the realisationDate. For delayed debit card
accounts, the realisation date can be at the beginning of the month whereas
the date is at the end of the month.

After the switch to incremental queries, the wrong transactions were used,
here we switch to using transactions from before, transactions that are
sorted by displayed date.

The test has to be updated to account for that since now even in the UI 
test, transactions are sorted by displayed date. Since only the top 
transactions are really displayed, we could not click on transactions that
were at the bottom.

### DIFF
--- a/src/ducks/transactions/Transactions.jsx
+++ b/src/ducks/transactions/Transactions.jsx
@@ -216,7 +216,6 @@ export class TransactionsDumb extends React.Component {
       showTriggerErrors,
       filteringOnAccount,
       className,
-      transactions,
       TransactionSections,
       onReachTop,
       onReachBottom,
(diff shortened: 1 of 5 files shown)


src/ducks/transactions/Transactions.jsx
```diff
@@ -225,7 +224,7 @@ export class TransactionsDumb extends React.Component {
 
     return (
       <>
-        <SelectionBar transactions={transactions} />
+        <SelectionBar transactions={this.transactions} />
```
JF-Cozy, line 226: Je comprends pas comment `this.transactions` n'est pas `undefined`... ?

ptbrowne: Avec le componentWillMount

JF-Cozy: ah pardon en effet !

JF-Cozy (approved):

flohhhh (approved):

PR #2169, opened 2021-07-13
feat: Put skeleton behavior for balance account rows behind a flag

ptbrowne:
We are not sure if we want to keep the account row skeleton behavior
(account rows on home page are displayed with "loading" if they 
are "below the fold"). This was done to improve performance on mount
of the home page, but it seems that it can be a bit janky and
the loading state disappears too much time after the user has
scrolled to the element.

Here, this behavior is disabled: all account rows are "visible" by
default. If we want to reput this behavior or improve on it, it
will be easy to remove the flag.


ptbrowne:
cc @flohhhh and @fffflo

PR #2175, opened 2021-07-20
fix: Don't use ListSecondaryAction

Crash--:
If there is no action attached to the content.
If you want to display an icon aligned to the right
let's just use the icon directly


flohhhh:
There is no visual impact on the UI ?

Crash--:
Capture d’écran 2021-07-20 à 14 31 09
Capture d’écran 2021-07-20 à 14 31 18

flohhhh (approved):

PR #2179, opened 2021-08-02
Fix/add test case

flohhhh:
- Add test case to check frequency in this case
The objective is to check that the new operation (mayTransaction) has been 
attached to the bundle and that this frequency remains "each month" after the recurrency service is passed.

- refactor: use getT(option) instead of new Polyglot
Allows to facilitate the recovery of t during the
tests

### DIFF
--- a/src/ducks/recurrence/fixtures/fixtures6.json
+++ b/src/ducks/recurrence/fixtures/fixtures6.json
@@ -1,0 +1,58 @@
+{
+  "io.cozy.bank.operations": [
+    {
+      "_id": "february",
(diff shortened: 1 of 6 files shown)


JF-Cozy:
> Add test case to check frequency in this case

C'est quel cas spécifiquement qu'on souhaite tester ici ?

flohhhh:
> > Add test case to check frequency in this case
> 
> C'est quel cas spécifiquement qu'on souhaite tester ici ? Quel est le contexte ?

Done

PR #2182, opened 2021-08-03
feat: Use client.saveAll to save transaction categories

ptbrowne:
Depends on https://github.com/cozy/cozy-client/pull/995

Now that cozy client has saveAll, we use it instead of calling
repeatedly save.

## Failure

Batch edits can fail because of conflicts (a version of a document has changed without cozy-client knowing about it) or document being deleted.

In case of failure, operations being recategorized are refetched to mitigate the chance of a future conflict.

Here is a video where I try to recategorize 3 operations and I force a conflict

- One of the operations is changed "behind the back" of cozy-client, so the version that cozy-client has is stale
- We can see that the first batch edit fails somehow (but two operations have been recategorized correctly)
- When the failure happens, operations are refetched and we see that cozy-client now has the correct version for the 3 operations
- The subsequent batch edit works since cozy-client has the latest version.

https://user-images.githubusercontent.com/465582/128198486-6b1b2eca-b00f-42c4-9965-24d534c7d16b.mp4


src/ducks/transactions/TransactionModal/TransactionCategoryEditorDialog.jsx
```diff
@@ -26,7 +26,7 @@ const TransactionCategoryEditorDialog = ({ transaction, onClose }) => {
   return (
     <TransactionCategoryEditor
       beforeUpdates={handlePop}
-      afterUpdate={onAfterUpdate}
+      afterUpdates={onAfterUpdate}
       onCancel={handlePop}
       transactions={[transaction]}
     />
```
JF-Cozy, line 32: nit: On peut peut-être renommer `onAfterUpdate` en  `onAfterUpdates` ?

ptbrowne: Done

src/ducks/transactions/TransactionCategoryEditor.jsx
```diff
@@ -1,58 +1,59 @@
...
 
-      await Promise.all(promises)
+      try {
+        await client.saveAll(newTransactions)
+      } catch (e) {
+        const qdef = Q(TRANSACTION_DOCTYPE).getByIds(
+          newTransactions.map(t => t._id)
+        )
+
+	// eslint-disable-next-line no-console
```
flohhhh, line 42: nit: Indent comment

ptbrowne: Done.

JF-Cozy (approved):

Issue #2181, opened 2021-08-03
Banks does not work on mobile


ptbrowne:
It seems like the introduction of 3a7d27a3b breaks the mobile
build. When building for mobile and launching on a native device (at the moment I have tested through a virtual device via Xcvode), the banks page remains blank and in the console I have a "parse error: unexpected token undefined" due
to root.dataset.cozy being undefined on mobile.

I reverted temporarily to continue working on my feature.

PR #2183, opened 2021-08-03
fix: Crash on mobile version

flohhhh:
Fix [issue #2181](https://github.com/cozy/cozy-banks/issues/2181)

In mobile version we don’t have dataset and so we
can’t parse this data.
Now we check root.data.cozy value before
to parse it.

### DIFF
--- a/src/utils/lang.spec.jsx
+++ b/src/utils/lang.spec.jsx
@@ -1,0 +1,16 @@
+import { getLanguageFromDOM } from './lang'
+
+describe('lang', () => {
+  it('should return "en" locale as a default', () => {
+    const lang = getLanguageFromDOM()
+    expect(lang).toEqual('en')
+  })
+
+  it('should rely on the data-cozy field if it exists', () => {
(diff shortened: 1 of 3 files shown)


Crash-- (approved):

src/utils/lang.spec.jsx
```diff
@@ -0,0 +1,16 @@
+import { getLanguageFromDOM } from './lang'
+
+describe('lang', () => {
+  it('should return "en" locale', () => {
```
ptbrowne, line 4: I think it'd be clearer with "should return "en" locale as a default"

src/utils/lang.spec.jsx
```diff
@@ -0,0 +1,16 @@
...
+    const lang = getLanguageFromDOM()
+    expect(lang).toEqual('en')
+  })
+
+  it('should return "fr" locale', () => {
```
ptbrowne, line 9: I think it'd be clearer with "should rely on the data-cozy field if it exists"

PR #2194, opened 2021-08-10
feat: Warmup queries to have indexes on account / date / date & account [apk]

ptbrowne:
Add warmup queries to populate the most important indexes.

It improves for example the loading time of the analysis page in the following scenario:

- Pouch has finished its replication for the transaction doctype
- When going to the analysis page, a query for the current account(s) transactions is fired
- Pouch has not yet indexes for the given query and it takes a long time to build them
- So while the index is being built, the user sees a spinner

With warmed up queries, the cozy pouch link does not consider itself ready while the warm up queries have not returned. So it will pass the query down to the next link = stack link here. It means that while the index is being built, the stack is responding to the query. After the warm up query succeed, it means the index has been built, and now the cozy pouch link is active, and it responds faster.

I've added a warm up query for the "date" index in a previous PR, but it turns out that for other use cases (listing transactions for a particualr account or group of accoutns), we need also the date+account and account.

I've added a bit of documentation here : https://github.com/cozy/cozy.github.io/pull/213

### DIFF
--- a/src/ducks/client/links.js
+++ b/src/ducks/client/links.js
@@ -1,5 +1,6 @@
 /* global __POUCH__ */
 
+import fromPairs from 'lodash/fromPairs'
 import { StackLink, Q } from 'cozy-client'
 import { offlineDoctypes } from 'doctypes'
 import { isMobileApp, isIOSApp } from 'cozy-device-helper'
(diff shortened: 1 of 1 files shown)


JF-Cozy (approved):

flohhhh:
What does it improve?

ptbrowne:
Sorry, my description was too succinct, I've edited it.

PR #2238, opened 2021-09-22
fix: Update Cozy UI with Sprite excluded from Icon

flohhhh:
See [PR#1908](https://github.com/cozy/cozy-ui/pull/1908)

This update reduces the size of the bundle.

Before:
before

After:
after


Crash--:
-140KB 

Go go go :rocket:

Crash-- (approved):

PR #2239, opened 2021-09-23
fix: Use correct css var color for warning and success

JF-Cozy:
introduced by https://github.com/cozy/cozy-banks/pull/2221
We didn't make all the changes to the BC in v49 of cozy-ui, but since the old version of the cozy-bar was using a version prior to v49 of cozy-ui this was not a problem. Now this is not the case anymore.

⚠️ à rapatrier sur Master https://github.com/cozy/cozy-banks/pull/2240

### DIFF
--- a/src/ducks/balance/AccountRow.styl
+++ b/src/ducks/balance/AccountRow.styl
@@ -8,7 +8,7 @@
 
 .AccountRow--hasWarning
     .AccountRow__figure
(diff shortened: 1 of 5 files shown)


Crash--:
linked to https://github.com/cozy/cozy-banks/pull/2053 also

Crash-- (approved):

PR #2248, opened 2021-09-27
fix: Search when clicking on the magnifying glass

Crash--:
If the transaction label contained a "/" then the search did not work
since the character was not escaped.

### DIFF
--- a/src/ducks/search/SearchPage.jsx
+++ b/src/ducks/search/SearchPage.jsx
@@ -44,7 +44,9 @@ const SearchPage = () => {
   const { t } = useI18n()
   const { isMobile } = useBreakpoints()
 
-  const [search, setSearch] = useState(params.search || '')
+  const [search, setSearch] = useState(
+    params.search ? decodeURIComponent(params.search) : ''
+  )
   const [resultIds, setResultIds] = useState([])
 
   useTrackPage('recherche')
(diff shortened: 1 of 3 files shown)


flohhhh (approved):

src/ducks/search/SearchPage.jsx
```diff
@@ -44,7 +44,9 @@ const SearchPage = () => {
   const { t } = useI18n()
   const { isMobile } = useBreakpoints()
 
-  const [search, setSearch] = useState(params.search || '')
+  const [search, setSearch] = useState(
+    params.search ? decodeURIComponent(params.search) : ''
```
flohhhh, line 48: Can you add a simple test plz ?

Crash--: What do you want me to test?

flohhhh: By checking that the content is correctly put in the href (${encodeURIComponent(label)}) and correctly displayed in the search page?

Crash--: So testing encodeURIComponent and decodeURIComponent? And testing react-router? I'll not do that ^^' 

There is already a test to check that what is in the URL in used to make the search. I personally think that it's enough

flohhhh: It wasn't really the idea. But rather to check that if a transaction's label has a "/" (for example), so it must be encoded and not displayed in raw in order to avoid this kind of situation.

PR #2284, opened 2021-11-12
fix: Local categorization

paultranvan:
See https://github.com/konnectors/libs/pull/816

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "cozy-harvest-lib": "6.10.0",
     "cozy-interapp": "0.6.2",
     "cozy-keys-lib": "3.8.0",
-    "cozy-konnector-libs": "4.30.3-beta.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -173,7 +173,7 @@
     "cozy-harvest-lib": "6.10.0",
     "cozy-interapp": "0.6.2",
     "cozy-keys-lib": "3.8.0",
-    "cozy-konnector-libs": "4.30.3-beta.1",
```
Crash--, line 176: I don't like that. Are we sur that what we had in the beta was merged?

paultranvan: This was introduced by https://github.com/cozy/cozy-banks/pull/1818 , which is two-years old, and already related to the categorization. So I assumed we should be good?

flohhhh (approved):

PR #2291, opened 2021-11-29
Feat/show alert error

flohhhh:
Show alert message when a request failed (allows to reload the app)


src/components/App.jsx
```diff
@@ -1,14 +1,16 @@
 import React, { useEffect } from 'react'
 
 import flag from 'cozy-flags'
-import { queryConnect } from 'cozy-client'
+import { queryConnect, useClient } from 'cozy-client'
 import { withRouter } from 'react-router'
 import compose from 'lodash/flowRight'
+import throttle from 'lodash/throttle'
```
JF-Cozy, line 7: nit: on pourrait en profiter pour réorganiser les imports

JF-Cozy (approved):

package.json
```diff
@@ -165,7 +165,7 @@
     "cozy-authentication": "2.1.0",
     "cozy-bar": "^8.3.0",
     "cozy-ci": "0.4.1",
-    "cozy-client": "24.1.0",
+    "cozy-client": "27.4.0",
```
Crash--, line 168: this is not a "chore" commit. Since the code is present in the bundle...

Crash--: You're not concerned by the breaking change? 

Specially this one: https://github.com/cozy/cozy-client/blob/master/CHANGELOG.md#2500-2021-10-12 

There is a few places where banks use save() https://github.com/cozy/cozy-banks/search?q=save%28 .

src/components/App.jsx
```diff
@@ -34,7 +38,27 @@ const KeyboardAwareSidebar = ({ children }) => {
 
 const App = props => {
   const { showBottomNav, settingsCollection } = props
+  const { t } = useI18n()
+  const client = useClient()
   const settings = getDefaultedSettingsFromCollection(settingsCollection)
+
+  const showAlert = t => {
```
Crash--, line 45: why `t` is passed as an argument ?

src/components/App.jsx
```diff
@@ -34,7 +38,27 @@ const KeyboardAwareSidebar = ({ children }) => {
...
+
+  useEffect(() => {
+    const onError = e => {
+      if (!getActivatePouch() && hasFetchFailedError(e)) {
+        throttle(showAlert(t), 300)
+      }
+    }
+
+    client.setOnError(onError)
+  }, [client, t])
```
Crash--, line 60: If `t` changes, we'll call again the `setOnError` method and that'll throw an error. Why not just running this setOnError with an empty dep's array?

src/components/App.jsx
```diff
@@ -34,7 +38,27 @@ const KeyboardAwareSidebar = ({ children }) => {
...
+  }
+
+  useEffect(() => {
+    const onError = e => {
+      if (!getActivatePouch() && hasFetchFailedError(e)) {
+        throttle(showAlert(t), 300)
+      }
+    }
+
+    client.setOnError(onError)
```
Crash--, line 59: weird to call setOnError(undefined), no? It'll assign this.onError = undefined in CozyClient side. I find it weird. You should not call setOnError if not needed?

package.json
flohhhh: Oops, thx

src/components/App.spec.jsx
```diff
@@ -0,0 +1,98 @@
...
+// due to pouchdb-browser. Here we are not concerned with this
+// component
+jest.mock('cozy-pouch-link', () => () => null)
+jest.mock('components/AppSearchBar', () => () => null)
+jest.mock('ducks/commons/Nav', () => () => null)
+jest.mock('lodash/throttle', () => ({
+  default: jest.fn(fn => fn),
+  __esModule: true
+}))
+jest.mock('lodash/throttle')
```
Crash--, line 22: you mock two times throttle? 🤔

src/components/App.jsx
flohhhh: Because it is used in show Alert. `t('Error.fetch-error')` and `t('General.reload')`

src/ducks/client/links.js
```diff
@@ -8,7 +8,7 @@ import { offlineDoctypes, TRANSACTION_DOCTYPE } from 'doctypes'
 import { pickAdapter, getAdapterPlugin } from 'ducks/client/linksHelpers'
 import { getWarmupQueries } from './linksHelpers'
 
-const activatePouch = __POUCH__ && !flag('banks.pouch.disabled')
+export const getActivatePouch = () => __POUCH__ && !flag('banks.pouch.disabled')
```
Crash--, line 11: isPouchActive? Since we return a boolean... Should be added to https://github.com/cozy/cozy-guidelines#naming-of-functions

src/components/App.jsx
flohhhh: Yes absolutely

src/components/App.spec.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+      fireEvent.click(root.getByText('Fetch'))
+
+      await wait(() => {
+        expect(Alerter.error).toHaveBeenCalled()
+
+        const errorMsg = Alerter.error.mock.calls[0][0]
+        expect(errorMsg).toBe(
+          'Connection lost. Try again later or verify your connection and reload the page.'
+        )
```
Crash--, line 91: nit: you should avoid testing the translated string. You should only rely on the key.

src/ducks/transactions/helpers.spec.js
```diff
@@ -25,6 +25,9 @@ import CozyClient from 'cozy-client'
 import { createClientWithData } from 'test/client'
 import { getCategoryIdFromName } from 'ducks/categories/helpers'
 
+// eslint-disable-next-line no-console
+console.warn = jest.fn()
```
Crash--, line 29: Why do you have a warning now? You should fix it instead of ignoring it!

src/targets/services/helpers.spec.js
```diff
@@ -101,7 +101,7 @@ describe('getOptions', () => {
 })
 
 describe('fetchChangesOrAll', () => {
-  const client = new CozyClient()
+  const client = new CozyStackClient()
```
Crash--, line 104: this change is weird. Can you explain it?

Crash-- (changes requested):
mocking console.warn is not a solution. Please don't do that

src/components/App.jsx
Crash--: But you already have access to `t` within this scope, no? What is the gain to have in passed in args?

src/components/App.jsx
flohhhh: Here onError is defined. setOnError is not mandatory so if you want use it it's possible. Maybe are you talking about other things?

src/components/App.spec.jsx
flohhhh: Removed

src/components/App.jsx
Crash--: No you're right. My bad, thanks.

src/components/App.jsx
flohhhh: Well seen

package.json
Crash--: Can you also upgrade cozy-stack-client and cozy-pouch-link please?

flohhhh: It's upgraded, and it seems ok about `_type` in save.

Crash-- (approved):

PR #2321, opened 2022-02-09
feat: Implement WebviewIntentProvider

acezard:
In order to make all **Cozy Apps** compatible with **Amirale**, we have to wrap them in the `cozy-intent` provider.

To do so, we have to:
- Add `cozy-intent`
- Upgrade `cozy-device-helper` to the latest version (necessary for `cozy-intent` as its peer dependency)
- Wrap the application at the highest possible level with the `<WebviewIntentProvider />`


JF-Cozy (approved):
👍 👏

PR #2340, opened 2022-03-02
fix: Inject cozy-bar callback into cozy-intent

acezard:
Also require cozy-intent latest version


src/AppContainer.jsx
```diff
@@ -21,6 +21,7 @@ import BanksProvider from 'ducks/context/BanksContext'
 import SelectionProvider from 'ducks/context/SelectionContext'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
 import { initTranslation } from 'cozy-ui/transpiled/react/I18n'
+import cozyBar from './utils/cozyBar'
```
Crash--, line 24: no relative import. Please use alias 🙏

Crash-- (approved):

PR #2342, opened 2022-03-09
fix(documentSelector): remove useless memoization after cozy-client fix

trollepierre:
This PR reverts the fix - but not the styles of this [commit](https://github.com/cozy/cozy-banks/commit/fe4c90cc21c3c7744350b6a99612a21a5f50b737).

Memoization is not needed anymore, since [cozy-client has been fixed](https://github.com/cozy/cozy-client/pull/1122/files).


JF-Cozy (approved):
on pourrait peut-être expliquer un peu mieux le pourquoi du comment ?

PR #2355, opened 2022-04-20
fix: Ugrade harvest to 8.2.1

doubleface:
To get a correct mapping of bnp_es and cic_es https://github.com/cozy/cozy-libs/commit/f304f8978f1857f3f1d1e74c0c19385ab96dde24

To fix https://trello.com/c/DOoMt0WV/381-les-connecteurs-bnpp-incoh%C3%A9rents

The BC in version 8.0.0 of cozy-harvest-lib says that we need at leat cozy-ui v60.6.0 and we have 62.2.0.


JF-Cozy (approved):

package.json
```diff
@@ -167,7 +167,7 @@
     "cozy-device-helper": "^1.17.0",
     "cozy-doctypes": "1.82.2",
     "cozy-flags": "2.4.0",
-    "cozy-harvest-lib": "^7.3.1",
+    "cozy-harvest-lib": "8.2.1",
```
JF-Cozy, line 170: "cozy-harvest-lib": "^8.2.1" serait mieux, il manque le `^`

JF-Cozy (approved):

package.json
trollepierre: I don't see a problem. Is it a nit? Why would the `^` help?

Is there a guideline in cozy about that?

ps: @JF-Cozy  should we add this discussion inside a FrontMeeting?

trollepierre (approved):

package.json
JF-Cozy: oui c'est un nit, j'ai validé la PR en même temps. Oui il y a une guildeline sur ça, les lib cozy sont avec des `^` dans les app cozy.

PR #2374, opened 2022-05-18
fix: Add replaceHistory method for Harvest

acezard:
Harvest uses a new API for internal routing.
The MountPointContext provides a replaceHistory method.
But it doesn't provide it in cozy-banks because the context is
not provided by Harvest but by the application.
It is needed to add the method to cozy-banks context.
It will use the same method as pushHistory,
as it's not a real router method here but a component state.
To avoid such problems in the future: use typed interfaces,
or do not replace old interfaces with new interfaces.
Instead, try something like pushHistory(args, withReplace?: boolean).

### DIFF
--- a/src/ducks/settings/HarvestSwitch.jsx
+++ b/src/ducks/settings/HarvestSwitch.jsx
@@ -68,7 +68,10 @@ class Switch extends React.Component {
     }
 
     this.pushHistory = this.pushHistory.bind(this)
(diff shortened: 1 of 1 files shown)


Crash-- (approved):

Crash--:
(no need to add this line to the changelog, it's a fix for a bug added to this release)

PR #2404, opened 2022-06-22
fix(account): Update trigger without fetch policy

trollepierre:
Two PR are needed to remove the error message staying, after reconnection of a konnector, this one and https://github.com/cozy/cozy-libs/pull/1697 . We need to update cozy-harvest-lib also to fix the bug. => draft

The [HarvestBankAccountSettings](https://github.com/cozy/cozy-banks/blob/master/src/ducks/settings/HarvestBankAccountSettings.jsx#L253) give data to HarvestAccountModal using the [HarvestLoader](https://github.com/cozy/cozy-banks/blob/dbc8d428fbeac34bafea11213607fd565e8daed7/src/ducks/settings/HarvestBankAccountSettings.jsx#L179) .
I noticed the data "trigger" that contains `trigger.current_state.status` (knowing if errored or done) was given inside the "accountsAndTriggers" . 

I also modified AccountModal.componentDidUpdate to update the data (flowState), used by harvest lib, to remove the error message displayed. (Other PR in cozy-harvest-lib)

Then, I noticed the request to refresh trigger was executed sometimes. When fetchPolicy returns previous konnector with "errored" status, then the error message was not removed. When a fresh request happens, updating the triggers, then the status was done and the error message was removed.

That's why the solution is to remove the fetchPolicy, as we are not using realtime here + our date trigger can change in this scenario  (This PR in cozy-banks)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## 🐛 Bug Fixes
 
+* Update trigger without fetch policy in AccountModal, while reconnecting a konnector
+
 ## 🔧 Tech
 
 # 1.44.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 ## ✨ Features
 
+* Update cozy-harvest-lib
+  * 9.15.1 : Update trigger while status changed [[PR]](https://github.com/cozy/cozy-libs/pull/1697)
+
 ## 🐛 Bug Fixes
 
-* Update trigger without fetch policy in AccountModal, while reconnecting a konnector
+* Update trigger without fetch policy in AccountModal, while reconnecting a konnector [[PR]](https://github.com/cozy/cozy-banks/pull/2404)
 
 ## 🔧 Tech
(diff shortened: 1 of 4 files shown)


Crash-- (approved):

JF-Cozy (approved):
penser à mettre à jour harvest également, ainsi que le changelog je pense ?

JF-Cozy:
> @JF-Cozy : tu peux revoir le dernier commit, et si ça te convient, on merge !

@trollepierre c'est un feat et non un fix (même si l'upgrade de la lib apporte un fix). Il faut donc nommer le commit en fonction, et mettre dans les feat dans le changelog 👍

trollepierre:
ok

> > @JF-Cozy : tu peux revoir le dernier commit, et si ça te convient, on merge !
> 
> @trollepierre c'est un feat et non un fix (même si l'upgrade de la lib apporte un fix). Il faut donc nommer le commit en fonction, et mettre dans les feat dans le changelog 👍

ok, j'update et je merge

PR #2413, opened 2022-06-29
feat: Add intentsApi prop to HarvestBankAccountSettings

doubleface:
This allows any app overriding cozy-banks to customize how harvest
handle an inApp Browser


Crash-- (approved):
Did you look at all the call to pass these props?

doubleface:
@Crash-- I suppose. Now I know I missed it because this one was not overridden by the maif app yet and a grep could not detect it. I tested it with a modified version of intentsApi and we get the correct intentsApi in the reconnect screen. Then it should work in the maif native app.

PR #2425, opened 2022-07-16
fix: avoid using planes (for travelling) or cars (for public transport)

arthurlutz:


arthurlutz:
SVGs copied from https://www.iconfinder.com/iconsets/vehicles-and-travel-1 if you take them as is, the project will probably have to reference the author https://www.wishforge.games/freebies (I think)

Crash--:
Thank you @arthurlutz for the PR. 

We understand what you wanted to do here. 

@joel-costa what do you think about this idea?

PR #2434, opened 2022-07-21
feat: Upgrade Harvest and CozyClient

doubleface:
Harvest to 9.22.2 to get :
 * fix: Do not use biBankId from cache
 * feat: Allow disabled banks to be clickable
 * fix: Add BI bankIds to createTemporaryToken params
Cozy-client to 32.2.6 to get :
 * fix temporary token cache when beginning with an empty cache [PR](https://github.com/cozy/cozy-client/pull/1198)


JF-Cozy (commented):
on peut tout merger dans un commit je pense...

Par contre on oublie d'upgrade cozy-stack et pouch c'est normal ? 🤔 Il faut que les 3 soient synchro

PR #2437, opened 2022-07-21
fix: Avoid notifying bill notifications twice

taratatach:
Some notifications sent by the `onOperationOrBillCreate` service can
be sent multiple times as we're not saving anywhere the fact that
they've already been sent.
This is the case of `HealthBillLinked` notifications for example.
These are sent when a reimbursement bill is linked to a transaction
but if that transaction gets updated somehow later, then the service
will send another notification once it runs.

The `TransactionGreater` notification class had a mechanism based on
the `cozyMetadata` to save this information on the transactions
themselves so we just use it for all transaction based notifications
(i.e `HealthBillLinked`, `LateHealthReimbursement` and
`TransactionGreater`).

This will generate changes on the transactions as we're saving the
state in their metadata but now that the recurrence service does not
call itself in a loop anymore, this should be fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Fix recurrence service that was triggering itself in an infinite loop [[PR]](https://github.com/cozy/cozy-banks/pull/2429)
 * Fix typo in French translation [[PR]](https://github.com/cozy/cozy-banks/pull/2430)
 * Transactions actions style on mobile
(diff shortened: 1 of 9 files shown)


JF-Cozy (approved):
est-ce que tu peux rajouter une entrée dans le changelog stp ?

taratatach:
Ah yes, j'oublie à chaque fois…

PR #2453, opened 2022-08-03
fix: make the back button from `TagDialog` custom

PolariTOON:
This also prepares for the override


Merkur39:

Merkur39 (approved):

PR #2456, opened 2022-08-04
feat: Add a centered spinner after bi webview

doubleface:


src/main.jsx
```diff
@@ -48,7 +49,13 @@ if (__TARGET__ === 'mobile') {
...
 
 const initRender = () => {
-  if (handleOAuthResponse()) return
+  if (handleOAuthResponse()) {
+    render(
+      <Spinner size="xxlarge" middle={true} />,
+      document.querySelector('[role=application]')
+    )
+    return
+  }
```
Crash--, line 58: Should we add it to the Home app too? 

Also we spoke about adding a text like "Finalization in progress"

doubleface: This will be added to the home application too (after this review).

Now I added the text without localization.
Adding localization may make the load time longer in this case and I would have to duplicate the loading of translations. Do you think I should do it ?

Crash--: You should have describe that if the previous version of the PR. I would not asked about the text. 

Let's remove it instead of having in a the wrong language

doubleface: Updated

Crash-- (approved):

PR #2457, opened 2022-08-04
Add a button to create a tag in the settings

PolariTOON:


Merkur39:

Crash-- (approved):

PR #2463, opened 2022-08-26
fix: Mango request optimization

Crash--:
Since the "baseQuery" can have a selector, we need to add this
selector to the indexField in order to have a performant index.

To be able to sort correctly, we need to add the selector to the
sort. So here it is.


JF-Cozy (approved):
tu vas râler si je te parle de changelog.md ?

trollepierre (approved):

PR #2471, opened 2022-08-30
fix: Duplicate CSS

Crash--:
Since https://github.com/cozy/create-cozy-app/commit/fa07a257c90e4b359097b56761d929755b4b5f17
we duplicate the generated CSS. It's bad since we load twice the
same CSS, so in fact we load at least 4 CSS files for Banks.

These CSS are pretty heavy, so let's revert to only get the ones
we really need without duplication.

We'll fix the issue in cozy-scripts later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## ✨ Features
 * Optimization of the Mango query to get first / last date for an account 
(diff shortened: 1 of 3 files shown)


Merkur39 (approved):
Thx 👍

PR #2476, opened 2022-08-31
fix: Upgrade harvest to 9.26.6

doubleface:
To have a better handling of duplicate accounts : locale + realtime in bank accounts


Crash--:
@doubleface CI is red

PR #2477, opened 2022-08-31
fix: The label entry must be limited directly

Merkur39:
~~The `maxLength` property is not supported on Android.
This problem is probably caused by the predictive text feature.~~

It's better to make the field controlled
and adapt the previous fix on spaces (0b7cfc9)

### DIFF
--- a/src/components/Tag/TagAddNewTagModal.jsx
+++ b/src/components/Tag/TagAddNewTagModal.jsx
@@ -9,6 +9,8 @@ import Button from 'cozy-ui/transpiled/react/Buttons'
 
 import { TAGS_DOCTYPE } from 'doctypes'
 
+const labelMaxLength = 30
+
 const TagAddNewTagModal = ({ onClick, onClose, withLabel }) => {
   const client = useClient()
   const { t } = useI18n()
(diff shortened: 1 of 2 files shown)


Crash--:
I don't know. I understand that chrome only respect the limits when the user click on enter / space or something like that since the user can select a proposed word by its keyboard (cf https://bugs.chromium.org/p/chromium/issues/detail?id=431747#c11) 

https://caniuse.com/maxlength says for Android user: " Allows text beyond maxlength to be entered at first, but removes all characters past the maxlength when focus is lost." 

So shouldn't we only blur before submiting? Having a controlled input makes the input feels laggy since React has to do its things

Merkur39:
En effet, le comportement est bien celui expliqué dans le lien.
Si l'utilisateur tape un label trop long, il sera automatique tronqué à la validation (ou lors d'un ajout d'espace)

Mais la volonté du client est bien de forcer la limitation lors de la saisie.
Il nous faut donc procéder ainsi 🤷

Crash-- (approved):

Merkur39:

PR #2481, opened 2022-09-01
feat: Add wait jobs

doubleface:
Wait jobs allow harvest to tell to banks it is waiting for a job to be
triggered by a BI webhook.

This wait job will displayed like a real job in progress but will be
replaced and removed by the real job with the same slug or after 5
minutes.

Harvest can trigger this wait jobs by sending a realtime notify on
io.cozy.jobs with the slug of the expected job.

Since these wait jobs are triggered before any account or trigger
creation, they don't have any associated account and will be replaced by
any job with the same expected slug

### DIFF
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -255,9 +255,9 @@ class Balance extends PureComponent {
       checkedAccounts.length === accounts.length
         ? undefined
         : {
-          nbCheckedAccounts: checkedAccounts.length,
-          nbAccounts: accounts.length
-        }
+            nbCheckedAccounts: checkedAccounts.length,
+            nbAccounts: accounts.length
+          }
 
     return (
       <Fragment>
@@ -304,8 +304,9 @@ const addTransactions = Component => {
     const transactions = useFullyLoadedQuery(conn.query, conn)
     return <Component {...props} transactions={transactions} />
   }
-  Wrapped.displayName = `withTransactions(${Component.displayName || Component.name
-    })`
+  Wrapped.displayName = `withTransactions(${
+    Component.displayName || Component.name
+  })`
   return Wrapped
 }
 
--- a/src/ducks/balance/Balance.jsx
+++ b/src/ducks/balance/Balance.jsx
@@ -226,6 +226,7 @@ class Balance extends PureComponent {
   }
 
   render() {
+    const { hasJobsInProgress } = this.props
     if (isLoading(this.props)) {
       return (
         <Fragment>
@@ -241,7 +242,7 @@ class Balance extends PureComponent {
     const hasNoAccount = accounts.filter(a => !isVirtualAccount(a)).length === 0
 
     if (
-      hasNoAccount ||
+      (hasNoAccount && !hasJobsInProgress) ||
       flag('balance.no-account') ||
       flag('banks.balance.account-loading')
     ) {
--- a/src/ducks/balance/BalanceWithBanksJobs.jsx
+++ b/src/ducks/balance/BalanceWithBanksJobs.jsx
@@ -1,0 +1,10 @@
+import React from 'react'
+import { useBanksContext } from 'ducks/context/BanksContext'
+import Balance from 'ducks/balance/Balance'
+
+const BalanceWithBanksJobs = () => {
+  const { hasJobsInProgress } = useBanksContext()
+  return <Balance hasJobsInProgress={hasJobsInProgress} />
+}
+
+export default BalanceWithBanksJobs
--- a/src/ducks/balance/index.js
+++ b/src/ducks/balance/index.js
@@ -1,4 +1,4 @@
-import Balance from 'ducks/balance/Balance.jsx'
+import Balance from 'ducks/balance/BalanceWithBanksJobs'
 
 // components
 export { Balance }
--- a/src/ducks/context/BanksContext.jsx
+++ b/src/ducks/context/BanksContext.jsx
@@ -46,9 +46,14 @@ const BanksProvider = ({ children, client }) => {
           const slug = jobsInProgress.konnector
           return `${KONNECTOR_DOCTYPE}/${slug}`
         })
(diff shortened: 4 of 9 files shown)


Crash--:
How does the integration with our current workflow works? Do we need to add this feature behind a flag?

doubleface:
This behavior is triggered by a realtime message from harvest. If this message is not triggered, there is no visible change. [The associated change in harvest](https://github.com/cozy/cozy-libs/pull/1804) will need to be behind a flag.

Crash--:
@doubleface can you rebase please?

src/ducks/balance/Balance.jsx
```diff
@@ -325,5 +327,6 @@ export default compose(
     })
   ),
   withClient,
+  withBanksJobsInProgress,
```
Crash--, line 330: Pas moyen qu'on remette ce HOC sur Balances. Ça provoque quelques re-render et c'est vraiment pas bon pour les perfs. 

Par contre, tu peux : 
- Wrapper Balance dans un nouveau composant 
- Ce nouveau composant tu le fais en function 
- Tu peux utiliser le context sans le HOC 
- Tu peux passer les props à Balance

src/ducks/context/JobsContext.spec.jsx
```diff
@@ -42,7 +53,7 @@ describe('Jobs Context', () => {
       <JobsContext.Consumer>
         {({ jobsInProgress }) => {
           return jobsInProgress.map(job => (
-            <div key={job.account}>
+            <div key={job.konnector}>
```
Crash--, line 56: Pourquoi ce changement ?

doubleface: On n'affiche qu'un job in progress par connecteur et de plus, les nouveau jobs d'attente de job n'ont pas d'account associé mais ont bien un connecteur

src/ducks/balance/index.js
```diff
@@ -1,4 +1,4 @@
-import Balance from 'ducks/balance/Balance.jsx'
+import Balance from 'ducks/balance/BalanceWithBanksJobs'
```
Crash--, line 1:
```
import BalanceWithBanksJobs from 'ducks/balance/BalanceWithBanksJobs'
```

grep test is hard when the name is not the same.

src/ducks/context/BanksContext.jsx
```diff
@@ -90,3 +95,20 @@ export default BanksProvider
...
+ */
+export const withBanksJobsInProgress = Component => {
+  const Wrapped = props => {
+    const { jobsInProgress = [], hasJobsInProgress } = useBanksContext()
+    return <Component {...props} jobsInProgress={jobsInProgress} hasJobsInProgress={hasJobsInProgress} />
+  }
+  Wrapped.displayName = `withBanksJobsInProgress(${Component.displayName || Component.name
+    })`
+  return Wrapped
+}
```
Crash--, line 121: to remove from this commit?

src/ducks/settings/AccountsListSettings.jsx
```diff
@@ -116,12 +116,13 @@ const AccountsListSettings = ({
           const isLoading =
             jobsInProgress.some(j => j.account === connectionId) ||
             accounts[0].inProgress
+          const inProgress = accounts?.[0]?.inProgress
```
Crash--, line 119: Why do we need to create a `inProgress` var since `isLoading` already read `accounts[0].inProgress` if needed ?

src/ducks/context/BanksContext.jsx
Crash--: This component is not used anymore, right? We can remove it.

src/ducks/context/JobsContext.spec.jsx
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 
 import JobsProvider, { JobsContext } from '../context/JobsContext'
-import { render } from '@testing-library/react'
-import { createMockClient } from 'cozy-client'
+import { render, act } from '@testing-library/react'
+import CozyClient from 'cozy-client'
+import CozyRealtime from 'cozy-realtime'
+
+jest.mock('cozy-realtime')
```
Crash--: You should squash the latest commit with this one. There is no need to have an intermediate version of this using CozyRealtime directly.

src/ducks/balance/BalanceWithBanksJobs.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { useBanksContext } from 'ducks/context/BanksContext'
+import Balance from 'ducks/balance/Balance'
+
+const BalanceWithBanksJobs = props => {
```
Crash--, line 5: Are you sure that it takes props? We should avoid spreading `...props` when we know there is no props

doubleface: No it doesn't. I removed it

src/ducks/balance/index.js
doubleface: But BalanceWithBanksJobs is still greppable. And If modify the exported name to BalanceWithBanksJobs, I will have to modify the maif override accordingly, Do you think it is worth it ?

src/ducks/settings/AccountsListSettings.jsx
doubleface: They don't have the same meaning. I renamed them to isAnyJobLoading and isCurrentJobInProgress

doubleface:
@Crash-- Can you check it one more time ?

src/ducks/context/JobsContext.jsx
```diff
@@ -10,6 +10,7 @@ import { JOBS_DOCTYPE } from 'doctypes'
 import logger from 'cozy-logger'
 import isFunction from 'lodash/isFunction'
 import PropTypes from 'prop-types'
+import { WaitJobQueue } from '../settings/WaitJobQueue'
```
Crash--, line 13: no relative import please.

Crash-- (approved):

PR #2492, opened 2022-09-12
fix: Make in progress item in setting not clickable

doubleface:
Or else, this will display a meaningless overlay for such items


JF-Cozy (approved):

Crash--:
@doubleface can you merge it?

PR #2502, opened 2022-09-19
fix: Check if we get relationships or not before maping

Crash--:
```
### 🐛 Bug Fixes

* Since `tags` are behind a flag, it can appears that we don't have the `tags` relationships at all. Let's be defensive about that case in order to not throw an error. Fix an error when the user can not open a specific transaction 

```


Merkur39 (approved):
👍

PR #2508, opened 2022-09-22
Correctly retrieve the setting that hides amounts in notifications

PolariTOON:
This is a fix for #2493 which is not yet in beta so the changelog is omitted.

Amounts used to be hidden even with the setting disabled.
Basically the variable `amountCensoring` was an object and not a boolean and thus was always truthy.


JF-Cozy (approved):

Crash--:
Y avait pas un test pour ça ?

PR #2522, opened 2022-10-11
deps: Upgrade date-fns to v2.29.3

taratatach:
We want to use the `add` and `sub` functions that are not available
in v1.30.1.

Since there's a large gap between the two versions, we have a lot of
modifications to accommodate API changes:
- imports of functions whose names are made of multiple words have
  changed from a snake case format to a camel case one
- locales are not exposed as separate modules anymore; they need to be
  imported from `date-fns/locales`
- the format used by `format()` and `parse()` is now based on Unicode
  Technical Standard https://github.com/cozy/cozy-banks/pull/35 (see https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table)
  with a few additions (see https://date-fns.org/v2.29.3/docs/format)
  → `YYYY` becomes `yyyy`
  → `D` becomes `d`
  → `DD` becomes `dd`
  → `ddd` becomes `EEE`
  → the escape character is now `'` (single quote) instead of `[`
- all `date-fns` functions now require arguments to be `Date` objects
  or numbers and do not accept strings anymore. Therefore, all
  date strings are now parsed before being passed as `date-fns`
  function arguments and all test data mocks have been updated to use
  only ISO formatted date strings
- `isWithinRange` becomes `isWithinInterval` which is inclusive
  (i.e. `isWithinRange(3, 1, 3) === false` vs
  `isWithinInterval(3, { start: 1, end: 3 }) === true`)

```
### ✨ Features

*

### 🐛 Bug Fixes

*

### 🔧 Tech

* Upgrade `date-fns` to v2.29.3 and accommodate API changes.
```

### DIFF
--- a/src/ducks/recurrence/Bundles.jsx
+++ b/src/ducks/recurrence/Bundles.jsx
@@ -1,18 +1,17 @@
 import React, { useMemo } from 'react'
 import cx from 'classnames'
-import distanceInWords from 'date-fns/distance_in_words'
+import intlFormatDistance from 'date-fns/intlFormatDistance'
 import CompositeRow from 'cozy-ui/transpiled/react/CompositeRow'
 import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import { Bd, Img, Media } from 'cozy-ui/transpiled/react/Media'
 import Figure from 'cozy-ui/transpiled/react/Figure'
+
 import CategoryIcon from 'ducks/categories/CategoryIcon'
 import Table, { TdSecondary } from 'components/Table'
 import { useHistory } from 'components/RouterContext'
 
-import frLocale from 'date-fns/locale/fr'
-import enLocale from 'date-fns/locale/en'
-import esLocale from 'date-fns/locale/es'
+import { fr as frLocale, en as enLocale, es as esLocale } from 'date-fns/locale'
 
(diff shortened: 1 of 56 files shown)


taratatach:
This PR is waiting for other libraries to upgrade `date-fns` (e.g. cozy-ui) as we deduplicate libs and the new version of `date-fns` is incompatible with the old one used in our own libs.

src/ducks/recurrence/Bundles.jsx
```diff
@@ -1,18 +1,17 @@
...
 import Figure from 'cozy-ui/transpiled/react/Figure'
+
 import CategoryIcon from 'ducks/categories/CategoryIcon'
 import Table, { TdSecondary } from 'components/Table'
 import { useHistory } from 'components/RouterContext'
 
-import frLocale from 'date-fns/locale/fr'
-import enLocale from 'date-fns/locale/en'
-import esLocale from 'date-fns/locale/es'
+import { fr as frLocale, en as enLocale, es as esLocale } from 'date-fns/locale'
```
PolariTOON, line 14: This version of `date-fns` does not export any `en` locale anymore. You likely want `enUS` instead.

PolariTOON:
We have a webpack config in this app that defines a `ContextReplacementPlugin` for `date-fns` locales. The regexp should likely be changed from `/(en|fr|es)\/index\.js/` to ` /(en-US|fr|es)\/index\.js/`.

PR #2524, opened 2022-10-14
Include virtual accounts in the account total count in balance details

PolariTOON:
The transactions from the virtual accounts were already included in the transaction list on the balance details page when checked on the balance page, but the total count displayed on the account selector was taking into account only the real accounts.

This fixes this mismatch by making the counter also include virtual accounts.

Virtual accounts still can't be directly selected in the selector menu and can only be accessed from the balance page.

```
### 🐛 Bug Fixes

* Include virtual accounts in the account total count in balance details
```


JF-Cozy (approved):
good catch 👍 
très bonne description de PR également ;)

PR #2533, opened 2022-11-04
feat: Ignore some webhooks for some events

doubleface:
At the moment, banks display an import success toast when any bank job
related to a banking connector successed.

Since it is now possible to have multiple konnector job successes for
the creation of one connection ore even on the deletetion of a
connection, we need to filter out some jobs.

Here are the rules to consider a job :

- The job is not a webhook (still possible run manually by the user)
- The job is a webhook of type CONNECTION_SYNCED

ACCOUNT_ENABLED, ACCOUNT_DISABLED and CONNECTION_DELETED jobs are
ignored for the detection of running jobs and for the display of success
or error toasts

```
### ✨ Features

* Do not display success or error toast for webhooks other than CONNECTION_SYNCED

```

### DIFF
--- a/src/ducks/context/JobsContext.spec.jsx
+++ b/src/ducks/context/JobsContext.spec.jsx
@@ -4,19 +4,39 @@ import JobsProvider, { JobsContext } from '../context/JobsContext'
 import { render, act } from '@testing-library/react'
 import CozyClient from 'cozy-client'
 
-export const createKonnectorMsg = (state, konnector, account) => ({
+export const createKonnectorMsg = (
+  state,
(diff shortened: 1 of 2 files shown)


src/ducks/context/JobsContext.spec.jsx
```diff
@@ -82,7 +86,7 @@ describe('Jobs Context', () => {
   })
 
   it('should not display job in progress', () => {
```
Crash--, line 88: Having a more representative test name wouldn't hurt

Crash-- (approved):

PR #2537, opened 2022-11-15
fix: Remove realtime on accounts

doubleface:
This causes rendering issues on Harvest at the moment.
Another [temporary fix](https://github.com/cozy/cozy-libs/pull/1878) has been
found to update account from harvest waiting for a better fix which will
make OAuthWindow in harvest independent from Harvest components tree.

```
### 🐛 Bug Fixes

* Remove realtime on accounts

```


Crash--:
I added the "temporary" keywork in your description. Please fix lint issue (or configure your IDE to do that automatically for you on save)

PR #2545, opened 2022-11-25
feat: Activate realtime on cozy accounts

doubleface:
Harvest is also upgraded to support the update of the current account
even when it is displaying a popup or an InAppBrowser- feat: Activate realtime on cozy accounts

```
### ✨ Features

* Add realtime update of cozy accounts
* Harvest upgrade to 9.30.1 : 
    * Add independent confirm dialog
    * Use independant OAuth window
    * Better locale for delete BI connection confirmation button
```

### DIFF
--- a/src/AppContainer.jsx
+++ b/src/AppContainer.jsx
@@ -17,6 +17,7 @@ import flag from 'cozy-flags'
 
 import { TrackerProvider } from 'ducks/tracking/browser'
 import JobsProvider from 'ducks/context/JobsContext'
+import { CozyConfirmDialogProvider } from 'cozy-harvest-lib'
 import BanksProvider from 'ducks/context/BanksContext'
 import SelectionProvider from 'ducks/context/SelectionContext'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
@@ -82,6 +83,7 @@ const AppContainer = ({ store, lang, history, client }) => {
                     <BanksProvider client={client}>
                       <SelectionProvider>
                         <StoreURLProvider>
+                          <CozyConfirmDialogProvider>
                           <MuiCozyTheme>
                             <RealTimeQueries doctype={TRIGGER_DOCTYPE} />
                             <RealTimeQueries doctype={ACCOUNT_DOCTYPE} />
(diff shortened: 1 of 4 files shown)


src/AppContainer.jsx
```diff
@@ -82,13 +84,16 @@ const AppContainer = ({ store, lang, history, client }) => {
...
+                              <RealTimeQueries doctype={ACCOUNT_DOCTYPE} />
+                              <RealTimeQueries doctype={COZY_ACCOUNT_DOCTYPE} />
+                              <RealTimeQueries doctype={TRANSACTION_DOCTYPE} />
+                              <RealTimeQueries doctype={GROUP_DOCTYPE} />
+                              <Router history={history} routes={AppRoute()} />
+                            </MuiCozyTheme>
+                          </CozyConfirmDialogProvider>
                         </StoreURLProvider>
                       </SelectionProvider>
                     </BanksProvider>
```
Crash--, line 99: nit: It's easier to read, understand  and maintain if this commit was split into both commits: 
- One for the harvest upgrade and the addition of CozyConfirmDialogProvider 
- One other for the use of cozy_account_doctype realtime.

Crash-- (approved):

src/AppContainer.jsx
```diff
@@ -82,13 +84,16 @@ const AppContainer = ({ store, lang, history, client }) => {
...
+                          <CozyConfirmDialogProvider>
+                            <MuiCozyTheme>
+                              <RealTimeQueries doctype={TRIGGER_DOCTYPE} />
+                              <RealTimeQueries doctype={ACCOUNT_DOCTYPE} />
+                              <RealTimeQueries doctype={COZY_ACCOUNT_DOCTYPE} />
+                              <RealTimeQueries doctype={TRANSACTION_DOCTYPE} />
+                              <RealTimeQueries doctype={GROUP_DOCTYPE} />
+                              <Router history={history} routes={AppRoute()} />
+                            </MuiCozyTheme>
+                          </CozyConfirmDialogProvider>
```
PolariTOON, line 96: Should `MuiCozyTheme` be above `CozyConfirmDialogProvider`? Otherwise dialogs will not have access to the theme as far as i understand.

Crash--: Totally right! Can you create the fix?

PR #2547, opened 2022-11-29
fix: Stop using cozy-doctypes' Document.updateAll

taratatach:
Use cozy-client's `saveAll` instead as it is up-to-date and will
cleanup data before sending it to cozy-stack.
This avoids having CouchDB errors because documents have attributes
starting with an underscore (`_type` in this case).

```
### 🐛 Bug Fixes

* Use `CozyClient.saveAll()` instead of `Document.updateAll()` from deprecated `cozy-doctypes` to avoid CouchDB errors with attributes starting with an underscore (#2547)

### 🔧 Tech

* Add more logs around LateHealthReimbursement onSuccess method (#2547)
```

### DIFF
--- a/src/ducks/notifications/HealthBillLinked/index.js
+++ b/src/ducks/notifications/HealthBillLinked/index.js
@@ -3,7 +3,6 @@ import merge from 'lodash/merge'
 
 import log from 'cozy-logger'
 import { toText } from 'cozy-notifications'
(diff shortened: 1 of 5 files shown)


JF-Cozy (approved):

Crash-- (approved):
SOOOOO COOOOOOOL! Congrats! 

👏 👏 👏

PR #2548, opened 2022-11-30
feat: Upgrade cozy-realtime to 4.2.9

doubleface:
To get the sendNotification method, needed by harvest

```
### 🔧 Tech

* upgrade cozy-realtime to 4.2.9 to get the sendNotification method
```


Crash--:
Harvest should have got a Breaking Change for this..

PR #2552, opened 2022-12-02
fix: Swap providers to make some Harvest dialogs match the theme

PolariTOON:
This is an (untested) fix following #2545, that should make sure Bank disconnect dialogs match Cozy UI theme (and also its override variation).
```
### 🐛 Bug Fixes

* Swap providers to make some Harvest dialogs match the theme
```


Crash-- (approved):
👍 Thanks and nice catch

PR #2564, opened 2022-12-05
feat: Ignore account delete and account disabled jobs

doubleface:
When a konnector job is successful, we display an "Import success"
message.

We do not want to display this "Import success" message for job
triggered when an account (io.cozy.account) is deleted and for job
triggered by "ACCOUNT_DISABLED" webhooks.

```
### ✨ Fix

* Ignore account delete and account disabled jobs for import success message
```


Crash--:
let's add a few tests

doubleface:
@Crash-- I finally put the code in JobsContext and then added a few tests

Crash-- (approved):

PR #2565, opened 2022-12-05
fix: Do not display an error icon for all accounts with the same slug

doubleface:
Now, trigger error are indexed by their corresponding account id
(io.cozy.account).

This way, when a trigger with account_type "slug1" is in error, not all
accounts related to this account_type are displayed as in error but only
the account related to the trigger in error.

```
### 🐛 Bug Fixes

* Do not display an error icon for all accounts with the same slug
```


Crash-- (approved):
Approving, it order to merge it and make a beta ASAP. But please add test for this in a new PR.

PR #2591, opened 2023-01-05
feat: Remove cozy-bar v7 assets

zatteo:
cozy-bar v7 assets were still imported in production mode. It became visible with recent cozy-bar releases where CSS was really different between v7 and v8.

```
### 🐛 Bug Fixes

* Fix bad display of settings menu
```

Screenshot 2023-01-05 at 12 23 16

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -16,7 +16,7 @@
   <script src="cordova.js" defer></script>
   <% } else if (__STACK_ASSETS__) { %>
   {{.CozyClientJS}}
-  {{.CozyBar}}
+  {{.CozyFonts}}
   <% } else { %>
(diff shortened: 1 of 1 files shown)


src/index.ejs
```diff
@@ -16,7 +16,6 @@
   <script src="cordova.js" defer></script>
   <% } else if (__STACK_ASSETS__) { %>
   {{.CozyClientJS}}
-  {{.CozyBar}}
```
cballevre, line 19: `{{.CozyBar}}` was in charge to import the fonts.css so now you have to import it with `{{.CozyFonts}}` (cc: https://github.com/cozy/cozy-drive/pull/2812)

Crash-- (approved):

PR #2608, opened 2023-02-14
fix: Prevent extraneous recurrence service calls

taratatach:
The recurrence service is built in such a way that each call modifying
transactions will trigger another service call to make sure all
transaction-recurrence associations that could be made have been made.
This can result in a long chain of calls, potentially infinite, if
transactions are modified on each run.

We try to limit this effect by not modifying transactions that are
already associated with a recurrence. This should prevent situations
were the service juggles between 2 recurrences with a given transaction.

The fix has the added benefit of not modifying associations that were
made by the user and to present a stable automatic association.

```
### 🐛 Bug Fixes

* The recurrence service should not modify transactions already associated with a recurrence

### 🔧 Tech

* Clarified recurrence service functions and variables names
```

### DIFF
--- a/src/ducks/recurrence/service.js
+++ b/src/ducks/recurrence/service.js
@@ -106,10 +106,12 @@ export const logRecurrencesLabelAndTransactionsNumber = ({
 
 /**
  * Fetches
- *   - transactions in the last 100 days
+ *   - transactions from the last 100 days with no recurrences
  *   - current recurrences
- * and update recurrences and operations according to recurrence matching algorithm.
- * The date taken into account to create recurrences is `transaction.date` and not `transaction.realizationDate`.
+ * and updates recurrences and transactions according to the recurrence matching
+ * algorithm.
+ * The date taken into account to create recurrences is `transaction.date` and
+ * not `transaction.realizationDate`.
  *
  * Called inside service.
  */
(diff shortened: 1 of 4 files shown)


src/ducks/recurrence/service.js
```diff
@@ -129,15 +131,23 @@ const main = async ({ client }) => {
     )
     const transactions = await client.queryAll(transactionQuery)
 
+    // Filter out transactions already associated with a recurrence as we don't
+    // want to change it.
+    const newTransactions = transactions.filter(
+      transaction =>
+        transaction.relationships == null ||
+        transaction.relationships.recurrence == null
+    )
```
Crash--, line 140: Why not using a mango selector or a partial index for that?

Crash-- (approved):

src/ducks/recurrence/service.js
taratatach: We discussed this point with @nono and figured we might be better off filtering the transactions in the service since we have good bandwidth and a low latency on workers.
Do you see an efficient way to build a mango query for this? It looks like we'd need to use an `$or` operator which is quite slow if my memory serves me well.

Crash--: > We discussed this point with @nono and figured we might be better off filtering the transactions in the service since we have good bandwidth and a low latency on workers.

It depends if there is a lot of transactions with a recurrence assigned. There is a lot of instances with 6/10 years of bankings transactions. Fetching 1000 per 1000 is not efficient at all : 
- for the total duration of fetching all the transactions 
- for the total duration of the recurrence service
- for the RAM that takes on the worker 

The first 2 are a concern to me since the recurrence service is made to be call pretty often and it takes a lot of time to execute. If we can do better, we should try even if this is to win only an execution round

But yes, `$or` will make a full scan in memory... I'll not block this PR since I don't have any time to think about it, maybe @paultranvan will have an idea, but we can merge it and improve later if we can

taratatach: Old instances shouldn't be an issue here. The query only fetches transactions from the last 100 days. There shouldn't be many instances with more than 10 transactions per day on average so we should rarely do more than one request.

paultranvan: Why not a partial filter ? 
```js
Q('io.cozy.banks.transactions')
  .where({
      date: {
        $gt: 100DaysBefore
      }
  })
  .partialFilter({
      'relationships.recurrence': {
        $exists: false
    })
  ```
  
With a partial filter, there is no performance issue at querying time since the filter is applied at indexing time. Also, I assume you only need to evaluate if there is no recurrence relationship? If yes, you just need the `$exists` predicate on `relationships.recurrence`.

taratatach: Thanks for the tip. This is done in https://github.com/cozy/cozy-banks/pull/2610.

PR #2610, opened 2023-02-16
perf: Improve recurrence service filtering

taratatach:
We can skip transactions with a recurrence or marked as not recurrent
via a partial filter instead of requesting all the transactions first
and then loop over them to remove those that we don't want to modify.

```
### 🔧 Tech

* Improved performance of transactions filtering in the recurrence service
```

### DIFF
--- a/src/ducks/recurrence/service.js
+++ b/src/ducks/recurrence/service.js
@@ -19,20 +19,34 @@ export const NB_DAYS_LOOKBACK = 100
 
 /**
  * Fetch transactions from 3 months back if there is already recurrences,
- * otherwise all transactions
+ * otherwise all transactions.
+ * We filter out transactions already associated with a recurrence as we don't
+ * want to change it.
+ *
  * @param {array} recurrences - Hydrated bundle of recurrences
  * @returns transactions
  */
 const makeQueryForTransactions = recurrences => {
   if (recurrences.length === 0) {
+    // We still filter out transactions that were explicitely marked as not
+    // recurrent.
     return Q(TRANSACTION_DOCTYPE)
+      .partialIndex({
+        'relationships.recurrence': { $exists: false }
+      })
+      .limitBy(1000)
   } else {
     const lookbackDateLimit = addDays(new Date(), -NB_DAYS_LOOKBACK)
-    const query = Q(TRANSACTION_DOCTYPE).where({
-      date: {
-        $gt: lookbackDateLimit.toISOString().slice(0, 10)
-      }
-    })
+    const query = Q(TRANSACTION_DOCTYPE)
+      .where({
+        date: {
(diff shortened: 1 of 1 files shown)


nono (changes requested):
Are you really sure about this change? I don't know cozy-client syntax, but it looks like the filtering on the date is not longer done at indexing time, but at query time, and if it is the case, it would be a bad regression for performances.

src/ducks/recurrence/service.js
```diff
@@ -19,20 +19,31 @@ export const NB_DAYS_LOOKBACK = 100
...
  * @returns transactions
  */
 const makeQueryForTransactions = recurrences => {
   if (recurrences.length === 0) {
-    return Q(TRANSACTION_DOCTYPE)
+    // We still filter out transactions that were explicitely marked as not
+    // recurrent.
+    return Q(TRANSACTION_DOCTYPE).partialFilter({
+      'relationships.recurrence': { $exists: false }
+    })
```
paultranvan, line 35: You may want to specify a `.limit(1000)`, otherwise, the pagination will be 100 by default, resulting in ~10x more queries

paultranvan, line 35: For this particular case, I'm not sure the partialFilter is actually worth it: this will create an index for probably few docs (the ones marked as not recurrent), and will use the `_find` endpoint instead of the `_all_docs` endpoint. And the later is actually more efficient

nono:
It looks like I didn't understand the index/query made by cozy-client.

src/ducks/recurrence/service.js
taratatach: As I don't have any information on the number of marked documents I guess it's a good bet to remove the partial index but we might have more marked documents than we think when there aren't any recurrence. Especially if the user deleted all their recurrences (all associated transactions are then marked as not recurrent).

taratatach: Ah, one more thing : removing the partial filter here means we'll need to loop over the fetched transactions to skip the few ones marked as not recurrent :/

Crash-- (approved):
Can you squash the two commits before rebase & merge ? 🙏

taratatach:
Yes, I never leave fixups around :).

src/ducks/recurrence/service.js
paultranvan: Well, if we can't estimate the potential number of marked documents, and therefore the actual impact, I would say it's your call :)
I tend to think the average case would be in favor of removing the partial filter, but I don't have any data to support this, and the need to loop over all the transactions afterwards to filter out the recurrences is cumbersome. So either way, it's fine by me

paultranvan (approved):

PR #2628, opened 2023-03-17
feat: Export loan attributes and vendor account ID

taratatach:
The vendor account ID will be used during the import to reconciliate
accounts while the loan attributes are just data we need to export for
users to have a copy.

```
### ✨ Features

* Vendor account IDs are exported for every account
* Loan attributes are exported for these accounts

### 🐛 Bug Fixes

* Transactions value dates are exported as short strings (i.e. `YYYY-MM-DD`)
```


Merkur39 (approved):
LGTM 👍

PR #2631, opened 2023-03-21
feat: Add CSV import service

taratatach:
This service, invoked programmatically with a `fileId` field, will
fetch the `fileId` file from the Cozy, parse it according to our CSV
schema and import the transactions, accounts, tags and recurrences it
contains.

Accounts without transactions won't be imported though as they're of
questionnable use (they're stale data).

If the Cozy already contains banking data, we'll try our best to
reconciliate the existing and new data to avoid duplicates and
recreate missing relationships.

We use `cozy-doctypes`' `BankAccount` and `BankTransaction`
reconciliation algorithms to make any debbuging easier but we had to
come up with our own reconciliation algorithms for tags and
recurrences (they're really simple though).

```
### ✨ Features

* Add CSV import service
* Allow grouping bank accounts by institution labels in Settings
```

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -107,8 +107,12 @@
     "export": {
       "type": "node",
       "file": "export.js",
-      "comment": "This service must be called programmatically from Banks.",
-      "debounce": "5m"
+      "comment": "This service must be called programmatically from Banks."
+    },
+    "import": {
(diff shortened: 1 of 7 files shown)


Merkur39 (approved):
LGTM 👏

PR #2653, opened 2023-04-19
feat: Save imported transactions count

taratatach:
We want to tell the user how many transactions were imported upon
success so they can verify everything they wanted to import was
actually imported (n.b. not importing any transaction is not
considered as a service failure since no errors were raised).

Since the service cannot communicate with the GUI via its
`io.cozy.jobs` document, we'll save this count into the
`io.cozy.bank.settings` document.

```
### ✨ Features

* Save imported transactions count in `io.cozy.bank.settings` document

### 🐛 Bug Fixes

* Fix typo in French import success title
```


Merkur39 (approved):
LGTM 👍

PR #2659, opened 2023-05-03
fix: Balance(Lower|Greater) notifications WIP

Merkur39:
#### Current behavior:
If the conditions defined in “low balance” and/or “high balance” are met, then a notification is sent to the user to inform him that an account (or group of accounts) has reached this balance.
If the balance changes, and the condition is still met, then a 2nd notification is sent.

#### New behavior:
If the balance of the account (account group) changes, but the condition is maintained, then there are no new notifications.
If the balance of the account (account group) goes out of the condition, then the notification can be sent when the condition is met again.

```
### 🐛 Bug Fixes

* Reduced notification sending about greater/lower balances.
```


src/ducks/notifications/BalanceLower/index.js
```diff
@@ -82,20 +89,24 @@ class BalanceLower extends NotificationView {
...
+          this.filterForRule(rule, acc, lastBalances)
         )
       }))
       .filter(({ accounts }) => accounts.length > 0)
   }
 
-  fetchData() {
-    const matchingRules = this.findMatchingRules()
+  async fetchData() {
+    const { balancesNotifications: lastBalances } = await fetchSettings(
```
JF-Cozy, line 105: nit: si on a besoin de renommer ici pour mieux comprendre, peut-être qu'on pourrait renommer `balancesNotifications` partout en `lastBalancesNotifications` ou un truc du genre ? 🤔

JF-Cozy (approved):

src/ducks/settings/constants.js
```diff
@@ -5,6 +5,7 @@ export const DEFAULTS_SETTINGS = {
   _type: 'io.cozy.bank.settings',
   _id: 'configuration',
   id: 'configuration',
+  balancesNotifications: {},
```
Crash--, line 8: can you update the cozy-doctypes repo if you store more information?

Merkur39: yes it is planned, now, here => https://github.com/cozy/cozy-doctypes/pull/230

PR #2681, opened 2024-01-03
fix: Do not create AppSuggestion for non stable konnector

Crash--:
```

### 🐛 Bug Fixes

* We need to check the `latest_version` attribute to know if there is a stable version or not.
Before the fix, we created app suggestion even for beta or dev konnectors.
```


Merkur39 (approved):
Good, thanks for the doc

PR #2683, opened 2024-01-09
fix: Throw error when there is an error catch

Crash--:
We catch the error, but don't throw them after that. Resulting that services were considered OK...

```
### ✨ Features

*

### 🐛 Bug Fixes

* Make the services failed if they... failed.

### 🔧 Tech

*
```


Merkur39 (approved):
On en profiterait pas pour ajouter également un `throw` sur les autres fonctions utilisées dans `onOperationOrBillCreate` ? (`doSendNotifications`, `doAppSuggestions`)

PR #2688, opened 2024-01-23
fix: Crash of onOperationOrBillCreate service due to `window`

Merkur39:
This service no longer worked since the work on `brands`, the cause coming from the call to the `getBrands` function called in `getBillRegexp`.
We now pass the `brands` to him via his arguments.

This is a fix to get the service back up and running while waiting for a better approach.

```
### 🐛 Bug Fixes

* Fix crash of onOperationOrBillCreate service due to `window`
```

### DIFF
--- a/src/ducks/billsMatching/Linker/Linker.js
+++ b/src/ducks/billsMatching/Linker/Linker.js
@@ -267,12 +267,14 @@ class Linker {
     bill,
     debitOperation,
     allOperations,
-    options
+    options,
+    brands
   ) {
     const creditOperation = await findCreditOperation(
       bill,
       options,
-      allOperations
+      allOperations,
+      brands
     )
 
     const promises = []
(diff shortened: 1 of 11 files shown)


src/ducks/billsMatching/Linker/Linker.js
```diff
@@ -345,7 +348,7 @@ class Linker {
    * @see https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.bills/
    * @see https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.bank/#iocozybankoperations
    */
-  async linkBillsToOperations(bills, operations, options) {
+  async linkBillsToOperations(bills, operations, options, brands) {
```
Crash--, line 351: can you update the JSDoc please?

src/ducks/billsMatching/Linker/Linker.js
```diff
@@ -308,23 +310,24 @@ class Linker {
    *
    * @returns {(Object|false)} The io.cozy.bank.operations that matched or false
    */
-  async linkBillToDebitOperation(bill, allOperations, options) {
+  async linkBillToDebitOperation(bill, allOperations, options, brands) {
```
Crash--, line 313: JSDoc please 🙏

Crash-- (approved):
Nice thanks!

PR #2694, opened 2024-01-26
[VO-293] fix: Regression discover from the previous library update into 1.60.0

cballevre:
```
### 🔧 Tech

* Update cozy-harvest-lib from 19.0.1 to 22.0.1
```


Crash-- (approved):
Nice !

PR #2695, opened 2024-02-05
[VO-195] fix: The group sum was not updated after the account was toggled

cballevre:
I wasn't able to recover the interaction by splitting the component into subcomponents. I removed the React.memo for the bug. The performance problems are higher up in the component hierarchy. The balance graph takes a long time to render. We should probably review the calls in queries that re-render all the children multiple times. So this is a quick fix 😕

```
### 🐛 Bug Fixes

* Update the group sum after hiding account
```

**Example of the bug :**
Capture d’écran 2024-02-05 à 09 11 30


src/ducks/balance/GroupPanel/index.js
```diff
@@ -0,0 +1 @@
+export { default as GroupPanel } from 'ducks/balance/GroupPanel/GroupPanel'
```
Crash--, line 1: nit: Barrel files are bad:
- They are bad for webpack treeshaking 
- They are bad for webpack performance (because in a large codebase we need to import hundreds / thousands barrel files, and it hurts perf)

src/ducks/balance/GroupPanel/GroupPanel.jsx
```diff
@@ -98,4 +98,4 @@ GroupPanel.defaultProps = {
   withBalance: true
 }
 
-export default React.memo(GroupPanel)
+export default GroupPanel
```
Crash--, line 101: This is because `group` was updated but `GroupPanelSummary` was not re-render?

src/ducks/balance/GroupPanel/index.js
cballevre: I hadn't realised that I was doing it to simplify my developer experience, but it shouldn't come at the sacrifice of the user experience

Crash--: It comes at the cost of your DX too. When I talk about webpack performance, this is on your machine ;). Also, now your IDE should be smart enough to be able to import the file correctly without barrel file. Your DX should not be impacted by that change.

src/ducks/balance/GroupPanel/GroupPanel.jsx
cballevre: The GroupPanel was never re-rendered even if the group changed. The rows are updated by side-effect because their props triggers change. This gives the impression that everything is fine except for the summary.



Capture d’écran 2024-02-05 à 16 42 35

cballevre: The graph is super expensive. I tried to avoid render re-renders by splitting component, but I didn't manage to optimise that part. You'd probably have to use useQuery more than connects to avoid them rendering the whole tree for each query result.

cballevre:
@Crash--  can I have a new review without taking performance into account?

Crash-- (approved):

PR #2700, opened 2024-02-15
[VO-400] fix: Show konnector error on transaction list

cballevre:
```
### 🐛 Bug Fixes

* Show konnector error on transaction list
```

`isBankTrigger` didn't have the `bankingSlug` when it was run with the HOC. As the function with `bankingSlug` is also exposed via a context, so I chose this way. There's also the advantage of having one less request because you don't have to execute the hook several times.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
 # General code owners
-*      @JF-Cozy @Merkur39
+* @cballevre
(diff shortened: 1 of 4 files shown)


Ldoppea (approved):
I'm not sure to understand the initial problem. Was it about the component not being updated when using `withBankingSlug`?

If yes does this means the problem exists on other class component that use it? (PersonalInfoDialog)

Crash-- (approved):
nice

cballevre:
@Ldoppea Yes, the problem was that the component didn't render after the `bankingSlug` list reference change in the function ([code](https://github.com/cozy/cozy-banks/blob/fix/show-konnector-error-on-transaction-list/src/hooks/useBankingSlugs.jsx#L35)). The problem is that the component is memoised for performance reasons. Its parent is re-rendered each time it is modified in the cozy-client redux store as a side effect. Using a context gets around this problem.

I've chosen not to rework the `PersonalInfoDialog` component because it isn't memoised and it's part of a feature that hasn't been released for a long time: transfers.

PR #2702, opened 2024-02-19
feat: edit transaction label in list and modal

feugy:
### ✨ What's in there?

Hi Cozy! As discussed [on the forum](https://forum.cozy.io/t/cozy-banks-categories-personnalisees-commentaires-texte/7016/9?u=feugy) this is an attempt to allow users editing their transaction labels.

image

After considering adding a new comments field, I've leveraged the existing `manualLabel` which is used in recurring transactions.

Users can now:
- edit their transaction labels in the transaction list (and search results). A new "pen" icon button turns the read-only cell into a text field (`LabelField` React component).
- edit their transaction labels in the transaction modal, in a similar fashion.
- see original labels, in case they changed some.
- reset to the original label, with a dedicated "restore" button in the text field.
- from the transaction modal, search for the manual or original label.
- take advantage of keyboard navigation: when entering edition mode, the text field grabs focus. While editing, `Enter` will close and save, `Escape` will close without saving. All icon are buttons which are in the tab cycle. 

### 🧪 How to test?

Here is a possible scenario:
1. Grab the code, locally, assuming you already have Cozy server set up and ready, with some fixtures install
1. Connect to http://banks.cozy.localhost:8080/#/balances/details
    > in the transaction list, you'll have new pen icons
1. Click on a pen icon
    > the label is replaced with a text field with the original label as value, focus, and a close button in it  
1. Change the value, and click on the close button/hit Escape key/click somewhere else
    > the manual value is saved, text field replaced with read-only value, and the original value is visible below.
1. Click on the transaction row
    > it opens the modal, with the manual and original labels, both followed by search icons
1. Click on the manual label pen icon
    > the manual field is editable, it has focus, close and restore icon
1. Hit Tab twice to focus on the restore icon, then Enter to trigger it
    > the manual label is discarded, text field replaced by read-only value. The original label row is gone, and the value in transaction list is also restored

There's plenty of other ways to test it, you can also try to search for manual labels.

### ❗ Notes to reviewers

This PR is draft, and I'd like to collect some feedback before adding automated tests.
In particular, I'd like to draw reviewers' attention on some tradeoffs:
1. the "pen" icon button are very small, which isn't convenient on mobile. However making them bigger makes the transaction row higher, which does not look good

1. the "close" button comes before "restore" button to make keyboard edition smoother.

1. the transaction is only saved when the text field is losing focus: no complicate throttle needed

1. I've styled the TextField so it looks like a Input field with "tiny" size, like in cozy-drive [FilenameInput](https://github.com/cozy/cozy-drive/blob/master/src/modules/filelist/FilenameInput.jsx) 

Finally it worth calling out the great dev material you made to ease contributions. The setup isn't too complicated, and it's easy for a React dev to contribute.
I couldn't make HMR work (despite running my cozy-app-dev docker container with `-e COZY_DISABLE_CSP=1`), but it's not too bad.

### DIFF
--- a/src/ducks/transactions/TransactionRow/LabelField.jsx
+++ b/src/ducks/transactions/TransactionRow/LabelField.jsx
@@ -1,0 +1,130 @@
+import { useClient } from 'cozy-client'
+import React, { useCallback, useRef, useState } from 'react'
+
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import IconButton from 'cozy-ui/transpiled/react/IconButton'
+import CrossCircleOutlineIcon from 'cozy-ui/transpiled/react/Icons/CrossCircleOutline'
+import RestoreIcon from 'cozy-ui/transpiled/react/Icons/Restore'
+import PenIcon from 'cozy-ui/transpiled/react/Icons/Pen'
+import { CircularProgress } from 'cozy-ui/transpiled/react/Progress'
+import TextField from 'cozy-ui/transpiled/react/TextField'
+import Typography from 'cozy-ui/transpiled/react/Typography'
+import {
+  getLabel,
+  hasEditedLabel,
+  setTransactionLabel
+} from 'ducks/transactions/helpers'
+
+import styles from 'ducks/transactions/TransactionRow/LabelField.styl'
+
+export default function LabelField({ transaction, variant = 'body1' }) {
+  const client = useClient()
+  const [isEditing, setIsEditing] = useState(false)
+  const [isSaving, setIsSaving] = useState(false)
+  const inputRef = useRef()
+  const containerRef = useRef()
+
+  const saveAndClose = useCallback(
+    async label => {
+      if (
+        label !== transaction.manualLabel &&
+        label !== getLabel(transaction, true)
+      ) {
+        const sanitized = label?.trim()
+        try {
+          setIsSaving(true)
+          await client.save(
+            setTransactionLabel(
+              transaction,
+              sanitized?.length ? sanitized : null
+            )
+          )
+        } finally {
+          setIsSaving(false)
+        }
+      }
+      setIsEditing(false)
+    },
+    [client, setIsEditing, setIsSaving, transaction]
+  )
+
+  const handleEdit = useCallback(
+    event => {
+      event.stopPropagation()
+      setIsEditing(true)
+    },
+    [setIsEditing]
+  )
+
+  const handleClose = useCallback(
+    () => saveAndClose(inputRef.current.value),
+    [saveAndClose]
+  )
+
+  const handleRestore = useCallback(() => {
+    saveAndClose(null)
+  }, [saveAndClose])
+
+  const handleKeyDown = useCallback(
+    ({ key }) => {
+      if (key === 'Enter') {
+        handleClose()
+      } else if (key === 'Escape') {
+        setIsEditing(false)
+      }
+    },
+    [handleClose, setIsEditing]
+  )
+
+  const handleBlur = useCallback(
+    ({ relatedTarget }) => {
+      if (containerRef.current?.contains(relatedTarget)) {
+        // do not close when focused element is inside the component
+        return
+      }
+      saveAndClose(inputRef.current.value)
+    },
+    [saveAndClose]
+  )
+
+  return isEditing ? (
+    <div
+      onBlur={handleBlur}
+      ref={containerRef}
+      onClick={e => e.stopPropagation()}
+    >
+      <TextField
+        autoFocus
+        fullWidth
+        defaultValue={getLabel(transaction)}
+        className={styles.EditableLabel}
+        inputRef={inputRef}
+        onKeyDown={handleKeyDown}
+        variant="outlined"
+        InputProps={{
+          endAdornment: isSaving ? (
+            <CircularProgress size={24} />
+          ) : (
+            <>
+              <IconButton size="small" onClick={handleClose}>
+                <Icon icon={CrossCircleOutlineIcon} size={14} />
+              </IconButton>
+              {hasEditedLabel(transaction) ? (
+                <IconButton size="small" onClick={handleRestore}>
+                  <Icon icon={RestoreIcon} size={14} />
+                </IconButton>
+              ) : null}
+            </>
+          )
+        }}
+      />
+    </div>
+  ) : (
+    <div className={styles.FixedLabel}>
+      <Typography variant={variant}>{getLabel(transaction)}</Typography>
+      <IconButton size="small" onClick={handleEdit}>
+        <Icon icon={PenIcon} color="var(--coolGrey)" size={14} />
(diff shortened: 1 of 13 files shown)


cballevre:
Thank you for taking the time to make this PR 👏  I'm testing all these tomorrow morning

src/ducks/transactions/TransactionRow/TransactionRowDesktop.jsx
cballevre, line 1: We do not wish to add the edition button into transaction row. We understand the desire to speed up editing but here are the reasons why:
1. Displaying an edit button for all rows clutters the interface for passive users. A solution based on hovering could work on the desktop, but it would be inconsistent with mobile devices because the notion of hovering does not exist
2. We can unintentional opening of the detail modal, for example when a text is selected and the click is dropped outside the input zone, can confuse users

src/ducks/transactions/TransactionRow/LabelField.jsx
```diff
@@ -0,0 +1,130 @@
...
+                  <Icon icon={RestoreIcon} size={14} />
+                </IconButton>
+              ) : null}
+            </>
+          )
+        }}
+      />
+    </div>
+  ) : (
+    <div className={styles.FixedLabel}>
```
cballevre, line 123: nit: We could replace this classes with utility classes like `u-flex u-flex-items-center` and `u-ml-haft` on the `IconButton`

cballevre (commented):
The only issue concerns the edit button in the transaction row. Otherwise, the other points are improvements and the PR seems ready to me 👍

src/ducks/transactions/TransactionRow/TransactionRowDesktop.jsx
feugy: Thanks for the feedback. I'll obviously implement it, but I felt the necessity, as a long-time user, to express my disagreement.

Adjusting labels is the #1 missing feature of Cozy banks in my experience. Bank labels are often truncated, opaques and sometimes unusable. After 10 years of self-accounting with MS Money, then Grisbi, I really miss the ability to change them, and what we're adding in this PR will only solve half of the problem.

I doubt mobile users would often edit their labels. Hence why I champion the idea to only edit within the modal.
However, we can come up with a different solution for desktop users, that enables a painless and efficient editing.

What do you think of:
1. mobile transaction row: read-only label. Tap to open modal. In the modal, the label is always a text field.
2. desktop transaction row: turn read-only label to text field on hover and focus, allowing in-place editing (pretty much like a spread-sheet). Click elsewhere on row to open modal. In the modal, the label is always a text field.

Despite my disagreement, be assured that I'll implement whatever solution your team prefers. Cozy banks is a wonderful product already!

cballevre: I'll check with your designer to see if we can't find a compromise 🙂

cballevre: Just a few questions to clarify my answer. What need made you want to edit the label of a transaction?  Did you want to edit several transactions? For the same type of label or each one unique? Or something else altogether?

feugy: Some examples of labels which needs to be edited:
1. checks
   image
1. bank labels which don't make sense
   image
1. DAB
   image
1. incomplete, truncated labels
   image


Half of the time edited label is ad-hoc (it's not reused), like for checks. Multiple editing isn't needed here.

But sometimes I'll be always replacing one label (like "4f" and "Le Court Circui" examples, which where truncated).
For these, the ideal solution would be to automatically apply the override when receiving new transactions.
Multiple editing could be used here, but I'm not 100% convinced.

I obviously can't share too many screenshots here, but I estimate that I'll edit a good half of the transaction I'm receiving. Automation would be huge pain relief, and that was my next objective.

cballevre: Thanks for the details. It's true that multiple edition doesn't meet the need, it's just a nice to have.

Concerning the label edition into rows, our designer is sticking to his first vision. Our interfaces are designed to be mobile-first, and his decisions reflect this constraint.

However, by explaining your problem to him better, he came back with [mockups](https://www.figma.com/file/Cc7l3ga0Ki5pMNn44t4ri0/Banks---Specs?type=design&node-id=1930-7900&mode=design&t=R434Ft0VapnG3otC-0) of a different approach. It may not be as fast on the short term but it will benefit from automation. A bit like Proton with their filters : https://proton.me/support/email-inbox-filters. This would require adding a rules tab in the settings to view them and add the more complex ones.

As this would require you to open the modal, then click on the edit button to modify it, we could also offer an alternative route via the quick action menu when selecting a transaction. 

I'm not very familiar with how transaction processing works. I'm going to take the opportunity to look at the services that manage this part to see what we could do in terms of automation.

feugy: Hello!

Back from a week off, I'll probably resume working on this next weekend, if I find time. The code will be very different from this PR, so I'll open a new one.

About the mockups, can your designer check my understanding of his proposal?
1. given the transaction list is opened, 
   when clicking/tapping on a transaction, 
   then it opens the transaction details modal (desktop) or drawer (mobile),
   and there is a "pen" icon button next to the transaction label
   
1. given the transaction details modal/drawer is visible
   when clicking on the "pen" icon button
   then it opens a modal to edit the transaction label
   
On mobile, there will be a modal on top of the drawer, but on desktop, it's a modal on top of another modal. Is it intended?
Also, where do we display the original label when users have provided their own version?

joel-costa: Hi feugy,

I'm Joël, the guy in charge of design. To answer your question:

- Yes, you understood my proposal
- Yes, the modal on top of another modal is wanted. But be careful, you're talking about "modal" in desktop and "drawer" in mobile. We use a single component, the CozyDialog component: https://docs.cozy.io/cozy-ui/react/#/CozyDialogs 
The expected behavior is to open a `cozydialog` in `medium` to display the transaction details, and to open a `cozydialog` in `small` to display the renaming on top.
- I've added a screen here to show where the original name can be displayed: https://www.figma.com/file/Cc7l3ga0Ki5pMNn44t4ri0/Banks---Specs?type=design&node-id=1947%3A38952&mode=design&t=0nkJ9eWLdmMeEp75-1

I also took the opportunity to modify this screen to give it a very long original name: https://www.figma.com/file/Cc7l3ga0Ki5pMNn44t4ri0/Banks---Specs?type=design&node-id=1930%3A14079&mode=design&t=0nkJ9eWLdmMeEp75-1
In my artboard, I've written the name with several lines, whereas today it's simply cut off. I think it's important for a mobile user to be able to read the full name, so I've slipped this change in, but don't really feel obliged to touch it, I can always bother cballevre with it on occasion 😇

